PR #9, opened 2018-11-09
Markdown: Removed/replaced closing tags that were never opened.

m-mohr:
Would be great to have a 0.3.2 release with these changes as it breaks my documentation. ;-)


fmvilas:
It's already released as 0.3.2! Thanks!

m-mohr:
Thank you!

PR #12, opened 2018-11-12
Pure markdown headings

m-mohr:
Markdown template: I replaced the HTML headings with pure markdown for two reasons:

1. Markdown should contain as much markdown as possible instead of HTML.
2. If you use documentation generators for markdown such as MkDocs, it will fail to generate good table of contents as it is missing all the HTML headings. This leads to a not very useful TOC such as:

```
Table of Contents
Topics
Message
Payload
Example
Message
Payload
Example
Message
Payload
Example
Message
Payload
Example
Message
Payload
Example
...
```

### DIFF
--- a/templates/markdown/.partials/message.md
+++ b/templates/markdown/.partials/message.md
@@ -1,9 +1,4 @@
-<h3 class="message__header">
-  {{messageName}}
-  {{#if message.deprecated}}
-  <em>Deprecated</em>
-  {{/if}}
-</h3>
+### {{#if message.deprecated}} *Deprecated*{{/if}} {{messageName}}
 {{#if message.summary}}
 {{{message.summary}}}
 
--- a/templates/markdown/.partials/message.md
+++ b/templates/markdown/.partials/message.md
@@ -1,4 +1,4 @@
-### {{#if message.deprecated}} *Deprecated*{{/if}} {{messageName}}
+### {{messageName}} {{#if message.deprecated}} (**deprecated**){{/if}} 
 {{#if message.summary}}
 {{{message.summary}}}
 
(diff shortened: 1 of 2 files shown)


templates/markdown/.partials/message.md
```diff
@@ -1,9 +1,4 @@
-<h3 class="message__header">
-  {{messageName}}
-  {{#if message.deprecated}}
-  <em>Deprecated</em>
-  {{/if}}
-</h3>
+### {{#if message.deprecated}} *Deprecated*{{/if}} {{messageName}}
```
fmvilas: I'd use parenthesis or square brackets around the word "deprecated" to make it easier to understand it's not actually part of the name. I know it wasn't there before, just taking advantage that we're changing it.

templates/markdown/.partials/topic.md
```diff
@@ -1,10 +1,6 @@
 <a name="topic-{{topicName}}"></a>
-<h3>
-{{~#if topic.deprecated}}<em>Deprecated</em>{{~/if}}
-{{~#if topic.publish}}<code>publish</code>{{~/if}}
-{{~#if topic.subscribe}}<code>subscribe</code>{{~/if}}
-{{topicName}}
-</h3>
+
+### {{~#if topic.deprecated}} *Deprecated*{{~/if}}{{~#if topic.publish}} `publish`{{~/if}}{{~#if topic.subscribe}} `subscribe`{{~/if}} {{topicName}}
```
fmvilas: Ditto

templates/markdown/.partials/message.md
m-mohr: Sure, changed it.

templates/markdown/.partials/topic.md
m-mohr: Sure, changed it.

m-mohr:
@fmvilas What's the release cycle of this tool? (aka: Kind question to when can I expect these changes to be introduced to my documentation? ;-) ) Thanks!

fmvilas:
There's no release cycle for now, but answering your question, this is already on npm. Thanks! :)

m-mohr:
Thanks a lot!

fmvilas:
Thanks for your contributions! Really appreciate!

PR #13, opened 2018-11-12
Don't show "messages" header if there are no messages defined.

m-mohr:
The TOC item is not shown, but the header was still in the markdown file. This PR makes it consistent.


fmvilas:
Awesome, thanks!

PR #14, opened 2018-11-12
Clean-up HTML / Markdown

m-mohr:
The markdown generated is pretty dirty as it has a lot of HTML markup in the files. It makes the temples prone to errors as HTML is more complex (see #9) and can have side effects with some markdown tooling (see #12). Additionally, there are CSS classes all over the place, which can easily be replaced with simple CSS element selectors. Also, list of tags for example can easily be represented as lists for example, instead of nested divs. So this is my proposal to clean things up and simplify the markdown templates.

### DIFF
--- a/templates/markdown/.partials/tags.md
+++ b/templates/markdown/.partials/tags.md
@@ -1,7 +1,5 @@
-<div class="tags">
+
 {{#each tags}}
-  <div class="tags__tag">{{./name}}</div>
-{{else}}
-  <div class="tags__no-tags">No tags</div>
+* {{./name}}
 {{/each}}
-</div>
+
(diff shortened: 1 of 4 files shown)


templates/markdown/.partials/tags.md
```diff
@@ -1,7 +1,5 @@
-<div class="tags">
+
 {{#each tags}}
-  <div class="tags__tag">{{./name}}</div>
-{{else}}
-  <div class="tags__no-tags">No tags</div>
```
m-mohr: Note to this change: the else-condition is never reached as all template includes that reference this file have a conditional that checks for tags in-place (something like `if x.tags ...`).

PR #17, opened 2018-11-21
Feature/support binary files

rmelian:
adding support for ignore files on templates


templates/java-spring/config.json
```diff
@@ -0,0 +1,5 @@
+{
```
fmvilas: Use another name for the file. `config.json` can easily collide with some project template needs. I'd suggest something like `.ag-config.json`.

lib/generator.js
```diff
@@ -50,22 +51,34 @@ const renderFile = (filePath, data) => new Promise((resolve, reject) => {
...
   const target_dir = options.target_dir;
   const file_name = options.file_name;
   const root = options.root;
   const data = options.data;
 
-  renderFile(path.resolve(root, file_name), data)
-    .then((parsed_content) => {
-      const template_path = path.relative(templates_dir, path.resolve(root, file_name));
-      const generated_path = path.resolve(target_dir, template_path);
+  if (isFileIgnored(file_name, template_config.ignore)) {
```
fmvilas: Maybe we should come up with a better name here and in the config file? Ignoring sounds like it shouldn't be even copied.

templates/java-spring/config.json
```diff
@@ -0,0 +1,5 @@
...
+  "ignore": [
```
fmvilas: Ignore sounds like something different. Shall we use another name?

templates/java-spring/config.json
rmelian: what about `.tp-config.json`?

fmvilas: That sounds good. Whatever is not "easy to collide with" 😊

Issue #16, opened 2018-11-20
Support marking files as executable


fmvilas:
### What?
When copying a file from the template to the destination directory, it should respect the file's chmod configuration. At least the executable part. For instance, when generating a Java project, we want `gradlew` to be executable.

### Why?
We should offer a ready-to-use result from the templates.

rmelian:
I started working on this...

fmvilas:
You're on fire mate! :fire:

rmelian:
we can close this as the code is merged in develop branch

PR #19, opened 2018-11-27
#16 change gradlew permissions, keep template files permissions

rmelian:


lib/generator.js
```diff
@@ -72,15 +72,28 @@ const generateFile = options => new Promise((resolve, reject) => {
         const template_path = path.relative(templates_dir, path.resolve(root, file_name));
         const generated_path = path.resolve(target_dir, template_path);
 
-        fs.writeFile(generated_path, parsed_content, 'utf8', (err) => {
-          if (err) return reject(err);
-          resolve();
+        stat(source_path)
+          .then(stats => {
+            fs.writeFile(generated_path, parsed_content, {encoding: 'utf8', mode: stats.mode}, (err) => {
```
rmelian: @fmvilas there is probably a cleaner way to do this. I mean the promise inside another one

Issue #42, opened 2019-02-21
Java-spring templates have undefined members


RobertDiebels:
Both the MqttConfig and AmqpConfig partial template files will currently result in a compiler error when a spec contains a `topic` object with a  `publish` property.

In the MqttConfig template the [setDefaultTopic function](https://github.com/asyncapi/generator/blob/cbdae35abc594bf8bef7bb92fb3045b465b5592e/templates/java-spring/.partials/MqttConfig.java#L100) will be called with an undefined variable. Since the template only creates the corresponding private members if the topic is of type [subscribe](https://github.com/asyncapi/generator/blob/master/templates/java-spring/.partials/MqttConfig.java#L39).

The same applies to the AmqpConfig template file where the [setExchangeName function](https://github.com/asyncapi/generator/blob/cbdae35abc594bf8bef7bb92fb3045b465b5592e/templates/java-spring/.partials/AmqpConfig.java#L122) will be called with an undefined variable.  Since the template only creates the corresponding private members if the topic is of type [subscribe](https://github.com/asyncapi/generator/blob/cbdae35abc594bf8bef7bb92fb3045b465b5592e/templates/java-spring/.partials/AmqpConfig.java#L37).

rmelian:
Hi @RobertDiebels thank you for reporting this, I'll take a look ASAP

RobertDiebels:
@rmelian Happy to help. I can confirm that the expected output is generated if the handle-bars if-statement for the private members in MqttConfig is removed. As for the AqmpConfig partial, changing the if-statment condition to check for `publish` rather than `subscribe` for the Exchange private members should fix the issue.

rmelian:
do you want to submit a PR?

RobertDiebels:
@rmelian I'll submit a PR in an hour or 2-3 if that's ok.

RobertDiebels:
Done.

rmelian:
hi @RobertDiebels can you share the asyncapi definition you are using

rmelian:
I just approved your PR @RobertDiebels great work

PR #44, opened 2019-02-21
fix(undefined-vars): [ISSUE-42] - Resolves issue #42.

RobertDiebels:
Removes handlebars if-statements and changes some conditions so topic, exchange and queue variables are no longer undefined.


rmelian:
good catch

Issue #43, opened 2019-02-21
Java-spring partial template AmqpConfig wrongly adds a comma


RobertDiebels:
The java-spring template partial AmqpConfig adds a comma when it is not the [last topic](https://github.com/asyncapi/generator/blob/cbdae35abc594bf8bef7bb92fb3045b465b5592e/templates/java-spring/.partials/AmqpConfig.java#L70). 

When a situation occurs when the last topic is a subscribe topic the comma will be added to the generated source. This will cause a compile error.

Afaik handlebars can only solve this with a custom helper. Or by splitting the subscribe and publish topics into two different lists.

rmelian:
Hi @RobertDiebels good catch too, do you want to send another PR :)

fmvilas:
In case it helps: you might need to use `#unless @../last` instead of `#unless @last`.

RobertDiebels:
@rmelian I'll see if I can create a PR within an hour or 8-9.
@fmvilas Thanks! I'll check it out.

RobertDiebels:
@rmelian @fmvilas I created a PR that should fix this. I looked into the `@../last` but I got the impression @../last goes to the parent object and tries to get the last item from the parent. Where the parent would be the asyncapi itself. I prefer this fix however, since it removes the need for a bunch of if-statements in the Config partials. Let me know what you think.

rmelian:
Hi @RobertDiebels I'm sorry it took me so long to review this. This looks fantastic, I have just a few comments.

rmelian:
actually, my only comment was to change x-service-name to operationId as this will be officially supported in version 2.0.0 of the specification but I guess it was my who implemented it this way :)

great work on this, I'm approving this PR

PR #46, opened 2019-02-22
fix(comma-placement): [ISSUE-43] - Resolves issue #43

RobertDiebels:
- Added a map with subscribe-topics only to the beautifier. - Added a map with publish-topics only to the beautifier. - Removed if-statements that checked for topic type using maps instead. - Added x-service-name property to topics in ./test/docs/sample.yml

### DIFF
--- a/lib/beautifier.js
+++ b/lib/beautifier.js
@@ -110,6 +110,11 @@ module.exports = (asyncapi) => {
     });
   }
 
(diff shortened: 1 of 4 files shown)


rmelian:
this looks fantastic, tomorrow will take a closer look

thank you for this PR

PR #47, opened 2019-03-22
Hide the Topics block when no topic is describeed

ppseprus:
Based on https://github.com/asyncapi/docgen/issues/42

### DIFF
--- a/templates/html/.partials/topics.html
+++ b/templates/html/.partials/topics.html
@@ -1,4 +1,6 @@
+{{#if asyncapi.topics}}
 <h2 class="section-header">Topics</h2>
+{{/if}}
 
 {{#each asyncapi.topics}}
   {{~>topic topic=. topicName=@key ~}}
--- a/templates/html/.partials/topics.html
+++ b/templates/html/.partials/topics.html
@@ -1,6 +1,4 @@
-{{#if asyncapi.topics}}
 <h2 class="section-header">Topics</h2>
-{{/if}}
 
 {{#each asyncapi.topics}}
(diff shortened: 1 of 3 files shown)


fmvilas:
Sorry, I just realized this is on the generator. My fault.

templates/html/.partials/topics.html
```diff
@@ -1,4 +1,6 @@
+{{#if asyncapi.topics}}
```
fmvilas: I don't think this is needed. We're not including the partial in the other change, so it's safe.

ppseprus: I've checked. It does appear without the condition...

fmvilas: Yes, because this is the place where you should place the `if` condition: https://github.com/asyncapi/generator/blob/master/templates/html/.partials/content.html#L3

fmvilas: The one on the index.html is the sidebar that only appears when you download the documentation.

ppseprus: I moved the condition to the content.html.
Regarding the sidebar: Well, Im using the generator locally. Generating an html, which will be hosted in an intranet... and, the sidebar has the Topics block. Thats why I made it conditional too.

fmvilas: Yeah, that's good to me. Thanks for taking the time!

ppseprus: Haha! Well, thanks for this repo and the clarification over Sclack too! ; )

PR #51, opened 2019-03-31
Add support for version 2

### LINKED ISSUE #56: Update generator to work with 2.0.0
We should update the generator package so it works with AsyncAPI 2.0.0. It should not be backward compatible so AsyncAPI 1.x documents will not work.

fmvilas:
### What?
Work in progress. Adds support for AsyncAPI 2.0.0-rc1. This is a temporary implementation before the parser is finished.

### Why?
To offer a way to play with AsyncAPI 2.0.0 in the editor/playground.

Fixes #56


fmvilas:
I'm moving forward and gonna merge this PR. Please, if you see something wrong, missing, or just don't agree with the changes, **don't hesitate to leave a comment after it's merged**. I'll be happy to address these questions afterward.

PR #54, opened 2019-04-18
Make schema HTML support events that are tuples at root level

yli-cpr:


templates/html/.partials/schema.html
```diff
@@ -4,6 +4,25 @@
   <h4>{{schemaName}}</h4>
   {{/unless}}
 
+  {{#if schema.items.[0]}}
```
fmvilas: Just out of curiosity: doesn't `schema.items.length` work?

yli-cpr: Cool. Just verified that it works.

fmvilas: Awesome! I wasn't sure. Handlebars drives me nuts sometimes.

fmvilas:
It's now [deployed to npm](https://www.npmjs.com/package/asyncapi-generator). Thanks for the PR, @yli-cpr!

yli-cpr:
Thank you @fmvilas !

Issue #25, opened 2018-12-21
Array-level examples do now show up in Topics section


fridgerator:
They do show up in the Schemas section however:

```yml
topics:
  example:
    subscribe:
      description: Example topic
      payload:
        $ref: '#/components/schemas/example'

components:
  schemas:
    example:
      type: array
      items:
        type: number
      example: [35.4, 38.6, 46.7, 66.5]
```

Assuming an HTML generated document.

The example within the "Topics" section just shows an empty code block.  However in the schemas section section the example shows correctly.

This works the same using the online editor, as well as with the `ag` cli.

fmvilas:
Thanks for reporting @fridgerator. Will have a look at soon as I can.

my0n:
First PR I guess, tell me what you think. I tried to make the changes as noninvasive as possible. I also noticed that the same issue happens with messages, so I fixed that in the html template, too.

PR #55, opened 2019-04-26
#25 operation and message templates hide schema example and show formatted example

my0n:
I ran all given tests (including the one I added) before and after changes, and checked outputs with kdiff. This only touches the html template. Hopefully this commit PR is OK, I realize that I kind of just appeared out from nowhere!


fmvilas:
This looks good to me @my0n. Thanks for taking the time to fix it, really appreciate.

PR #59, opened 2019-05-24
Typo in variable name

marcortw:
Seems like there's an "s" missing here.


fmvilas:
Good catch! Thank you!

fmvilas:
It's on npm now. Thanks!

PR #62, opened 2019-06-05
Removed lowercase class from case-sensitive property values

csimi:
Both pattern and the values of the enum array are case-sensitive. There shouldn't be a lowercase class on them.


xino12:
Thanks!

PR #66, opened 2019-06-30
Add cli support template parameters

### LINKED ISSUE #8: CLI support template parameters
Right now the CLI doesn't support passing parameters specific to a template. 

Having a way to pass additional information to a template can open the door to a lot of flexibility. 
For example, for the java generator, we could use the following commands to generate the same code but with different build tools like maven or gradle:  
`ag asyncapi.yaml java-spring maven` 
`ag asyncapi.yaml java-spring gradle` 

I'm not sure if the commander library supports this type of workflow.  I'll try to find that out.

netoec84:
Fixes #8

### DIFF
--- a/cli.js
+++ b/cli.js
@@ -15,6 +15,10 @@ let asyncapi;
 let template;
 
 const parseOutput = dir => path.resolve(dir);
+const parseJSON = value => {
+  try { return JSON.parse(value); }
+  catch (e) { return {}; }
+};
 
 const showErrorAndExit = err => {
   console.error(red('Something went wrong:'));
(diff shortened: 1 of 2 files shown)


cli.js
```diff
@@ -15,6 +15,10 @@ let asyncapi;
 let template;
 
 const parseOutput = dir => path.resolve(dir);
+const parseJSON = value => {
+  try { return JSON.parse(value); }
+  catch (e) { return {}; }
```
fmvilas: I love the idea of "not breaking" and continue with normal execution using a default value of `{}`. What do you think about putting a warning here? I'm thinking in the case people make a mistake in their JSON string. They would not understand what's happening.

netoec84: You're right. Maybe adding something like:
```js
console.log(yellow(`Warning: values of --params will not be available in the templates, there was a error parsing: ${value}`));
console.log(magenta(e.message));
```
If you agree, I can update the pull request

fmvilas: Looks good to me 👍

PR #68, opened 2019-07-03
Dockerfile for generator

treeder:
Build with:

```sh
docker build -t asyncapi/generator .
```

Would be good to auto build and release this along with the npm releases.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -11,7 +11,5 @@ RUN npm install
 
 # copy code
 COPY . .
-# RUN chmod a+x /app/dockerboot.sh
 
-# CMD [ "node", "cli.js" ]
 ENTRYPOINT [ "/app/dockerboot.sh" ]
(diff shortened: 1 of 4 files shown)


Dockerfile
```diff
@@ -0,0 +1,17 @@
...
+
+# install dependencies
+COPY package*.json ./
+RUN npm install
+
+# copy code
+COPY . .
+# RUN chmod a+x /app/dockerboot.sh
+
+# CMD [ "node", "cli.js" ]
```
fmvilas: Should we get rid of this commented line and the one above?

treeder: Yes. :) Done.

treeder:
Grabbed the `asyncapi` org on Docker Hub too, happy to hand it over to if you give me your Docker Hub username.

fmvilas:
Awesome! My Docker Hub username is `fmvilas`. Very kind of you. Thanks!

Issue #69, opened 2019-07-09
use tags for structure


rv0:
Currently the "tags" system only add tags under each operation.
There's no further functionality as far as I can see.

Swagger also has tags, and uses them for structure. See the pet store example: https://petstore.swagger.io/#/ .. Pet, store, user, etc.. are tags and they provide structure to the document by grouping.

I'm currently documenting a rather big MQTT API and it's very hard to navigate through it at the moment.
The left sidebar is just one big (randomised order) list of summaries of operations. Ideally there would be a hierarchical (collapsible) structure 
```
tag
  - operations
tag2
  - operations
```

The content of the page could also add header titles with tags (see the above pet store example) and even better: have a table of contents / nav with all tags displayed above (anchor links)

fmvilas:
Thanks for opening the issue, @rv0. To clarify:

* We're talking about the HTML template, right?
* If we add a tags menu to the sidebar then the table of contents shouldn't be needed. Am I right?
* When you say hierarchical collapsible structure you mean 2 collapsible items like, for instance, Operations and Tags? Or you mean hierarchical by looking at the common beginning of a channel name?

rv0:
- yes, html template
- yes, tags in sidebar seems like nicest solution
- hierarchical with first level tags and 2nd lavel operations, like the small example above.. hierarchical with channel names would be very cool but seems outside of the scope of this FRQ

fmvilas:
I see. Just as a side note for the future implementer: the hierarchical organization on the sidebar could be driven by a special parameter for the HTML template. See #8.

For instance:

```
ag asyncapi.yaml html --params '{"sidebarOrganization":"hierarchical"}'
```

GianniCoene:
I'm currently working on this issue. So far I have been able to implement the functionality of tags in the sidebar.
The user can click the tag which will reveal all topics associated with that particular tag.

However I ran into some trouble when trying to require the special parameter for the HTML template.
In the code I check if the parameter is present using:
` {{#if params.sidebarOrganization}} ` as mentioned in the referenced issue.
But when I use the CLI to try and generate the docs using the command:
`ag asyncapi.yaml html --params '{"sidebarOrganization":"hierarchical"}'`
I will get a JSON parse error saying: Unexpected token ' in JSON at position 0.

I've tried the parameter in a couple of different notations but none of them seem to work. Any ideas?

fmvilas:
[...]

PR #78, opened 2019-08-12
Passing 'params' on from generateTemplateFile()

marcortw:
Passing 'params' on to use it with .generateTemplateFile() as a follow-up to issue #69. Also adding a short description and tags to the streetlights Example for demonstration purposes of the sidebarOrganization param.


fmvilas:
It's now on npm under version 0.8.2. Thanks!

PR #81, opened 2019-08-16
Migrate to Nunjucks

fmvilas:
While working on the new templates, I realized that we were very limited by Handlebars functionality and was hard to write even simple comparisons or simply call a function. Therefore, I decided to investigate other templating languages and found [Nunjucks](https://mozilla.github.io/nunjucks) (from Mozilla) and it's incredibly easier to write templates compared to Handlebars.

This PR makes the necessary changes to use Nunjucks instead of Handlebars and migrates all the templates.


fmvilas:
I'm moving forward and gonna merge this PR. Please, if you see something wrong, missing, or just don't agree with the changes, **don't hesitate to leave a comment after it's merged**. I'll be happy to address these questions afterward.

PR #83, opened 2019-08-18
Issue-76 Added missing channel description

### LINKED ISSUE #76: Missing description of channel
When a channel has a description, it is not generated and printed in HTML.

See the following demo specification and use https://playground.asyncapi.io/

---
```
asyncapi: '2.0.0-rc1'
id: 'urn:demo'
info:
  title: Demo
  version: '1.0'

channels:

  demo/channel:
    description: The description of demo channel.
    publish:
      summary: Channel summary
      message:
        $ref: '#/components/messages/demoMessage'

components:

  messages:

    demoMessage:
      name: demoMessage
      payload:
        $ref: '#/components/schemas/demoHeader'

  schemas:

    demoHeader:
      type: object
      properties:
        id:
          type: string
          description: The id of the header
        name:
          type: string
          description: The name of the header
```

grafik

WaleedAshraf:
Closes #76 

I have also added package `handlebars` as it was missing and I wasn't able to run the cli locally.

Also, currently we don't have channels and channel partials for HTML templates. Everything is being done in operation partial. We should have them like in markdown templates. This can be done later on.


package.json
```diff
@@ -32,6 +32,7 @@
     "asyncapi-parser": "^0.3.5",
     "commander": "^2.12.2",
     "fs.extra": "^1.3.2",
+    "handlebars": "^4.1.2",
```
fmvilas: You should not need Handlebars. We're using Nunjucks now.

templates/html/.partials/operation.html
```diff
@@ -17,6 +17,8 @@ <h3 class="font-mono text-base">
       {{ slicedString(channelName) }}
     </h3>
 
+    <div class="mt-4 mb-4 markdown">{{ channel.descriptionAsHTML | safe }}</div>
```
fmvilas: Channel description should be at channel level, not at operation level.

fmvilas: Sorry, just saw your comment in the description.

package.json
WaleedAshraf: 👍 it was still being used. Removed un-used code.

PR #88, opened 2019-08-21
Fix Docker docs

fmvilas:
FYI @treeder and @YBogomolov. I already pushed this image to Docker Hub. It's tagged with `latest` and `0.10.0`.

It removes the need for `asyncapi-node-codegen` as the current generator already supports Node.js.

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,7 +13,7 @@ npm install -g asyncapi-generator
 Or use all the commands below using Docker by prefixing them with:
 
 ```bash
-docker run --rm -it -v $PWD:/app -w /app asyncapi/generator [ag|anc] [COMMAND HERE]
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -13,7 +13,7 @@ npm install -g asyncapi-generator
 Or use all the commands below using Docker by prefixing them with:
 
 ```bash
-docker run --rm -it -v $PWD:/app -w /app asyncapi/generator [ag|anc] [COMMAND HERE]
+docker run --rm -it -v $PWD:/app -w /app asyncapi/generator ag [COMMAND HERE]
```
treeder: Could probably just have it call `ag` by default so user doesn't have to put it here.

fmvilas: You're right. That would be better.

PR #94, opened 2019-08-21
Move Message Examples Back into Right Column in HTML Template

### LINKED ISSUE #96: Channel with Pub and Sub Operations displays poorly in XL devices
Third times a charm...

I've already figured out the fix for this, but don't have time in the moment to submit a PR - so I'll at least capture my findings here and work on the PR later (or if someone else wants to pick this up).

When a Channel has both a Pub and Sub Operation, and your browser width is greater than (approximately) 1248 pixels, the Pub and Sub Operations are displayed in the same row and share the remaining row width (with some overflow depending on how small your resolution is). This can be seen here (intentionally showing the x-overflow issue as well):

Screen Shot 2019-08-22 at 08 52 53

This can be fixed by moving each Operation's HTML into a separate `<div class="responsive-container">`, whereas currently both Pub and Sub Operations are nested under the same responsive-container div.

Picture for clarity. Just need to insert `</div><div class="responsive-container">` above the highlighted row to fix the issue.
Screen Shot 2019-08-22 at 08 55 15

After applying this fix (in dev console of browser):
Screen Shot 2019-08-22 at 08 49 48

jjanusch:
Fixes #96 

This fixes 3 bugs that were introduced in f411b4d during the migration to Nunjuck:

1. It forced the Message Examples into the center column rather than in the right column
2. If there were multiple messages, it would lay them out horizontally rather than vertically. 
3. If a channel had both a publish and a subscribe, it would lay them out horizontally rather than vertically. 

This patch fixes this and also reduces some of the padding that was caused by the fix.

Current/Broken:

Screen Shot 2019-08-21 at 3 41 21 PM

After patch:

Screen Shot 2019-08-21 at 3 59 22 PM

### DIFF
--- a/templates/html/.partials/operation.html
+++ b/templates/html/.partials/operation.html
@@ -4,36 +4,41 @@
 
 {% macro operation(operation, operationType, channelName, channel) %}
 
-<a name="operation-{{operationType}}-{{channelName}}"></a>
-<div class="center-block p-8">
-  <div class="operation pt-8 pb-8">
-    <h3 class="font-mono text-base">
-      {% if operationType === 'publish' %}
-      <span class="font-mono border border-blue text-blue uppercase p-1 rounded" title="Publish">Pub</span>
-      {% endif %}
-      {% if operationType === 'subscribe' %}
-      <span class="font-mono border border-green-dark text-green-dark uppercase p-1 rounded" title="Subscribe">Sub</span>
-      {% endif %}
-      {{ slicedString(channelName) }}
-    </h3>
+<div class="responsive-container">
+  <a name="operation-{{operationType}}-{{channelName}}"></a>
+  <div class="center-block p-8">
+    <div class="operation pt-8 pb-8">
+      <h3 class="font-mono text-base">
+        {% if operationType === 'publish' %}
+        <span class="font-mono border border-blue text-blue uppercase p-1 rounded" title="Publish">Pub</span>
+        {% endif %}
+        {% if operationType === 'subscribe' %}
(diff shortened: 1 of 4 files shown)


templates/html/.partials/operation.html
```diff
@@ -4,36 +4,41 @@
...
+
+      {{ tags(operation.tags()) }}
+    </div>
   </div>
 </div>
 
+{% if operation.hasMultipleMessages() %}
+  {% for msg in operation.messages() %}
+    <div class="responsive-container message-container">
+      {{ message(msg, showIndex=true, index=loop.index, open=false, prefixHtml='<p>Accepts <strong>one of</strong> the following messages:</p>') }}
```
fmvilas: This prefixHtml seems out of place. It's not the message who accepts **one of** the messages, it's the operation, therefore I think this logic shouldn't be delegated to the message template.

Also, seems a bit hacky to pass HTML as a param, don't you think so? I appreciate the effort you did here to DRY things but maybe making it a little bit less generic will work better. Why don't you simply put this HTML before calling the `message` template?

templates/html/css/main.css
```diff
@@ -109,6 +109,10 @@ h1, h2, h3, h4, h5, h6 {
   z-index: 1;
 }
 
+.message-container > .p-8 {
```
fmvilas: That's a hack. The class `p-8` actually means to set the padding of all sides the same size (2 rem). If you want to change padding go to where `message-container` class is applied and add the following classes: `py-0 px-8`.

fmvilas:
Closing as this was fixed at #113. Thanks anyway, @jjanusch.

PR #95, opened 2019-08-21
Inline all assets in HTML template, allow for single file generation

jjanusch:
This updates the HTML template so that third party libraries are pulled via CDN and the main.css and main.js files are inlined in header/footer. This makes standard generation of the template produce just 1 file, compared to 6 files and 2 folders. This was done to make it easier for deploying via CI


fmvilas:
This is great but maybe it should be behind a template param/flag because Internet connection is not always guaranteed or reliable. For instance, you're in a plane or your CI doesn't have access to Internet (some big companies don't allow it to prevent code leakages).

Also, take care with parsing main.js as a template. It may contain (now or in the future) characters that will match the `{{variable}}` or `{%block%}` syntax, making the render process fail or corrupting the file.

jjanusch:
> This is great but maybe it should be behind a template param/flag because Internet connection is not always guaranteed or reliable. For instance, you're in a plane or your CI doesn't have access to Internet (some big companies don't allow it to prevent code leakages).

That's a good call and actually the reason I didn't include it in #94, despite doing both simultaneously. I'll see if I can figure out a way to do it via parameters, though I suspect it will require a change to /lib/generator.js or cli.js to pull off; as far as I can tell, there's no way to ignore those files. So I could have it come through inline, but it would still spit out the JS and CSS folders. I'll have to do some digging.

> Also, take care with parsing main.js as a template. It may contain (now or in the future) characters that will match the {{variable}} or {%block%} syntax, making the render process fail or corrupting the file.

I had not even considered that. I'll see if there is a way to load the JS without rendering it. Not sure it is, but I could probably accomplish it via a filter if there aren't any other options. I suspect a syntax like `{{ 'path/to/file.ext' | loadraw }}` that returns the contents of the file would work, though that is definitely an abuse of the Nunjuck's filtering system. I'll see if there's anyway to do it without resorting to that.

Thanks for the feedback.

jjanusch:
After a quick dive into the Nunjucks source, it appears I can add functions via the `addGlobal` method. So I can add a simple load in a file in .filters and hopefully safely load the JS and CSS inline.

fmvilas:
Awesome!

In case it helps: the API has a way to [specify an entry point](https://github.com/asyncapi/generator/blob/master/API.md#new-generatortemplatename-targetdir-options) for the template rendering process, instead of rendering all the files. It would be awesome to add `entrypoint` as an option to the CLI.

This way, we could do something like this:

```
ag html asyncapi.yaml -o output --entrypoint index.html
```

That would only generate "index.html", leaving main.js and main.css untouched. Then on the template, you can check for the presence of "entrypoint". If it's there, you raw include main.js and main.css.

Does it make sense?

jjanusch:
Those commits revert the original changes and do the following:

- Introduces `--entrypoint`/`-e` that allows for setting the entrypoint from the CLI
- Fixes a bug with the entrypoint where it would generate the file but never actually write it anywhere
- Adds a new param to the HTML template called `inline` which inlines all of the CSS and JS.
- Passing `--entrypoint` automatically inlines the CSS and JS. If you want to use both `--entrypoint` and not inline the assets, you can pass `--param inline=0` and they load them remotely like normal
- Removes CDN dependencies introduced in my first commit

Hopefully that ticks all the boxes. If it doesn't, let me know and I will make further changes.

jjanusch:
As an aside, I also added a new file called `.filters/functions.js` where I added the global function for loading files and spitting out the content without rendering. I felt that was more appropriate than including it in `.filters/all.js`, which seemed focused on filters.

templates/html/index.html
```diff
@@ -4,9 +4,18 @@
   <meta charset="UTF-8">
   <title>{{asyncapi.info().title()}} {{asyncapi.info().version()}} documentation</title>
   <meta name="viewport" content="width=device-width, initial-scale=1.0">
-  <link href="css/tailwind.min.css" rel="stylesheet">
-  <link rel="stylesheet" href="css/atom-one-dark.min.css" />
-  <link rel="stylesheet" href="css/main.css" />
+
```
fmvilas: I've seen you're changing this file permissions to 755. Is there a reason for this? Wasn't it on purpose?

PR #97, opened 2019-08-22
Fix: Place pub and sub operations in their own responsive-containers.

joshg2k:
Placing pub and sub operations in their own responsive-containers causes them to wrap to their own rows in XL device widths.


jjanusch:
I fixed this in #94 as well.

fmvilas:
Yes, I'm closing this one as #94 already solves the problem. Thanks!

PR #117, opened 2019-09-05
Fixed isExpandable filter

### LINKED ISSUE #116: Parameter showing as expandable when there's nothing to expand
I tried to fix this one yesterday but came up short, maybe someone here can help:

The chevron to expand a parameter shows even when there's nothing to expand.

I'm assuming the problem is either around line 106 of templates/html/.partials/schema-prop.html or in the isExpanded filter on line 13 of templates/html/.filters/all.html.

Help would be appreciated! Thanks.

image
image

joaoritter:
Fixes #116 

- removed additionalItems, which wasn't actually a property. 
- checked the result of the function instead of the existence of a function to determine expandability.

### DIFF
--- a/templates/html/.filters/all.js
+++ b/templates/html/.filters/all.js
@@ -14,13 +14,12 @@ module.exports = ({ Nunjucks, Markdown, OpenAPISampler }) => {
     if (
       obj.type === 'object' ||
       obj.type === 'array' ||
-      obj.oneOf ||
-      obj.anyOf ||
-      obj.items ||
-      obj.additionalItems ||
-      obj.properties ||
-      obj.additionalProperties ||
-      obj.patternProperties
+      obj.oneOf() ||
+      obj.anyOf() ||
+      obj.items() ||
+      obj.properties() ||
+      obj.additionalProperties() ||
+      obj.patternProperties()
     ) return true;
 
     return false;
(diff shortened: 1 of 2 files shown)


joaoritter:
Fix to [issue 116](https://github.com/asyncapi/generator/issues/116)

templates/html/.filters/all.js
```diff
@@ -14,13 +14,12 @@ module.exports = ({ Nunjucks, Markdown, OpenAPISampler }) => {
     if (
       obj.type === 'object' ||
       obj.type === 'array' ||
-      obj.oneOf ||
-      obj.anyOf ||
-      obj.items ||
-      obj.additionalItems ||
```
fmvilas: This should not be removed. It's a valid property. Check https://json-schema.org/latest/json-schema-validation.html#rfc.section.6.4.2.

That said, it's missing on the JS parser and we should add it there. I opened an issue: https://github.com/asyncapi/parser-js/issues/25.

fmvilas: JS parser is now supporting `additionalItems`. You should just update the `asyncapi-parser` package to its latest version.

fmvilas:
Had to fix it urgently in #120 because it's also failing on playground.asyncapi.io. I'm closing this PR but thanks a lot for pointing out the solution 🙌

joaoritter:
Nice work!

fmvilas:
You did it! Thanks!

PR #122, opened 2019-09-06
Update schema-prop.html

joaoritter:


templates/html/.partials/schema-prop.html
```diff
@@ -1,6 +1,6 @@
 {% from "./type.html" import type %}
 
-{% macro schemaProp(prop, propName, open=false, root=fase, odd=false, specialName=false, required=false) %}
+{% macro schemaProp(prop, propName, open=false, root=false, odd=false, specialName=false, required=false) %}
```
fmvilas: What? Isn't it `fase`? :trollface:

PR #130, opened 2019-09-13
AVRO template

rogelio-o:
Template to generate AVRO schemas from AsyncAPI document.

Note: there are two schema properties that are not defined in AsyncAPI specification but that are useful to generate the AVRO schemas, namespace and defined_type (for company defined schemas, for example).

A more complex AsyncAPI document will be added to test folder in order to test the template.

### DIFF
--- a/lib/generator.js
+++ b/lib/generator.js
@@ -302,6 +302,11 @@ class Generator {
             await this.generateChannelFiles(asyncapiDocument, stats.name, root);
             const templateFilePath = path.relative(this.templateDir, path.resolve(root, stats.name));
             fs.unlink(path.resolve(this.targetDir, templateFilePath), next);
+          } else if (stats.name.includes('$$message$$')) {
+            // this file should be handled for each in asyncapi.components.messages
+            await this.generateMessageFiles(asyncapiDocument, stats.name, root);
+            const templateFilePath = path.relative(this.templateDir, path.resolve(root, stats.name));
+            fs.unlink(path.resolve(this.targetDir, templateFilePath), next);
           } else {
             const filePath = path.relative(this.templateDir, path.resolve(root, stats.name));
             if (!shouldIgnoreFile(filePath)) {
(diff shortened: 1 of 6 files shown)


rogelio-o:
Command to test the template: `./cli.js test/docs/store.yml avro`

fmvilas:
Hi @rogelio-o. Thanks for this PR. I appreciate the effort but this tool should not deal with something that's not JSON Schema unless necessary. For that purpose, we have the schema parsers on the AsyncAPI parser: https://github.com/asyncapi/parser-js/tree/master/lib/schema-parsers. You may want to create an Avro schema parser there, i.e., transforming Avro into JSON Schema. This way, the rest of the tooling doesn't have to handle one more schema language.

On the other hand, I think the part that's related to generating files for each message is interesting and maybe you can reorient your PR to only this. Thanks!

rogelio-o:
Is the schema-parser only for transforming from a schema to AsyncAPI? Or is it bidirectional?

What we would like is to have all the documentation in AsyncAPI and use the generator (or schema) to generate an intermediate schema. Why? Because the tool to generate code from Avro files exists (like the tools to generate code from XSD files). So our source of trust would be the AsyncAPI schema and the “intermediate” schemas would be use as a step to generate the code. So, would the generation of those schemas (multiple files: xsd, avsc,...) be in the parser?

fmvilas:
Oh! I see your point now. Yes, that should be here then. If you're going to translate JSON Schema Draft 07 (AsyncAPI schemas are a superset of it) to Avro then you may want to have a look at this: https://github.com/supermodel/supermodel/blob/master/packages/lib/src/convert/Avro.ts. And also, why not, have a look at [Supermodel.io](supermodel.io) from @zdne.

zdne:
Hi @rogelio-o, 
 FWIW, I am working on a tool that facilitates work with JSON Schemas, and also provides means to reuse JSON Schemas called [Supermodel](http://supermodel.io). 

My idea is to define models once in JSON schema and then reuse in different places like AsyncAPI and OpenAPI. 

The [Supermodel CLI](https://github.com/supermodel/supermodel/tree/master/packages/cli#readme) can already convert JSON Schemas into AVRO (and OpenAPI and GraphQL). 

The idea is to write something like this: 

```
asyncapi: 2.0.0
info:
  title: Starwars Application
  version: '0.1.0'
channels:
  film:
    publish:
      message:
        payload:
          $ref: 'http://supermodel.io/starwars/async/Film'
```             

Enjoy all the goodies that AsyncAPI provides and when it comes to the need for avro schema simply convert the `http://supermodel.io/starwars/async/Film` like so: 

```
$ supermodel install http://supermodel.io/starwars
$ supermodel schema avro ./supermodel/starwars/async/Film
```
 
Would this be helpful?

rogelio-o:
I see. It looks interesting! I'll take a look and see if we can use it. One thing I'm concern about is that it looks like the model have to be stored remotely, on supermodel servers. This could be a problem for some companies, the fact of having their domain in a remote server.

Issue #136, opened 2019-10-02
HTML generator does not render specification extensions 


ethanballinger:
Usage is documented here: https://www.asyncapi.com/docs/specifications/2.0.0/#specificationExtensions

When providing "x-..." extensions, these are not rendered in the HTML index file.

fmvilas:
How do you think they could be rendered? Specification extensions can have any content and can mean anything.

ethanballinger:
So in swagger, it renders it like an attribute to the property. 

So 

`      name:
        type: "string"
        example: "doggie"
        x-pi: false
        x-pci: false  `

Would get generated as: 

name* | string
example: doggie
x-pi: false
x-pci: false

fmvilas:
Ok, I see what you mean. Fair point.

PR #139, opened 2019-10-10
HTML template: Facilitate Specification Extensions

scottpopplewell:
Fixes https://github.com/asyncapi/generator/issues/136

### DIFF
--- a/templates/html/.filters/all.js
+++ b/templates/html/.filters/all.js
@@ -21,7 +21,7 @@ module.exports = ({ Nunjucks, Markdown, OpenAPISampler }) => {
       obj.additionalItems() ||
       (obj.properties() && Object.keys(obj.properties()).length) ||
       obj.additionalProperties() ||
-      obj.extensions() ||
+      (obj.extensions() && Object.keys(obj.extensions()).length) ||
       obj.patternProperties()
     ) return true;
 
--- a/templates/html/.filters/all.js
+++ b/templates/html/.filters/all.js
@@ -36,10 +36,6 @@ module.exports = ({ Nunjucks, Markdown, OpenAPISampler }) => {
     return typeof obj === 'object' && obj !== null;
   });
 
-  Nunjucks.addFilter('prettyPrint', (obj) => {
-    return JSON.stringify(obj, null, 4);
-  });
-
   Nunjucks.addFilter('contains', (array, element) => {
     if (!array || !Array.isArray(array)) return false;
     return array.includes(element);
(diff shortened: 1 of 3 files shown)


templates/html/.partials/schema-prop.html
```diff
@@ -89,6 +89,9 @@
             {% endfor %}
           </div>
         {% endif %}
+        {% for extensionName, extensionValue in prop.extensions() %}
+            <div class="text-sm markdown">{{ extensionName }}: {{ extensionValue }}</div>
```
fmvilas: `extensionValue` can be huge here depending on the extension. I wonder if we should think about another way of visualizing it. Also, if it's a JSON object, doesn't it render `[Object object]`?

scottpopplewell:
@fmvilas It is finished. Ready to be merged.

templates/html/.partials/schema-prop.html
```diff
@@ -108,6 +108,34 @@
...
+                    <span class="text-sm" style="word-break: break-word;">{{ extensionName }}</span>
+                    <svg class="expand" version="1.1" viewBox="0 0 24 24" x="0"
+                      xmlns="http://www.w3.org/2000/svg" y="0">
+                      <polygon points="17.3 8.3 12 13.6 6.7 8.3 5.3 9.7 12 16.4 18.7 9.7 "></polygon>
+                    </svg>
+                  </div>
+                </div>
+                <div class="children">
+                  <div class="{% if odd %}bg-grey-lighter{% else %}bg-grey-lightest{% endif %} {% if not root %}pl-8 pr-8{% endif %} rounded">
+                    <pre class="text-sm">{{ extensionValue | prettyPrint }}</pre>
```
fmvilas: You don't need a custom filter for "pretty printing". Instead use `dump(2)`.

```suggestion
                    <pre class="text-sm">{{ extensionValue | dump(2) }}</pre>
```

templates/html/.filters/all.js
```diff
@@ -30,6 +31,14 @@ module.exports = ({ Nunjucks, Markdown, OpenAPISampler }) => {
...
     return Array.isArray(arr);
   });
+  
+  Nunjucks.addFilter('isObject', (obj) => {
+    return typeof obj === 'object' && obj !== null;
+  });
+
+  Nunjucks.addFilter('prettyPrint', (obj) => {
+    return JSON.stringify(obj, null, 4);
+  });
```
fmvilas: As pointed out below, you'll not need this.

scottpopplewell: Beautiful. Thanks for that. It's a great change. I've updated all.js and schema-prop.html

Issue #145, opened 2019-10-29
Feature request - Improved logging for debugging


jonaslagoni:
So fooling around with the generator and the nunjucks template i found it difficult to pinpoint in what file i did something wrong. At the moment I have just added a logging at L530: https://github.com/asyncapi/generator/blob/802fdb1d93d1eec400186ebd29577c105b3df047/lib/generator.js#L528-L535

Saying which file it failed rendering. I cant imagine I will be the only one with this dilemma of having changed multiple files but no idea which one is at fault. Was giving the default nunjucks error:
```
Something went wrong:
Template render error: (unknown path) [Line 15, Column 13]
  unknown block tag: endfor
    at Object._prettifyError (/home/lagoni/Documents/SmartGit/Blackhawk rust/async api generator/node_modules/nunjucks/src/lib.js:36:11)
    at Template.render (/home/lagoni/Documents/SmartGit/Blackhawk rust/async api generator/node_modules/nunjucks/src/environment.js:542:21)
    at Environment.renderString (/home/lagoni/Documents/SmartGit/Blackhawk rust/async api generator/node_modules/nunjucks/src/environment.js:380:17)
    at Promise (/home/lagoni/Documents/SmartGit/Blackhawk rust/async api generator/lib/generator.js:507:21)
    at new Promise (<anonymous>)
    at Generator.renderString (/home/lagoni/Documents/SmartGit/Blackhawk rust/async api generator/lib/generator.js:506:12)
    at Generator.renderFile (/home/lagoni/Documents/SmartGit/Blackhawk rust/async api generator/lib/generator.js:532:19)
```

Is this something we or I could implement? Not quite sure to which extend it should be implemented, maybe a debugging flag? let me know :)

fmvilas:
This is definitely something that bothers me too. The key is here:

> Template render error: **(unknown path)** [Line 15, Column 13]

For some reason, Nunjucks is not able to tell you in which file/line it fails when using helpers. I'd be super happy if you could investigate this further and come up with a solution :)

jonaslagoni:
So based on this issue and [comment and resolution](https://github.com/mozilla/nunjucks/issues/1039#issuecomment-394926499). I was able to implement it, doing a PR now.

fmvilas:
Thanks for finding and fixing the problem. You rock!

PR #146, opened 2019-10-29
Fixing unknown path

jonaslagoni:
Fixed the issue with unknown path for nunjucks for issue #145


fmvilas:
It's now on npm. Thanks!

PR #149, opened 2019-11-01
Fix settings.gradle

jmaister:
Fix rootProject.name setup.


fmvilas:
Thanks, @jmaister. Actually, this template is completely broken because it's not yet migrated to AsyncAPI 2.0 and the new parser. I could merge this PR but the Java Spring needs to be completely rebuilt from the ground up. I'd be happy if you want to do it but just merging this PR would not change much. Hope you understand. And thanks again for your time!

PR #150, opened 2019-11-03
Fixed issue with tags not being found.

### LINKED ISSUE #148: html template using byTags sidebarOrganization does not look through all the tags
While solving #141 I saw that only the first tag are checked:
https://github.com/asyncapi/generator/blob/f1c196a8cb12a667d4dbae5b01af6b538ca184d9/templates/html/.partials/sidebar.html#L47
Meaning that if I define operations like:
```json
...
"tags": [{
	"name": "user1",
	"description": "User-related messages"
},
{
	"name": "user",
	"description": "User-related messages"
}
],
...
```
and the tags like:
```json
...
"tags": [
{
	"name": "user",
	"description": "User-related messages"
}
],
...
```
the operation would not be listed under the tags, guess this is not intended?

jonaslagoni:
Fixes #148

### DIFF
--- a/templates/html/.filters/all.js
+++ b/templates/html/.filters/all.js
@@ -43,7 +43,7 @@ module.exports = ({ Nunjucks, Markdown, OpenAPISampler }) => {
       tagsToCheck = [tagsToCheck];
     }
 
-    //Check that pubsub does not contain one of the tags to check.
+    //Check if pubsub contain one of the tags to check.
     let check = (tag) => {
       let found = false;
       for(let tagToCheckIndex in tagsToCheck){
@@ -55,10 +55,9 @@ module.exports = ({ Nunjucks, Markdown, OpenAPISampler }) => {
       return found;
     };
 
-    //Ensure pubsub tags are checked for the group tags
-    let pubsubContainsNoTag = object._json.tags ? object._json.tags.find(check) != null : false;
-    if(pubsubContainsNoTag === true) return true;
-    return false;
+    //Ensure tags are checked for the group tags
+    let containTags = object._json.tags ? object._json.tags.find(check) != null : false;
+    return containTags;
   });
 
   /**
--- a/templates/html/.filters/all.js
+++ b/templates/html/.filters/all.js
@@ -30,25 +30,26 @@ module.exports = ({ Nunjucks, Markdown, OpenAPISampler }) => {
   /**
    * Check if there is a channel which does not have one of the tags specified.
    */
-	Nunjucks.addFilter('containTags', (object, tagsToCheck) => {
-		if (!object) {
+  Nunjucks.addFilter('containTags', (object, tagsToCheck) => {
+    if (!object) {
       throw new Error("object for containsTag was not provided?");
     }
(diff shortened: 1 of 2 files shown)


templates/html/.partials/sidebar.html
```diff
@@ -44,7 +44,7 @@ <h2 class="text-xs uppercase text-grey mt-10 mb-4 font-thin">Operations</h2>
         </div>
         <div class="children">
           {% for channelName, channel in asyncapi.channels() %}
-             {% if channel.publish() and channel.publish().hasTags() and categorytags.name() === channel.publish().tags()[0].name() %}
+             {% if channel.publish() and channel.publish().hasTags() and channel.publish() | containTags(categorytags) %}
```
fmvilas: I think `channel.hasPublish` would be more appropriate here, don't you think so?

```suggestion
             {% if channel.publish() and channel.publish().hasTags() and channel.hasPublish() | containTags(categorytags) %}
```

templates/html/.partials/sidebar.html
```diff
@@ -61,7 +61,7 @@ <h2 class="text-xs uppercase text-grey mt-10 mb-4 font-thin">Operations</h2>
               </a>
             {% endif %}
 
-            {% if channel.subscribe() and channel.subscribe().hasTags() and categorytags.name() === channel.subscribe().tags()[0].name() %}
+            {% if channel.subscribe() and channel.subscribe().hasTags() and channel.subscribe() | containTags(categorytags) %}
```
fmvilas:
```suggestion
            {% if channel.subscribe() and channel.subscribe().hasTags() and channel.hasSubscribe() | containTags(categorytags) %}
```

templates/html/.filters/all.js
```diff
@@ -34,8 +34,16 @@ module.exports = ({ Nunjucks, Markdown, OpenAPISampler }) => {
 		if (!object) {
       throw new Error("object for containsTag was not provided?");
     }
+		if (!tagsToCheck) {
```
fmvilas: I'm seeing weird indentation problems here and in some lines above. I recommend you to install [EditorConfig](https://editorconfig.org/) in your fav code editor. Installing [ESLint](https://eslint.org/) will also help.

templates/html/.partials/sidebar.html
jonaslagoni: I am using the publish operation to figure out if it contains the current tag in the loop, since the `hasPublish()` returns a boolean there would be tags for it to find. Unless this behaves differently then I think it does ^^

templates/html/.partials/sidebar.html
jonaslagoni: Same as before.

templates/html/.filters/all.js
jonaslagoni: Have to double check, but thought I had it enabled. Fixed.

templates/html/.partials/sidebar.html
fmvilas: Oh, you're right. Nevermind.

PR #144, opened 2019-10-28
Feature/refactoring separated files code

jonaslagoni:
Refactoring the code to reduce duplicated code for how files are separated, as discussed in PR #142. The code have been refactored to not change the underlying `renderFile` method. This PR expect #142 to be merged before this one since some code conflicts might occur.

PR #154, opened 2019-11-05
Refactoring file separations

jonaslagoni:
Follow up from PR #144

### DIFF
--- a/lib/generator.js
+++ b/lib/generator.js
@@ -371,7 +371,7 @@ class Generator {
    * @param  {String} baseDir Base directory of the given file name.
    * @returns {Promise}
    */
-  generateSeperateFiles(asyncapiDocument, array, template, fileName, baseDir) {
+  generateSeparateFiles(asyncapiDocument, array, template, fileName, baseDir) {
     const promises = [];
 
     Object.keys(array).forEach((name) => {
--- a/lib/generator.js
+++ b/lib/generator.js
@@ -121,7 +121,7 @@ class Generator {
           await this.launchHook('generate:after');
           return;
         } else if (this.output === 'string') {
-          return this.renderString(asyncapiDocument, await readFile(entrypointPath, { encoding: 'utf8' }), entrypointPath);
+          return this.renderString(asyncapiDocument, await readFile(entrypointPath, { encoding: 'utf8' }));
         }
       }
       await this.validateTemplateConfig(asyncapiDocument);
@@ -296,20 +296,23 @@ class Generator {
       });
 
       walker.on('file', async (root, stats, next) => {
+        let didSeperateIntoFiles = false;
         try {
           if (stats.name.includes('$$channel$$')) {
             // this file should be handled for each in asyncapi.channels
-            await this.generateChannelFiles(asyncapiDocument, stats.name, root);
-            const templateFilePath = path.relative(this.templateDir, path.resolve(root, stats.name));
-            fs.unlink(path.resolve(this.targetDir, templateFilePath), next);
+            await this.generateSeperateFiles(asyncapiDocument, asyncapiDocument.channels(), 'channel', stats.name, root);
+            didSeperateIntoFiles = true;
           } else if (stats.name.includes('$$message$$')) {
             // this file should be handled for each in asyncapi.components.messages
-            await this.generateMessageFiles(asyncapiDocument, stats.name, root);
-            const templateFilePath = path.relative(this.templateDir, path.resolve(root, stats.name));
-            fs.unlink(path.resolve(this.targetDir, templateFilePath), next);
+            await this.generateSeperateFiles(asyncapiDocument, asyncapiDocument.components().messages(), 'message', stats.name, root);
+            didSeperateIntoFiles = true;
           } else if (stats.name.includes('$$securityScheme$$')) {
             // this file should be handled for each in asyncapi.components.securityScheme
-            await this.generateSecuritySchemeFiles(asyncapiDocument, stats.name, root);
+            await this.generateSeperateFiles(asyncapiDocument, asyncapiDocument.components().securitySchemes(), 'securityScheme', stats.name, root);
+            didSeperateIntoFiles = true;
+          }
+
+          if (didSeperateIntoFiles) {
             const templateFilePath = path.relative(this.templateDir, path.resolve(root, stats.name));
             fs.unlink(path.resolve(this.targetDir, templateFilePath), next);
           } else {
@@ -320,6 +323,7 @@ class Generator {
             }
             next();
           }
+
         } catch (e) {
           reject(e);
         }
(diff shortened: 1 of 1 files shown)


jonaslagoni:
Because we added Schemas I thought it wouldn't hurt adding the rest of the components that could be separated  :smile: not sure if message traits were needed though?

lib/generator.js
```diff
@@ -288,6 +288,15 @@ class Generator {
...
    */
   generateDirectoryStructure(asyncapiDocument) {
+    const fileNamesForSeperation = {
+      channel: asyncapiDocument.channels(),
+      message: asyncapiDocument.components().messages(),
+      securityScheme: asyncapiDocument.components().securitySchemes(),
+      scheme: asyncapiDocument.components().schemas(),
+      parameter: asyncapiDocument.components().parameters(),
+      messageTrait: asyncapiDocument.components().messageTraits()
+    };
```
fmvilas: I understand the temptation of adding more "components" but we should be careful. Let's add things as people request them instead of adding them because they are easy. Think that we'll have to maintain it and we may end up maintaining something that is not useful to anyone. And then removing it will mean that we'd have to release a major version :)

lib/generator.js
```diff
@@ -288,6 +288,15 @@ class Generator {
    * @return {Promise}
    */
   generateDirectoryStructure(asyncapiDocument) {
+    const fileNamesForSeperation = {
+      channel: asyncapiDocument.channels(),
+      message: asyncapiDocument.components().messages(),
+      securityScheme: asyncapiDocument.components().securitySchemes(),
+      scheme: asyncapiDocument.components().schemas(),
```
fmvilas: Let's use `schema` instead of `scheme`. We're already using `scheme` for "security schemes" and it can be confusing. Regarding usefulness, I actually find it useful for generating type definitions for typed languages.

lib/generator.js
```diff
@@ -121,7 +121,7 @@ class Generator {
           await this.launchHook('generate:after');
           return;
         } else if (this.output === 'string') {
-          return this.renderString(asyncapiDocument, await readFile(entrypointPath, { encoding: 'utf8' }), entrypointPath);
+          return this.renderString(asyncapiDocument, await readFile(entrypointPath, { encoding: 'utf8' }));
```
fmvilas: It's important we keep `entrypointPath` there, remember? https://github.com/asyncapi/generator/pull/146

lib/generator.js
```diff
@@ -297,22 +306,22 @@ class Generator {
...
+                const templateFilePath = path.relative(this.templateDir, path.resolve(root, stats.name));
+                fs.unlink(path.resolve(this.targetDir, templateFilePath), next);
+                wasSeparated = true;
+                //The filename can only contain 1 specifier (message, scheme etc)
+                break;
+              }
+            }
+          }
+          // If it was not separated process normally
+          if (!wasSeparated) {
```
fmvilas: Minor comment for the block above and some changes below too: you're using "seperation" and "separation" alternatively :)

lib/generator.js
jonaslagoni: Yes that's true, but on the other hand not having the feature right from the start might be frustrating to some if they use it for the first time, and since it is using the same logic being used it might be better off just adding them now 😄 But I can see both points of view,  let me know if I should delete it

lib/generator.js
jonaslagoni: Yep agreed.

lib/generator.js
jonaslagoni: Yep absolutely right, no idea why this was overwritten 🤔

lib/generator.js
jonaslagoni: Yep, spelling mistake 😄

lib/generator.js
fmvilas: Hehe! That's true. If you can think of a good use case for message traits to be in a separate, go for it. Otherwise, let's delete it :)

lib/generator.js
```diff
@@ -121,7 +121,7 @@ class Generator {
           await this.launchHook('generate:after');
           return;
         } else if (this.output === 'string') {
-          return this.renderString(asyncapiDocument, await readFile(entrypointPath, { encoding: 'utf8' }), entrypointPath);
+          return this.renderString(asyncapiDocument, await readFile(entrypointPath, { encoding: 'utf8' }, entrypointPath));
```
fmvilas:
```suggestion
          return this.renderString(asyncapiDocument, await readFile(entrypointPath, { encoding: 'utf8' }), entrypointPath);
```

😄

lib/generator.js
```diff
@@ -297,22 +306,22 @@ class Generator {
...
-            await this.generateSecuritySchemeFiles(asyncapiDocument, stats.name, root);
-            const templateFilePath = path.relative(this.templateDir, path.resolve(root, stats.name));
-            fs.unlink(path.resolve(this.targetDir, templateFilePath), next);
-          } else {
+          // Check if the filename dictates it should be separated
+          let wasSeparated = false;
+          for (const prop in fileNamesForSeparation) {
+            if (Object.prototype.hasOwnProperty.call(fileNamesForSeparation, prop)) {
+              if (stats.name.includes(`$$${prop}$$`)) {
+                await this.generateSeperateFiles(asyncapiDocument, fileNamesForSeparation[prop], prop, stats.name, root);
```
fmvilas:
```suggestion
                await this.generateSeparateFiles(asyncapiDocument, fileNamesForSeparation[prop], prop, stats.name, root);
```

lib/generator.js
```diff
@@ -352,146 +361,53 @@ class Generator {
...
    * @param  {AsyncAPIDocument} asyncapiDocument AsyncAPI document to use as the source.
-   * @param  {String} fileName Name of the file to generate for each channel.
+   * @param  {Array} array The components/channels to generate the separeted files for.
+   * @param  {String} template The template filename to replace.
+   * @param  {String} fileName Name of the file to generate for each security schema.
    * @param  {String} baseDir Base directory of the given file name.
    * @returns {Promise}
    */
-  async generateChannelFiles(asyncapiDocument, fileName, baseDir) {
+  generateSeperateFiles(asyncapiDocument, array, template, fileName, baseDir) {
```
fmvilas:
```suggestion
  generateSeparateFiles(asyncapiDocument, array, template, fileName, baseDir) {
```

lib/generator.js
```diff
@@ -352,146 +361,53 @@ class Generator {
...
     const promises = [];
 
-    asyncapiDocument.channelNames().forEach((channelName) => {
-      promises.push(this.generateChannelFile(asyncapiDocument, channelName, fileName, baseDir));
+    Object.keys(array).forEach((name) => {
+      const component = array[name];
+      promises.push(this.generateSeperateFile(asyncapiDocument, name, component, template, fileName, baseDir));
```
fmvilas:
```suggestion
      promises.push(this.generateSeparateFile(asyncapiDocument, name, component, template, fileName, baseDir));
```

lib/generator.js
```diff
@@ -352,146 +361,53 @@ class Generator {
...
-   *
-   * @private
-   * @param {AsyncAPIDocument} asyncapiDocument AsyncAPI document to use as the source.
-   * @param {String} securitySchemeName Name of the security schema to generate.
-   * @param {String} fileName Name of the file to generate for each security schema.
-   * @param {String} baseDir Base directory of the given file name.
-   * @returns {Promise}
-   */
-  async generateSecuritySchemeFile(asyncapiDocument, securitySchemeName, fileName, baseDir) {
+  async generateSeperateFile(asyncapiDocument, name, component, template, fileName, baseDir) {
```
fmvilas:
```suggestion
  async generateSeparateFile(asyncapiDocument, name, component, template, fileName, baseDir) {
```

lib/generator.js
jonaslagoni: Hmm, one use case I can think of is inheritance, multiple messages using the same trait and inherits it instead of having duplicated code (same with operation traits). However not quite sure people will care if they are generated anyway by the template 🤔

lib/generator.js
jonaslagoni: 🤦‍♂

lib/generator.js
jonaslagoni: Just gonna remove it for now, easy to add later :)

PR #156, opened 2019-11-05
Added parameter documentation

### LINKED ISSUE #100: Channel Parameters are not rendered in preview pane
This is less of an issue if your parameters are all contained in the path of the channel, but unfortunately (as far as I know) AWS API Gateway does not allow you to specify path parameters in their WebSocket API definitions. Therefore I have to use querystring parameters to pass parameters upon establishing a websocket connection to a particular channel.

jonaslagoni:
Partly fixes #100, still need to display query parameters.


jonaslagoni:
image
Not sure if the parameters should be inside the message box or not, let me know :thinking:

PR #160, opened 2019-11-07
Feature/file watcher

jonaslagoni:


jonaslagoni:
If references in the async api document contains references to other files, these will atm not be watched for changes. A few solutions:
1. Find all reference files through the parser, if this is possible to do?
2. Add a `--watch-dir` parameter to the cli where the user can specify custom dirs and files to watch.
Not sure which one is more suiting?

fmvilas:
Don't worry about references because some of them will not be available to "watch". Think that references may be in a remote server. Getting a local references graph would be cool but it's not possible yet and that would complicate this feature a lot. Maybe for a future improvement :)

fmvilas:
I had to revert this PR because it was causing the generator to always fail with the following error:

```
Something went wrong:
Template render error: ([object Object])
  TypeError: Converting circular structure to JSON
    at Object._prettifyError (/Users/fmvilas/www/asyncapi-generator/node_modules/nunjucks/src/lib.js:36:11)
    at Template.render (/Users/fmvilas/www/asyncapi-generator/node_modules/nunjucks/src/environment.js:542:21)
    at Environment.renderString (/Users/fmvilas/www/asyncapi-generator/node_modules/nunjucks/src/environment.js:380:17)
    at Promise (/Users/fmvilas/www/asyncapi-generator/lib/generator.js:481:21)
    at new Promise (<anonymous>)
    at Generator.renderString (/Users/fmvilas/www/asyncapi-generator/lib/generator.js:480:12)
    at Generator.renderFile (/Users/fmvilas/www/asyncapi-generator/lib/generator.js:505:19)
    at <anonymous>
```

It seems that `asyncapiDocument` variable at `lib/generator.js@481:21` contains circular references and can't be converted to JSON by Nunjucks. I've been looking at the code and can't figure out what can be causing this. Any ideas?

fmvilas:
Here's the commit reverting the changes: https://github.com/asyncapi/generator/commit/6058de71ac6648bc820f673d796c687bf1b923fa.

jonaslagoni:
Yea, somehow the fix for issue #146 are somehow reverted, this even happened on some of my other branches.

fmvilas:
I think you're not always creating the branches from the latest master and that's why you're seeing this weird behavior. Also, creating feature branches from other feature branches causes a lot of pain like the one you're describing. May be related to this :)

PR #165, opened 2019-11-08
Added fix to bug where examples in messages where not used correctly.

### LINKED ISSUE #163: When Operations have the same ref for message you cant choose examples in html template
Using the streetlight example provided in the generator and generating using the html template, it is not possible to select "headers" example in the `smartylighting/streetlights/1/0/action/{streetlightId}/turn/off` channel because the id is generated based on `{{channelName}}-{{msg.uid()}}` where `channelName` are not defined.
https://github.com/asyncapi/generator/blob/cd27d10704929eefeeb789be8b951de8ea331535/templates/html/.partials/example.html#L7

### LINKED ISSUE #164: Not possible to define examples in Message object
The examples are not possible to define in the html template because of these lines:
https://github.com/asyncapi/generator/blob/cd27d10704929eefeeb789be8b951de8ea331535/templates/html/.filters/all.js#L120
and
https://github.com/asyncapi/generator/blob/cd27d10704929eefeeb789be8b951de8ea331535/templates/html/.filters/all.js#L130

which gives the error:
```
 node ./cli.js -o ./out ./test/docs/streetlights.yml html
Something went wrong:
Template render error: (d:\rust plugins\Async-api generator\templates\html\index.html)
  Template render error: (d:\rust plugins\Async-api generator\templates\html\.partials\content.html)
  Template render error: (d:\rust plugins\Async-api generator\templates\html\.partials\operations.html) [Line 2, Column 17]
  TypeError: msg.find is not a function
    at Object._prettifyError (d:\rust plugins\Async-api generator\node_modules\nunjucks\src\lib.js:36:11)
    at d:\rust plugins\Async-api generator\node_modules\nunjucks\src\environment.js:567:19
    at eval (eval at _compile (d:\rust plugins\Async-api generator\node_modules\nunjucks\src\environment.js:637:18), <anonymous>:44:11)
    at d:\rust plugins\Async-api generator\node_modules\nunjucks\src\environment.js:575:11
    at eval (eval at _compile (d:\rust plugins\Async-api generator\node_modules\nunjucks\src\environment.js:637:18), <anonymous>:66:12)
    at d:\rust plugins\Async-api generator\node_modules\nunjucks\src\environment.js:575:11
    at eval (eval at _compile (d:\rust plugins\Async-api generator\node_modules\nunjucks\src\environment.js:637:18), <anonymous>:12:11)
    at d:\rust plugins\Async-api generator\node_modules\nunjucks\src\environment.js:617:9
    at eval (eval at _compile (d:\rust plugins\Async-api generator\node_modules\nunjucks\src\environment.js:637:18), <anonymous>:56:12)
    at d:\rust plugins\Async-api generator\node_modules\nunjucks\src\environment.js:617:9
```

jonaslagoni:
fixes #164 
fixes #163

### DIFF
--- a/templates/html/.filters/all.js
+++ b/templates/html/.filters/all.js
@@ -118,17 +118,8 @@ module.exports = ({ Nunjucks, Markdown, OpenAPISampler }) => {
 
   Nunjucks.addFilter('getPayloadExamples', (msg) => {
     if (Array.isArray(msg.examples()) && msg.examples().find(e => e.payload)) {
-      const ex = msg.examples().filter(e => e.payload);
-      if (ex) {
-        //Go through each found payload from filtering the examples array.
-        let returnArray = [];
-        for (var i in ex) {
-          returnArray = returnArray.concat(ex[i].payload);
-        }
-        return returnArray;
-      } else {
-        return undefined;
-      }
+      // Instead of flat or flatmap use this.
+      return msg.examples().map(e => e.payload).filter(Boolean).reduce((acc, val) => acc.concat(val), []);
     }
 
     if (msg.payload() && msg.payload().examples()) {
@@ -138,17 +129,8 @@ module.exports = ({ Nunjucks, Markdown, OpenAPISampler }) => {
 
   Nunjucks.addFilter('getHeadersExamples', (msg) => {
     if (Array.isArray(msg.examples()) && msg.examples().find(e => e.headers)) {
-      const ex = msg.examples().filter(e => e.headers);
-      if (ex) {
-        //Go through each found header from filtering the examples array.
-        let returnArray = [];
-        for (var i in ex) {
-          returnArray = returnArray.concat(ex[i].headers);
-        }
-        return returnArray;
-      } else {
-        return undefined;
-      }
+      // Instead of flat or flatmap use this.
+      return msg.examples().map(e => e.headers).filter(Boolean).reduce((acc, val) => acc.concat(val), []);
     }
 
     if (msg.headers() && msg.headers().examples()) {
--- a/templates/html/.filters/all.js
+++ b/templates/html/.filters/all.js
@@ -117,11 +117,9 @@ module.exports = ({ Nunjucks, Markdown, OpenAPISampler }) => {
   });
 
   Nunjucks.addFilter('getPayloadExamples', (msg) => {
-    if (Array.isArray(msg.examples())) {
-      let payloadElement = msg.examples().find(e => e.payload);
-      if (payloadElement) {
-        return payloadElement.payload;
-      }
+    if (Array.isArray(msg.examples()) && msg.examples().find(e => e.payload)) {
+      const ex = msg.examples().filter(e => e.payload);
+      return ex && ex[0].payload !== undefined ? ex[0].payload : undefined;
     }
 
     if (msg.payload() && msg.payload().examples()) {
@@ -130,11 +128,9 @@ module.exports = ({ Nunjucks, Markdown, OpenAPISampler }) => {
   });
 
   Nunjucks.addFilter('getHeadersExamples', (msg) => {
-    if (Array.isArray(msg.examples())) {
-      let headerElement = msg.examples().find(e => e.headers);
-      if (headerElement) {
-        return headerElement.headers;
-      }
+    if (Array.isArray(msg.examples()) && msg.examples().find(e => e.headers)) {
+      const ex = msg.examples().filter(e => e.headers);
+      return ex && ex[0].headers !== undefined ? ex[0].headers : undefined;
     }
 
     if (msg.headers() && msg.headers().examples()) {
--- a/templates/html/.filters/all.js
+++ b/templates/html/.filters/all.js
@@ -117,8 +117,11 @@ module.exports = ({ Nunjucks, Markdown, OpenAPISampler }) => {
   });
 
   Nunjucks.addFilter('getPayloadExamples', (msg) => {
-    if (Array.isArray(msg.examples()) && msg.find(e => e.payload)) {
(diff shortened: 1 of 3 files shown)


jonaslagoni:
For #164 the examples need to be defined as follows:
```yml
      examples:
        - "payload" :
          - "key1": "value1"
            "key2": "value2"
          - "key1": "value3"
            "key2": "value4"
        - "headers" :
          - "header1"
          - "header2"
```

jonaslagoni:
When I tried
```
   if (Array.isArray(msg.examples()) && msg.examples().find(e => e.headers)) {
      return msg.examples().filter(e => e.headers);
    }
```
And 
```
   if (Array.isArray(msg.examples()) && msg.examples().find(e => e.payload)) {
      return msg.examples().filter(e => e.payload);
    }
```
It showed 
image

fmvilas:
Yeah, it should be something like:

```js
if (Array.isArray(msg.examples()) && msg.examples().find(e => e.payload)) {
  return msg.examples().filter(e => e.payload).payload;
}
```

However, since `filter` can return `undefined`, you should check for existence first:

```js
if (Array.isArray(msg.examples()) && msg.examples().find(e => e.payload)) {
  const ex = msg.examples().filter(e => e.payload);
  return ex && ex.payload !== undefined ? ex.payload : undefined;
}
```

jonaslagoni:
Since filter returns an array I decided to loop over it and concat all arrays it found as examples i.e.
```yml
      examples:
        - "payload" :
          - "key1": "value1"
            "key2": "value2"
          - "key1": "value3"
            "key2": "value4"
        - "headers" :
          - "header32"
          - "header32"
        - "payload" :
          - "keysd": "valuesd"
            "keysad": "asd"
        - "headers" :
          - "header1"
          - "header2"
```
Since this is possible based on the specification.

templates/html/.filters/all.js
```diff
@@ -117,8 +117,18 @@ module.exports = ({ Nunjucks, Markdown, OpenAPISampler }) => {
...
+      if (ex) {
+        //Go through each found payload from filtering the examples array.
+        let returnArray = [];
+        for (var i in ex) {
+          returnArray = returnArray.concat(ex[i].payload);
+        }
+        return returnArray;
+      } else {
+        return undefined;
+      }
```
fmvilas: I think you can probably replace all this code with this:

```js
return msg.examples().map(e => e.payload).filter(Boolean);
```

The `.filter(Boolean)` trick will remove all the `undefined` items in the array.

templates/html/.filters/all.js
```diff
@@ -127,8 +137,18 @@ module.exports = ({ Nunjucks, Markdown, OpenAPISampler }) => {
...
+      if (ex) {
+        //Go through each found header from filtering the examples array.
+        let returnArray = [];
+        for (var i in ex) {
+          returnArray = returnArray.concat(ex[i].headers);
+        }
+        return returnArray;
+      } else {
+        return undefined;
+      }
```
fmvilas: Same as above

templates/html/.filters/all.js
jonaslagoni: Not a bad little 1 liner, problem still remains that we would need to use something like [flatMap ](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/flat) which are only supported in node >= 11 to flatten the array. Instead I used their alternative example provided and ended up with:
`return msg.examples().map(e => e.payload).filter(Boolean).reduce((acc, val) => acc.concat(val), []);`

The checks are still needed in order to ensure payload examples are used (if any) when message examples are not defined.

templates/html/.filters/all.js
jonaslagoni: Same as above :)

templates/html/.filters/all.js
```diff
@@ -117,8 +117,9 @@ module.exports = ({ Nunjucks, Markdown, OpenAPISampler }) => {
   });
 
   Nunjucks.addFilter('getPayloadExamples', (msg) => {
-    if (Array.isArray(msg.examples()) && msg.find(e => e.payload)) {
-      return msg.filter(e => e.payload);
+    if (Array.isArray(msg.examples()) && msg.examples().find(e => e.payload)) {
+      // Instead of flat or flatmap use this.
+      return msg.examples().map(e => e.payload).filter(Boolean).reduce((acc, val) => acc.concat(val), []);
```
fmvilas: That's fine to me but you have lodash in the context, so you could use `_.flatMap`: https://lodash.com/docs/4.17.15#flatMap.

Just add `_` here: https://github.com/asyncapi/generator/pull/165/files#diff-01dfe915256af0d2b6dc5d1ea3de4dedL1, like this:

```js
module.exports = ({ _, Nunjucks, Markdown, OpenAPISampler }) => {
```

templates/html/.filters/all.js
```diff
@@ -127,8 +128,9 @@ module.exports = ({ Nunjucks, Markdown, OpenAPISampler }) => {
   });
 
   Nunjucks.addFilter('getHeadersExamples', (msg) => {
-    if (Array.isArray(msg.examples()) && msg.find(e => e.headers)) {
-      return msg.filter(e => e.headers);
+    if (Array.isArray(msg.examples()) && msg.examples().find(e => e.headers)) {
+      // Instead of flat or flatmap use this.
+      return msg.examples().map(e => e.headers).filter(Boolean).reduce((acc, val) => acc.concat(val), []);
```
fmvilas: Same as above.

templates/html/.filters/all.js
jonaslagoni: Even better :+1:

PR #175, opened 2019-11-19
Fixing watcher on linux

### LINKED ISSUE #169: Watcher not working on linux
Given this information, https://nodejs.org/api/fs.html#fs_caveats the `recursive` option are only working on Mac and Windows.. So to support Linux, we have to use a module or hack the recursive function together our self (i.e. manually watching all sub directories).

jonaslagoni:
Fixes #169

### DIFF
--- a/lib/watcher.js
+++ b/lib/watcher.js
@@ -1,6 +1,4 @@
-
 const fs = require('fs');
-const Path = require('path');
 const chokidar = require('chokidar');
 
 /**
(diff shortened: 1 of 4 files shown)


lib/watcher.js
```diff
@@ -24,75 +28,65 @@ class Watcher {
...
+      if (fs.existsSync(listenerPath)) {
+        const newEventType = this.convertEventType(eventType);
+        this.filesChanged[changedPath] = { eventType: newEventType, path: changedPath};
+        // Since multiple changes can occur at the same time, lets wait a bit before processing.
+        if (this.fsWait) return;
+        this.fsWait = setTimeout(async () => {
+          await changeCallback(this.filesChanged);
+          this.filesChanged = {};
+          this.fsWait = false;
+        }, 500);
```
fmvilas: Should we make this time configurable?

jonaslagoni: Not sure what people could use it for :smile: This timer is to ensure that all files has been processed by the watcher before actually calling the callback. Else it would re-generate for each file that was changed.. Might have to refactor it a bit so it keeps waiting as long as there are changes and then refresh the timeout instead of waiting right after the first change was caught. Not sure

fmvilas: Ohh, I see what you mean. Totally misunderstood the purpose. Nevermind 👍

PR #176, opened 2019-11-21
Remove kebabCase on generateSeparateFile

fmvilas:
It removes `_.kebabCase` from the name of the file on the `generateSeparateFile` function. It was conflicting with file names like `MyClass.java`.


fmvilas:
You're right. We should probably convert names into filename-safe strings.

package.json
```diff
@@ -31,13 +31,12 @@
...
   "dependencies": {
     "ajv": "^6.10.2",
-    "asyncapi": "^2.2.0",
     "asyncapi-parser": "^0.12.0",
     "commander": "^2.12.2",
+    "filenamify": "^4.1.0",
     "fs.extra": "^1.3.2",
     "jmespath": "^0.15.0",
     "js-yaml": "^3.13.1",
-    "json-schema-ref-parser": "^5.1.0",
```
fmvilas: Took the opportunity to remove unused dependencies.

fmvilas:
Not really, only when invalid chars are found. For instance, if your schema name is `MySchema` it will remain as `MySchema`. In the case of a channel like `my/channel` then yes, it would become `my-channel`.

jonaslagoni:
> Not really, only when invalid chars are found. For instance, if your schema name is `MySchema` it will remain as `MySchema`. In the case of a channel like `my/channel` then yes, it would become `my-channel`.

But since that problem is that `my-channel` would not work in java, shouldn't this be converted to `MyChannel` :smile: ?

fmvilas:
The problem is on schemas and other objects like messages. I don't think channels will need to have their own class file. If they do, we can add an option later.

PR #177, opened 2019-11-21
Do not assume components exist

### LINKED ISSUE #172: Generator assumes there's a components section
Generator assumes there's a `components` section and fails when it doesn't exist. The `components` section is optional so we should always be checking for existence first.

fmvilas:
`components` are not mandatory, so we shouldn't assume they exist.

Fixes #172


WaleedAshraf:
components are not, but channels are mandatory. No?
On playground, if I remove channels, it says it is required.

jonaslagoni:
> components are not, but channels are mandatory. No?
> On playground, if I remove channels, it says it is required.

Might be required by the playground, but I cant find any information about its required in the AsyncAPI file. And one might want to only use AsyncAPI to generate a structure :thinking: I guess :smile:

fmvilas:
You're right, @WaleedAshraf. I don't even know the spec I created anymore 😅 . Thanks for pointing out.

PR #178, opened 2019-11-21
Fix shouldOverwriteFile call on separate files

fmvilas:


lib/generator.js
```diff
@@ -406,7 +406,7 @@ class Generator {
       const targetFile = path.resolve(this.targetDir, relativeBaseDir, newFileName);
       const relativeTargetFile = path.relative(this.targetDir, targetFile);
 
-      const shouldOverwriteFile = this.shouldOverwriteFile(relativeTargetFile);
+      const shouldOverwriteFile = await this.shouldOverwriteFile(relativeTargetFile);
```
jonaslagoni: To simplify we could do:
```js
if (!this.shouldOverwriteFile(relativeTargetFile)) return;
```
since the variable are never used.

WaleedAshraf: you also need to await. But, await with and NOT in a if check doesn't look nice.

WaleedAshraf: @jonaslagoni suggestion is also nice. But I'm fine with this fix also.

Note: should fix the other call first.

jonaslagoni: Yea you are right @WaleedAshraf since the function is `async` it should use `await` even though it only returns a boolean otherwise it will skip the step completely.

fmvilas:
I'm confused about this review. @jonaslagoni are you asking me to change something? If we do it in a single line it will look like this:

```js
if (!(await this.shouldOverwriteFile(relativeTargetFile))) {
}
```

Am I missing something?

PS: Check out point 6 [here](https://help.github.com/en/github/collaborating-with-issues-and-pull-requests/commenting-on-a-pull-request) to learn how to make suggestions on Github :)

WaleedAshraf:
@fmvilas it can be this?
```
if (!await this.shouldOverwriteFile(relativeTargetFile)) return;
```

jonaslagoni:
My tired brain switched around what was changed and what was removed :D all good!

fmvilas:
@WaleedAshraf not sure it can be this but it makes me think I'm negating await 😄

fmvilas:
@jonaslagoni have some rest mate!

PR #179, opened 2019-11-22
Add topicPrefix to Nodejs config file

fmvilas:
This option was already there but was not in the config file. It's useful when you're using free instances of cloud brokers, where the vendor usually prepend your username to every topic. I'm leaving it empty so, by default, it's not taken into account but it's possible to override the value from a `.env` file and access it like `config.broker.kafka.topicPrefix`.

### Example of free cloud Kafka instance at [cloudkarafka.com]()

`qw7yecbj` is my username and it prepends `qw7yecbj-` to every topic.

Screen Shot 2019-11-22 at 12 57 07

### DIFF
--- a/templates/nodejs/config/common.yml
+++ b/templates/nodejs/config/common.yml
@@ -40,7 +40,7 @@ default:
       brokers:
         - {{ asyncapi.server(params.server).url() | host }}
       consumerOptions:
-        groupId: 'replace-me'
+        groupId: {{ asyncapi.info().title() | camelCase }}
       topics:
       {%- for topic in asyncapi | channelNamesWithPublish %}
         - {{ topic | toKafkaTopic }}
--- a/templates/nodejs/config/common.yml
+++ b/templates/nodejs/config/common.yml
@@ -46,6 +46,7 @@ default:
         - {{ topic | toKafkaTopic }}
       {%- endfor %}
       topicSeparator: '__'
+      topicPrefix:
 {%- endif %}
 
 development:


templates/nodejs/config/common.yml
```diff
@@ -40,12 +40,13 @@ default:
       brokers:
         - {{ asyncapi.server(params.server).url() | host }}
       consumerOptions:
-        groupId: 'replace-me'
+        groupId: {{ asyncapi.info().title() | camelCase }}
```
fmvilas: Sorry for including an unrelated change in this PR. I'm just trying to reduce overhead.

WaleedAshraf:
Whats this Nodejs templating is for?

fmvilas:
What? @WaleedAshraf sorry I don't get the question.

WaleedAshraf:
We only had html/markdown/ generation. Good to see that now we support Node.js code also. But I'm not sure how useful it would be? As compared to the effort we put in. I tried to run it with streetlights example but didn't work.

fmvilas:
Ohh! I thought you knew about it. It's been there for months now 😅 

Can you clarify what's not working for you with the streetlights example?

WaleedAshraf:
Maybe I'm doing it wrong. We need to update readme also.
image

fmvilas:
🤔 Interesting. I don't know how you got there. The only wrong thing I can see is the `servers` param. It should be `server` instead. Also `dev` server doesn't exist, just `production`.

That's what I see when I do it like you (the error message makes sense):
Screen Shot 2019-11-22 at 19 58 02

fmvilas:
The error @WaleedAshraf was seeing has been resolved offline. Merging.

Issue #134, opened 2019-09-25
Java generator


AlexTwX:
I'm trying to use the java generator.

I converted my yaml to 2.0, it works with html & markedown but when i try to generate with java-spring, the files generated are juste an empty template.

Something is wrong with the generator ?

fmvilas:
Hey @AlexTwX. Yes, apparently there's a problem. @rmelian is going to work on fixing it as I'm not very proficient in Java. Also, any guidance or clues you may find are very welcome. Thanks for reporting it!

rmelian:
hi @AlexTwX can you please share your specification or some part of it so I can test. I'm happy to help you here

AlexTwX:
You can try with the example in the java-spring readme.
https://github.com/asyncapi/generator/tree/master/templates/java-spring
Before and after converted it with https://github.com/asyncapi/converter

wolfey:
any news here? this is pretty much unusable at the moment..

fmvilas:
Not yet. We'll soon work on this but can't tell you what soon exactly means yet. In any case, any contribution would be greatly appreciated :)

nickspacek:
Running into the same issue, going to have a peek at the code and see if anything obvious is happening to muck with the templates.

nickspacek:
Ah so I think the root of it is that the Java generator was not updated for the 2.0.0 schema. I'll share my work over here as I go: nickspacek/generator@02c2d1a

nickspacek:
So I'm mostly finished what I wanted to accomplish; the templates that exist are turning into Java code now. Would like to do a round of cleanup with lessons learned as I went through.

What is supposed to be happening with server definitions and server variables when processed by the templates? For example, a URL that has a port, and the port exists as a server variable. Is that supposed to remain a placeholder in the template, or should the generator itself or the templating code be filling in the port with the defaultValue()?

fmvilas:
Hey @nickspacek! This is great stuff. Currently, we're not processing server variables yet. It's been reported [here](https://github.com/asyncapi/generator/issues/182). A good first step would be to just connect to the raw `url` string, without performing any replacement. One thing we're doing on the Node.js template is to [make the `server` param mandatory](https://github.com/asyncapi/generator/blob/master/templates/nodejs/.tp-config.json#L3). This way, you only have to deal with a single server in the generated code.

As a later improvement, we can implement support for multiple servers and protocols and server variables. What do you think?

nickspacek:
[...]

PR #188, opened 2019-12-02
Java Spring generator 2.0.0

nickspacek:
I've updated the Java Spring template to work with AsyncApi 2.0.0. I tried to follow the existing formatting but let me know if I wandered or if you notice any inconsistencies.

This should address #134

### DIFF
--- a/templates/java-spring/src/main/resources/application.yml
+++ b/templates/java-spring/src/main/resources/application.yml
@@ -1,6 +1,7 @@
 {% set server = asyncapi.server(params.server) -%}
   {% if server.protocol() === 'amqp' -%}
 amqp:
+  # {{server.description()}}
   broker:
     host: localhost
     port: 5672
@@ -17,8 +18,10 @@ amqp:
   {%- endif -%}
   {% if server.protocol() === 'mqtt' -%}
 mqtt:
+  # {{server.description()}}
   broker:
-    host: tcp://{{server.url()}}
+    scheme: tcp
+    host: {{server.url()}}
     port: {{server.variable('port').defaultValue()}}
     username:
     password:
--- a/templates/java-spring/src/main/resources/application.yml
+++ b/templates/java-spring/src/main/resources/application.yml
@@ -1,5 +1,5 @@
-{% for server in asyncapi.servers %}
-  {% if server.scheme === 'amqp' %}
+{% set server = asyncapi.server(params.server) -%}
+  {% if server.protocol() === 'amqp' -%}
 amqp:
   broker:
     host: localhost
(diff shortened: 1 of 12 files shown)


templates/java-spring/src/main/resources/application.yml
```diff
@@ -1,34 +1,32 @@
...
+  {%- endif -%}
+  {% if server.protocol() === 'mqtt' -%}
 mqtt:
+  # {{server.description()}}
   broker:
-    host: tcp://{{server.url}}
-    port: {{server.variables.port.default}}
+    scheme: tcp
+    host: {{server.url()}}
+    port: {{server.variable('port').defaultValue()}}
```
fmvilas: This is failing for me. It's assuming there will be a `port` variable and it might not exist.

nickspacek: Thanks for the feedback, I'll have a go over and make sure that the pub/sub is logical! I thought I had followed what was there but perhaps the semantics changed some with channels vs. topics; I should have paid more attention to what the templates were actually outputting! I should be able to get those changes in shortly.

fmvilas: No worries, it happens to me all the time when working on the code generators.

nickspacek: Hmm, after reviewing things I'm less comfortable making changes 2 and 3. It seems that the intent of the original author was to have the java-spring generator create client code, so it's written from the perspective of "I downloaded someone else's AsyncApi document and I want to communicate with their API". So, the document indicating that there is a channel that can be published to results in generating code that can indeed publish to the channel, instead of creating code that can receive messages.

I think it's a great idea to have both, similar to OpenAPI: https://openapi-generator.tech/docs/generators.html. In my case, I actually am trying to generate the server side, so I will probably go ahead and create a new template that has the reverse logic as you had suggested. That will be a template that takes the specification and serves the specification, as well as the endpoints it is describing for consumers of the API.

Now, another issue I'm seeing is mostly to do with item 4, but also 1. The AsyncApi specification seems to be unclear in regards to the server.url. The examples given on XXX do not illustrate proper URLs; in fact, I don't see how you could tell whether some of them are supposed to be "relative URLs" or host names. I'm in favor of supporting proper URLs as you have illustrated over schema-less ones; it is unfortunate that the official examples do not follow this (I found asyncapi/asyncapi#274 that raises this concern).

I think what I am going to is treat the example schema I've been using as an example of a relative URL. In my opinion, the generator framework itself should be taking care of passing a proper JS URL entity to the templates. In the case of relative URLs it would require the user to specify a base URL as a command line option. I would be wiling to explore this concept in a branch.

jonaslagoni:
Hey @nickspacek, can you fix the following spelling mistakes suggested by @RobertDiebels in
https://github.com/asyncapi/generator/pull/191#discussion_r356219213?

nickspacek:
I can, although I was hoping to have some feedback from @fmvilas on my wall of text above.

I also neglected to re-incorporate the x-service-name as a naming option and opted to just rely on the operation ID, so I'll have to fix that to allow x-service-name as the primary source of the method name when it is present.

templates/java-spring/src/main/resources/application.yml
fmvilas: Sorry for the delay. I agree with you. My recommendation is that you forget about the intent of the original author because this template was always a proof of concept and was never completed. I'd go directly for the "server" implementation first as that's what the other templates are doing. We'll work on the perspective of "this is someone else's AsyncAPI document" later. In regard to the URLs, feel free to treat them as absolute URLs for now meanwhile we fix the other issue, which will probably require a new version of the spec.

fmvilas:
> I also neglected to re-incorporate the x-service-name as a naming option and opted to just rely on the operation ID, so I'll have to fix that to allow x-service-name as the primary source of the method name when it is present.

In Java, don't you think it makes more sense if you use `x-service-name` as a class name and `operationId` as the method name? IIRC that's how Swagger Codegen was using it.

fmvilas:
I'm closing since it's no longer needed. The current Java Spring generator already supports AsyncAPI 2.0.0 and a bunch of more features (kudos to @Tenischev). Thanks anyway, @nickspacek. Feel free to reopen if you think it makes sense.

PR #190, opened 2019-12-06
Fixing not all messages, schema, parameters and security schemes are separated

### LINKED ISSUE #170: Not all messages, schema, parameters and security schemes are separated
When messages are not specified in the AsyncAPI document under components they are unable to be separated into different files. Gonna use messages as example in the following, however this is also a problem for shema, parameters and security schemes.

Given that most use-cases does imply specifying messages as the streetlights example
https://github.com/asyncapi/generator/blob/82f2c809f18d15a32fc888bc2b994c5a18555901/test/docs/streetlights.yml#L87-L112

but into separate files and then referenced, I would expect that creating a template file with the name `$$message$$` would result in all messages regardless of where it is located in the AsyncAPI document, to be created into separate files.

However one could say that the `$$message$$` name should only be used for the messages specified in the components section, so I might need some verification whether this is a bug or :smile:

jonaslagoni:
fixes #170

### DIFF
--- a/lib/generator.js
+++ b/lib/generator.js
@@ -297,7 +297,7 @@ class Generator {
   generateDirectoryStructure(asyncapiDocument) {
     const fileNamesForSeparation = {
       channel: asyncapiDocument.channels(),
-      message: asyncapiDocument.components() ? asyncapiDocument.components().messages() : {},
+      message: asyncapiDocument.allMessages(),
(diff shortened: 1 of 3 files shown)


jonaslagoni:
Waiting for a new version of the parser to continue.

fmvilas:
New parser version has just been released :)

PR #191, opened 2019-12-07
Introducing testing

### LINKED ISSUE #158: Introducing testing for the generator and cli
While changing and fixing bugs for the generator there should be some indications that the generator and the cli function correctly within a PR and during development.

I am unsure if the templates themself should somehow be tested as well, not quite sure how that should be done 🤔

jonaslagoni:
fixes #158


jonaslagoni:
@fmvilas I need a quick review of the implementation if I am on the right track or if you have any suggestions :) `html` and `java-spring` with no parameters have been created and generated.

test/templates/java-spring/no-params/README.md
```diff
@@ -0,0 +1,104 @@
...
+  <em>Use your AsyncAPI definition to generate java code to subscribe and publish messages</em>
+</p>
+<br><br>
+
+## Usage
+
+### AsyncAPI extensions
+In order for the generator to know what names to use for some methods it's necessary to make use of [AsyncAPI specification extensions](https://www.asyncapi.com/v1/spec.html#Specification-Extensions). The extension to use is **x-service-name** and must be applied to all topics.
+
+here is an example of how to use it:
```
RobertDiebels: - Capitalization
- Precision
"Here is an example for a single topic:"

test/templates/java-spring/no-params/README.md
```diff
@@ -0,0 +1,104 @@
...
+```yml
+topics:
+  event.lighting.measured:
+    x-service-name: measures
+    publish:
+      $ref: '#/components/messages/lightMeasured'
+    subscribe:
+      $ref: '#/components/messages/lightMeasured'
+```
+here is a complete example
```
RobertDiebels: - Capitalization
- Precision

"Here is an example of a complete specification:"

test/templates/java-spring/no-params/README.md
```diff
@@ -0,0 +1,104 @@
+<h1 align="center">Java Spring generator</h1>
+<p align="center">
+  <em>Use your AsyncAPI definition to generate java code to subscribe and publish messages</em>
+</p>
+<br><br>
+
+## Usage
+
+### AsyncAPI extensions
+In order for the generator to know what names to use for some methods it's necessary to make use of [AsyncAPI specification extensions](https://www.asyncapi.com/v1/spec.html#Specification-Extensions). The extension to use is **x-service-name** and must be applied to all topics.
```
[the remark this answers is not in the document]
jonaslagoni: Damn thorough read! The suggestions have been brought on to the java spring 2.0 template :)

test/templates/java-spring/no-params/README.md
jonaslagoni: The suggestions have been brought on to the java spring 2.0

test/templates/java-spring/no-params/README.md
jonaslagoni: The suggestions have been brought on to the java spring 2.0

test/integration/templates/generation scripts/generate.js
```diff
@@ -0,0 +1,14 @@
...
+  try {
+    await generateHtml(pathToExpectedTemplates);
+    await generateJavaSpring(pathToExpectedTemplates);
+    // ... more templates to generate
+    console.log('Done generating html templates');
+  } catch (e) {
+    console.log(e);
+  }
+}
+generate();
```
RobertDiebels: This can throw a `UnhandledPromiseRejectionWarning`.
If you wrap it like this:
```
(async () => {
try{
    await generate();
}
catch(error){
  // Some elegant handling
}
})();
```
You'll avoid Node's warning en prevent failures when Node no longer allows uncaught promise rejections.

jonaslagoni: Good catch! Gonna wrap them as you as you suggested with a `handlePromiseGracefully` function :+1:

jonaslagoni:
So I came across this problem where the `html` template generates dynamic data, specifically a datetime, which results in the integration test will fail most times. How can that be handled, or should we just disallow dynamic data in that form in the templates :thinking: ? Thought about checking all the files for a regex of datetime formats but we will never find all the values and this does not provide any resolution for when other kind of dynamic data are used.

fmvilas:
In regards to the dynamic data, you might want to add a list of "accepted errors" for each template, so when the code sees these kinds of errors it ignores them. Not a perfect solution but it will better than what we have anyway.

jonaslagoni:
> I think this is going really well. Maybe would be nice to have a way to semi-automatically generate all the templates without having to touch the JS code. For instance, having the JS code read from a file with an array like this:
> 
> ```js
> module.exports = [
>   {
>     template: 'nodejs',
>     params: {
>       server: 'production'
>     }
>   },
>   {
>     template: 'html',
>     params: {
>       sidebarOrganization: 'byTags'
>     }
>   },
> ];
> ```
> 
> And potentially including more needed information to call the CLI. What do you think?

I like that idea yea, I mean for me I would rather have it directly included in the JS tests. However since people might not know anything about JS but most likely know JSON, this would be great. Gonna do it 👍 

> In regards to the dynamic data, you might want to add a list of "accepted errors" for each template, so when the code sees these kinds of errors it ignores them. Not a perfect solution but it will better than what we have anyway.

Hmm, maybe we can do that 🤔 Im gonna try and see what is possible, not quite sure it is that easy 😄 Depends on what kind of error [dir-compare](https://www.npmjs.com/package/dir-compare) returns and how we can determine the differences.

derberg:
@fmvilas @jonaslagoni 
guys what is the status of this PR, is there still something we could extract and could be useful or because of recent unit tests intro this one is obsolete?

jonaslagoni:
@derberg all I can see which could be useful is the [watcher test](https://github.com/asyncapi/generator/blob/cb14bc17efa0c52a9677f40b364cda0f00b83167/test/unit/lib/watcher/watcher.spec.js).

derberg:
@jonaslagoni awesome, I will create a follow up issue for watcher test and take the test into description and close this one, unless you want to create a separate PR for the test?

jonaslagoni:
@derberg do that 😄

derberg:
closing, watcher as followup here https://github.com/asyncapi/generator/issues/347

PR #197, opened 2020-01-04
Fix content overlapping examples

simonalbrecht:
This fixes an issue where the centered content might overlap the examples on the right on medium-sized screens (15").

It also removes the `Example` section within the schema/payload block, as this information is duplicated anyway by the examples on the right.

**Before**
Before

**After**
After

### DIFF
--- a/templates/html/.partials/schema-prop.html
+++ b/templates/html/.partials/schema-prop.html
@@ -89,6 +89,14 @@
             {% endfor %}
           </div>
         {% endif %}
+        {% if prop.examples() %}
+          <div class="text-xs">
+            Examples:
+            {% for value in prop.examples() %}
+              <span class="border text-orange rounded ml-1 py-0 px-2">{{value | dump(2) | safe }}</span>
+            {% endfor %}
+          </div>
+        {% endif %}
       </div>
     </div>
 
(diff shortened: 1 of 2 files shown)


templates/html/.partials/schema-prop.html
```diff
@@ -89,14 +89,6 @@
...
           </div>
         {% endif %}
-        {% if prop.examples() %}
-          <div class="text-xs">
-            Examples:
-            {% for value in prop.examples() %}
-              <span class="border text-orange rounded ml-1 py-0 px-2">{{value | dump(2) | safe }}</span>
-            {% endfor %}
-          </div>
-        {% endif %}
```
fmvilas: This doesn't remove the examples **only** for the payload block but for examples of inner properties too.

simonalbrecht: Ah, I must have missed that as I do not have examples for inner properties. I'll revert the commit related to this -- also taking your advice into account.

fmvilas:
Thanks a lot for the fix, @simonalbrecht! I'm merging.

fmvilas:
It's now published on npm. Thanks a lot!

Issue #205, opened 2020-02-04
[HTML template] Support <base href="{{param.href}}"></base> in index.html


stefanseeger:
When the generator is running inside an express app with multiple routings, the links to the css and js files are always linked to the domain root e.g.

* http://root/css/main.css
* http://root/js/main.js

With a dynamic <base href=""> it could be executed in any folder.

fmvilas:
Hi @stefanseeger. Thanks for opening the issue. I updated the title to include "[HTML template]" because I'm assuming this is the template you're referring to, right? If that's the case, we could implement it using template parameters as we did [here](https://github.com/asyncapi/generator/blob/master/templates/html/.partials/sidebar.html#L31).

As always any contribution is welcome. We'd be happy to guide you.

stefanseeger:
Problem solved.
Thank you for your help!

best regards,
Stefan

PR #206, opened 2020-02-04
#205 Adds baseHref param to index.html

stefanseeger:
Hi,
I addressed #205, to be able to support subfolders in domains by setting templateParam

````
{  templateParams: {baseHref: '/async-docs/'} }
````

Best regards,
Stefan

### DIFF
--- a/templates/html/index.html
+++ b/templates/html/index.html
@@ -2,6 +2,9 @@
 <html lang="en">
 <head>
   <meta charset="UTF-8">
+  {% if params.baseHref %}
+   <base href="{{params.baseHref}}"></base>
+  {% endif %}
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: templates/html/index.html [hunk 1]
```diff
@@ -2,6 +2,9 @@
 <html lang="en">
 <head>
   <meta charset="UTF-8">
+  {% if params.baseHref %}
+   <base href="{{params.baseHref}}"></base>
```
fmvilas: As per [HTML documentation](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/base) on the base tag, there shouldn't be a closing tag:

```suggestion
   <base href="{{params.baseHref}}">
```

fmvilas:
It's now published on npm. Thanks a lot!

PR #211, opened 2020-02-06
Fix bug in array items

fmvilas:
This PR fixes a bug that's happening when a Schema object of type `array` contains an `items` property which itself is an array.

An example to reproduce the bug:

```yaml
asyncapi: 2.0.0
info:
  title: Test
  version: 1.0.0

channels:
  test:
    publish:
      message:
        payload:
          type: array
          items:
            - type: object
```

The reason is that `items` can be either an array of objects or an object.

### DIFF
--- a/templates/html/.partials/schema-prop.html
+++ b/templates/html/.partials/schema-prop.html
@@ -141,8 +141,8 @@
           {% if prop.items() %}
             <p class="pl-6 mb-2 text-xs font-bold uppercase text-grey-darker">Items:</p>
             {% if prop.items() | isArray %}
-              {% for itName, it in prop.items() %}
-              {{ schemaProp(it, itName, odd=(not odd), required=(prop.required() | contains(pName))) }}
+              {% for it in prop.items() %}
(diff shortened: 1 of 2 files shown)


templates/html/.partials/schema-prop.html
```diff
@@ -141,8 +141,8 @@
           {% if prop.items() %}
             <p class="pl-6 mb-2 text-xs font-bold uppercase text-grey-darker">Items:</p>
             {% if prop.items() | isArray %}
-              {% for itName, it in prop.items() %}
-              {{ schemaProp(it, itName, odd=(not odd), required=(prop.required() | contains(pName))) }}
+              {% for it in prop.items() %}
+              {{ schemaProp(it, loop.index0, odd=(not odd), required=(prop.required() | contains(pName))) }}
```
fmvilas: Quoting the Nunjucks docs:

> loop.index0: the current iteration of the loop (0 indexed)

fmvilas:
FYI @zdne, I found the bug on the playground :) Will deploy it as soon as possible. Thanks for reporting!

PR #215, opened 2020-02-06
Add $$everySchema$$ to replace $$schema$$

### LINKED ISSUE #213: The schema files separator got broken recently.
If I have a $$schema$$ file, and this in my document:
```
components:
  schemas:
     Account:
        properties:
           name:
              type: string
```
the generator used to create a file called Account. Now it creates one called anonymous-schema-1.
I can force it to do the right thing by adding a $id property, but I don't think we should force users to do that.

fmvilas:
`$$schema$$` is producing a file for each schema, including parameters, headers, etc. That's not the desired behavior therefore, this PR makes `$$schema$$` generate a file for each schema defined under the components section.

To achieve the same behavior as before, you can use `$$everySchema$$`.

Fixes #213


fmvilas:
It's now on npm, Michael. Thanks!

Issue #187, opened 2019-11-29
Automate docker image releases


rv0:
docker pull asyncapi/generator:latest gives you version 0.22.0 now, this version is over 2 months old ( https://hub.docker.com/r/asyncapi/generator/tags ). 
Ideally new release also create new docker images

jonaslagoni:
Yea agree, should be automated 👍

derberg:
after some cleanup on local, I've just pushed to docker latest 0.28.3 version with tag `0.28.3` and `latest`. Now will work on automating it.

derberg:
Ok, my job is done, ready for review
Once it is merged I will set proper secrets for the GitHub Action flow in Settings/Secrets

derberg:
Docker image will now always be published at the same time NPM package, once we tag the repo.

You cannot see any action here https://github.com/asyncapi/generator/actions because it won't show up until the first run of the Action

PR #218, opened 2020-02-07
Cleanup setup of docker and add automated publishing

derberg:
GitHub Action that published docker image will be triggered on a new tag creation.

- changed base image in dockerfile to alpine version which makes image much smaller and is enough, and still official node image. Instead of 937MB it is 145MB now on local
- added license to package.json as npm didn't like it is missing
- added example of running docker on local to README.md
- sorry for badges but I'm just a huge fan....can remove if you do not like them...
- cleaned up makefile and release.sh
- GitHub action is very simple as most of the work was already done in scripts that I reused. I decided to keep flow with scripts instead of specific GitHub Action as I think having scripts in a repo is great if you have to push from local, and best if the same scripts are also used in CI
- echoing the password to stdin in docker login in github action - I took it from here https://github.com/elgohr/Publish-Docker-Github-Action/blob/a6e48211ae26c311681c44c45a0ca5130dd9c7aa/entrypoint.sh#L26 but to be honest I have no idea why it is better than just passing it as a parameter, like username. I assumed it is best practice, especially that it is used in this very popular GitHub Action. 

You can see GitHub action works like a charm on my fork where I configured it for the sake of this PR: https://github.com/derberg/generator/runs/432173556?check_suite_focus=true

Related to #187

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -1,0 +1,15 @@
+name: Publish docker image
+
+on:
+  release:
+    types: [created]
+
+jobs:
+  build:
+
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v2
+    - name: Run a one-line script
+      run: make docker release
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -12,4 +12,6 @@ jobs:
     steps:
     - uses: actions/checkout@v2
     - name: Run a one-line script
-      run: make docker release
+      run: |
+        echo ${{secrets.DOCKER_PASSWORD}} | docker login -u ${{secrets.DOCKER_USERNAME}} --password-stdin
+        make docker release
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -1,8 +1,9 @@
 name: Publish docker image
 
 on:
-  release:
-    types: [created]
+  push:
+    tags:
+    - '*'
 
 jobs:
   build:
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -7,12 +7,10 @@ on:
 
 jobs:
   build:
-
     runs-on: ubuntu-latest
-
     steps:
     - uses: actions/checkout@v2
-    - name: Run a one-line script
+    - name: Start release
       run: |
         echo ${{secrets.DOCKER_PASSWORD}} | docker login -u ${{secrets.DOCKER_USERNAME}} --password-stdin
         make docker release
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,12 +1,13 @@
-FROM node:11
+FROM node:13.8.0-alpine3.11
 
-WORKDIR /asyncapi
+WORKDIR /app
 
-# install dependencies
+# Install dependencies
 COPY package*.json ./
+
 RUN npm install
 
-# copy code
+# Copy sources
 COPY . .
 
-ENTRYPOINT [ "node", "/asyncapi/cli" ]
+ENTRYPOINT [ "node", "/app/cli" ]
--- a/README.md
+++ b/README.md
@@ -2,6 +2,9 @@
 <p align="center">
   <em>Use your AsyncAPI definition to generate literally anything. Markdown documentation, Node.js code, HTML documentation, anything!</em>
 </p>
+
+![npm](https://img.shields.io/npm/v/asyncapi-generator?style=for-the-badge) ![npm](https://img.shields.io/npm/dt/asyncapi-generator?style=for-the-badge)
+
 <br><br>
 
 > :warning: This package doesn't support AsyncAPI 1.x anymore. We recommend to upgrade to the latest AsyncAPI version using the [AsyncAPI converter](https://github.com/asyncapi/converter). If you need to convert documents on the fly, you may use the [Node.js](https://github.com/asyncapi/converter) or [Go](https://github.com/asyncapi/converter-go) converters.
@@ -12,10 +15,12 @@
 npm install -g asyncapi-generator
 ```
 
-Or use all the commands below using Docker by replacing `ag` in all the commands below with:
+Or just use Docker:
 
 ```bash
 docker run --rm -it -v $PWD:/app -w /app asyncapi/generator [COMMAND HERE]
+# Example that you can run inside generator directory after cloning this repository
+docker run --rm -it -v $PWD:/app -w /app asyncapi/generator -o ag -o ./my_output ./test/docs/streetlights.yml markdown
 ```
 
 ## Usage
--- a/README.md
+++ b/README.md
@@ -18,9 +18,15 @@ npm install -g asyncapi-generator
 Or just use Docker:
 
 ```bash
-docker run --rm -it -v [ASYNCAPI FILES LOCATION]:/app/my_files asyncapi/generator [COMMAND HERE]
-# Example that you can run inside generator directory after cloning this repository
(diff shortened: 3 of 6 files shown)


README.md
```diff
@@ -12,10 +15,12 @@
 npm install -g asyncapi-generator
 ```
 
-Or use all the commands below using Docker by replacing `ag` in all the commands below with:
+Or just use Docker:
 
 ```bash
 docker run --rm -it -v $PWD:/app -w /app asyncapi/generator [COMMAND HERE]
+# Example that you can run inside generator directory after cloning this repository
+docker run --rm -it -v $PWD:/app -w /app asyncapi/generator -o ag -o ./my_output ./test/docs/streetlights.yml markdown
```
jonaslagoni: This does not work in windows 10, neither in PowerShell or Git Bash terminal. We could replace `$PWD` with `${PWD}` which works for PowerShell and pretty sure on Linux systems too 😄

derberg: I just pushed `{`

do you have Windows and can check? I think for windows you simply need to replace `${PWD}` with `//C/Users/YourUser/`

Dockerfile
```diff
@@ -1,12 +1,13 @@
-FROM node:11
+FROM node:13.8.0-alpine3.11
```
RobertDiebels: NodeJS version is updated to 13.8.0 but Travis is building with 8.x.x [ [.travis.yml](https://github.com/asyncapi/generator/blob/41efaf8bce4098702995f4b58c55ef442d1ee593/.travis.yml#L3) ]. Would suggest bumping the build-version up to 13.8.0 as well.

.github/workflows/main.yml
```diff
@@ -0,0 +1,16 @@
+name: Publish docker image
```
RobertDiebels: The generator currently has Travis for continuous integration (CI) I'd suggest choosing one CI-system and sticking with that one.

Dockerfile
derberg: Not sure if it is a big problem if the previous version was 11, so still not the same as the one used in Travis. I suggest to leave it as it is and move travis to github actions later.

.github/workflows/main.yml
derberg: yeap, and I suggest github actions, they are very well integrated into Github, then we have one tool for code and its build, and super nicely integrated secrets management. This PR was a test how it works, and I have to say I'm pleased as I was able to integrate it in just few hours without any previous knowledge of the tool

Dockerfile
RobertDiebels: @derberg Alright I can live with that. As long as it's on the radar 👍 .

.github/workflows/main.yml
RobertDiebels: I have zero preference for one CI system over another. As long as they get the job done. However, having two CI-systems in the same code-base does make my spidey senses tingle. If you're ok with having two that's fine with me. Either way, I created a CI/CD issue today. I could add the move to GitHub actions in there to make sure its on the TODO-list? Issue -> [here](https://github.com/asyncapi/generator/issues/220)

derberg: I saw the issue, awesome stuff. I think that having consistent flow and tooling across all repos + dev-friendly setup is something that should enable easier onboarding of new contributors to the project

README.md
jonaslagoni: Ill check when I get home :+1:

derberg:
@jonaslagoni @RobertDiebels can you please have a look again and approve, I think all the comments were addressed or mitigated to different channels to continue discussion.

README.md
jonaslagoni: I am getting some weird errors 😄 

**PowerShell**
If I am using the command ` docker run --rm -it -v ${PWD}:/app -w /app gen -o ag -o /my_output ./test/docs/streetlights.yml markdown`  i get the following stacktrace:
```bash
internal/modules/cjs/loader.js:983
  throw err;
  ^

Error: Cannot find module '/app/cli'
    at Function.Module._resolveFilename (internal/modules/cjs/loader.js:980:15)
    at Function.Module._load (internal/modules/cjs/loader.js:862:27)
    at Function.executeUserEntryPoint [as runMain] (internal/modules/run_main.js:71:12)
    at internal/main/run_main_module.js:17:47 {
  code: 'MODULE_NOT_FOUND',
  requireStack: []
}
```
However when I bind the output path it works fine `docker run --rm -it -v ${PWD}:/app/my_output -w /app gen -o ag -o ./my_output ./test/docs/streetlights.yml markdown`
```
Done! ✨
Check out your shiny new generated files at /app/my_output.
```

**Suggestions**
In the example two output folders are specified, ` -o ag -o ./my_output` for the sake of clarity shouldn't we just include one? 

Second, we bind the app volume meaning the application code will also get included, which I assume we do not want when using docker 🤔 

Third, the workdir argument is not needed since this is already defined in the dockerfile.

So I suggest 
`docker run --rm -it -v ${PWD}:/app/output gen -o /output ./test/docs/streetlights.yml markdown`.
We might change the volume to `-v ${PWD}/output:/app/output` if this is more desired, however in windows the folders have to be manually created.

derberg: - remove `-w` - I totally agree, removed
- remove `-o ag` - 🤦‍♂ that is actually a stupid copy/paste mistake from my side, sorry for the confusion, removed
- mounting the whole app code - even though it is just an example it should follow best practices, good catch 👌 

After some tweaking, I think now command looks pretty ok for all users.
@jonaslagoni do we need to do something more for windows? I have to admit I got a bit lost in your comment

jonaslagoni: Gonna start a new conversation on the new code.

README.md
```diff
@@ -12,10 +15,12 @@
...
 ```
 
-Or use all the commands below using Docker by replacing `ag` in all the commands below with:
+Or just use Docker:
 
 ```bash
-docker run --rm -it -v $PWD:/app -w /app asyncapi/generator [COMMAND HERE]
+docker run --rm -it -v [ASYNCAPI FILES LOCATION]:/app/my_files asyncapi/generator [COMMAND HERE]
+# Example that you can run inside generator directory after cloning this repository
+docker run --rm -it -v ${PWD}/test/docs:/app/my_files asyncapi/generator -o ./my_files my_files/streetlights.yml markdown
```
jonaslagoni: I guess the common use case would be `use this AsyncAPI file and generate the output to some folder` therefore if we want to include an example of how to use the docker image, I think we have to split the volume for both the AsyncAPI file and the output, even for the custom template. 
If we just want to use a custom AsyncAPI file:
```bash
docker run --rm -it \
-v /my/output/path:/app/output \
-v /my/spec/path/to/file.yml:/app/asyncfile.yml \
asyncapi/generator -o ./output app/asyncfile.yml markdown
```
If we want to use it with a custom template called `CustomTemplate`:
```bash
docker run --rm -it \
-v /my/output/path:/app/output \
-v /my/spec/path/to/file.yml:/app/asyncfile.yml \
-v /my/template/path:/app/custom_template \
asyncapi/generator --templates app/templates -o ./output app/asyncfile.yml CustomTemplate
```
However the complexity increases all of the sudden, and we move away from a nice and easy one liner 😄 
What you think @derberg ?

derberg: I like the idea of a split between input and output, it looks clearer now I think - have a look

as it comes to the custom template example, don't you think it might be a bit to much? First of all example will not work by copy/paste, and I would kind of expece that main example already clearly introduces the concept that should be easy to modify, don't you think?

jonaslagoni: Yes, the problem/solution is the same as for the AsyncAPI file, so I agree we can leave the template parameter out.  

Looks fine 👍

PR #222, opened 2020-02-14
Set sane CSS defaults on the HTML template

fmvilas:
When using the HTML template on a dark background browser, the background color is not white but "dark" and the template doesn't look great. This PR provides a sane default for background and links.

Fixes https://github.com/asyncapi/vscode-extension/issues/1

### DIFF
--- a/templates/html/css/main.css
+++ b/templates/html/css/main.css
@@ -1,9 +1,18 @@
 html, body {
   font-weight: lighter;
+  background-color: white;
 }
 
 h1, h2, h3, h4, h5, h6 {
   font-weight: lighter;
+}
+
+a {
+  color: #333;
+}
+
+a:hover {
+  color: black;
 }
 
 .string-chunk {


derberg:
Hey, not sure what is actual problem. When I generate the HTML and open locally in a browser it all looks good and the same, no matter if I have this line or not https://github.com/asyncapi/generator/pull/222/files#diff-d3b5a7837814e025266678e527e4d3fdR3

Thing is different in the preview in VCS, I use HTML preview plugin from Thomas Haakon, not sure though how much it is customized and that maybe I'm getting different results only because of the plugin. But anyway:


This is what I see with `background-color: white`. Font should be darker
Screenshot 2020-02-14 at 13 53 42

This is what I see without `background-color: white`. Black looks black to me.
Screenshot 2020-02-14 at 13 53 59

fmvilas:
The VS Code extension uses the internal WebView (chromium) browser. These changes are the minimum ones required for it to work.

fmvilas:
I'm just moving the [hardcoded CSS here](https://github.com/asyncapi/vscode-extension/blob/master/src/preview-content-provider.ts#L219) to the template CSS.

PR #223, opened 2020-02-18
Add feature to support git repos

### LINKED ISSUE #180: Better integration with Git
## Problem
One of the common problems when generating code is the fear of removing business logic when regenerating code in an existing project. To avoid this, we introduced the `-n, --no-overwrite` option but this doesn't prevent people from forgetting about it or forgetting about a path they don't want to overwrite, and accidentally deleting code they can't recover.

## Solution
Check if the directory is a git repo. If that's the case, the generator should not allow you to regenerate the project if there are unstaged changes in the working tree. We can also provide a `-f, --force` option to eliminate this restriction.

I'm also thinking that, when the project is not a git repo, it shouldn't allow you to generate code in a directory that already contains files. Again, this restriction can be removed using `-f, --force`.

derberg:
Features:
- block generation if directory is a git repo and some changes are unstaged
- block generation if directory is not an empty directory
- add flag to force generation
- stuff not related to this issue https://github.com/asyncapi/generator/pull/223/files#diff-b58e2f1cee2acff55355343774e6d576R140 -> I changed it as when I wanted to learn what entrypoint is (was not clear for me after reading the docs) I noticed that it actually could not work as this was before `path.dirname(path.relative(this.templateDir), entrypointPath)` and `path.relative` requires 2 params

Fixes #180

### DIFF
--- a/lib/generator.js
+++ b/lib/generator.js
@@ -702,7 +702,7 @@ class Generator {
   /**
    * Check if given directory is a git repo with unstaged changes or is not empty
    * @private
-   * @param  {String} [dir] Directory that needs to be tested for a given condition.
+   * @param  {String} dir Directory that needs to be tested for a given condition.
   */
   async verifyTargetDir(dir) {
     const gitInfo = git(dir);
@@ -720,11 +720,11 @@ class Generator {
         const modifiedFiles = gitStatus.modified;
         const hasModifiedUstagedFiles = (modifiedFiles.filter(e => stagedFiles.indexOf(e) === -1)).length !== 0;
 
-        if (hasModifiedUstagedFiles || hasUntrackedUnstagedFiles) throw new Error(`"${this.targetDir}" is in a git repository with unstaged changes`);
+        if (hasModifiedUstagedFiles || hasUntrackedUnstagedFiles) throw new Error(`"${this.targetDir}" is in a git repository with unstaged changes. Please commit your changes before proceeding or use the --force-write flag to skip this rule (not recommended).`);
       } else {
         const isDirEmpty = (await readDir(dir)).length === 0;
 
-        if (!isDirEmpty) throw new Error(`"${this.targetDir}" is not an empty directory. You might override your work.`);
+        if (!isDirEmpty) throw new Error(`"${this.targetDir}" is not an empty directory. You might override your work. To skip this rule, please make your code a git repository or use the --force-write flag (not recommended).`);
       }
     } catch (e) {
       throw e;
--- a/lib/generator.js
+++ b/lib/generator.js
@@ -714,11 +714,11 @@ class Generator {
         const gitStatus = await gitInfo.status();
 
         //New files are not tracked and not visible as modified
-        const hasUntrackedUnstagedFiles = gitStatus.not_added.length != 0;
+        const hasUntrackedUnstagedFiles = gitStatus.not_added.length !== 0;
 
         const stagedFiles = gitStatus.staged;
         const modifiedFiles = gitStatus.modified;
-        const hasModifiedUstagedFiles = (modifiedFiles.filter(e => stagedFiles.indexOf(e) === -1)).length != 0;
+        const hasModifiedUstagedFiles = (modifiedFiles.filter(e => stagedFiles.indexOf(e) === -1)).length !== 0;
 
         if (hasModifiedUstagedFiles || hasUntrackedUnstagedFiles) throw new Error(`"${this.targetDir}" is in a git repository with unstaged changes`);
       } else {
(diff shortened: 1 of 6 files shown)


lib/generator.js
```diff
@@ -63,8 +65,9 @@ class Generator {
    * @param {String[]} [options.noOverwriteGlobs] List of globs to skip when regenerating the template.
    * @param {String[]} [options.disabledHooks] List of hooks to disable.
    * @param {String} [options.output='fs'] Type of output. Can be either 'fs' (default) or 'string'. Only available when entrypoint is set.
+   * @param {String} [options.forceWrite] Force writing of the generated files to given directory even if it is a git repo with unstaged files or not empty dir.
```
fmvilas: Should this parameter have a default value of `false`?

lib/generator.js
```diff
@@ -63,8 +65,9 @@ class Generator {
...
    */
-  constructor(templateName, targetDir, { templatesDir, templateParams, entrypoint, noOverwriteGlobs, disabledHooks, output = 'fs' } = {}) {
+  constructor(templateName, targetDir, { templatesDir, templateParams, entrypoint, noOverwriteGlobs, disabledHooks, output = 'fs', forceWrite } = {}) {
```
fmvilas: If the comment above makes sense, it could be:

```suggestion
  constructor(templateName, targetDir, { templatesDir, templateParams, entrypoint, noOverwriteGlobs, disabledHooks, output = 'fs', forceWrite = false } = {}) {
```

lib/generator.js
```diff
@@ -692,6 +698,38 @@ class Generator {
 
     return Promise.all(promises);
   }
+
+  /**
+   * Check if given directory is a git repo with unstaged changes or is not empty
+   * @private
+   * @param  {String} [dir] Directory that needs to be tested for a given condition.
```
fmvilas: I think `dir` is not optional, so we should remove the square brackets:

```suggestion
   * @param  {String} dir Directory that needs to be tested for a given condition.
```

lib/generator.js
```diff
@@ -692,6 +698,38 @@ class Generator {
...
+        const gitStatus = await gitInfo.status();
+
+        //New files are not tracked and not visible as modified
+        const hasUntrackedUnstagedFiles = gitStatus.not_added.length !== 0;
+
+        const stagedFiles = gitStatus.staged;
+        const modifiedFiles = gitStatus.modified;
+        const hasModifiedUstagedFiles = (modifiedFiles.filter(e => stagedFiles.indexOf(e) === -1)).length !== 0;
+
+        if (hasModifiedUstagedFiles || hasUntrackedUnstagedFiles) throw new Error(`"${this.targetDir}" is in a git repository with unstaged changes`);
```
fmvilas: We could improve the message here, with things like "Please commit your changes before proceeding or use the --force-write flag to skip this rule (not recommended).".

lib/generator.js
```diff
@@ -692,6 +698,38 @@ class Generator {
...
+      } else {
+        const isDirEmpty = (await readDir(dir)).length === 0;
+
+        if (!isDirEmpty) throw new Error(`"${this.targetDir}" is not an empty directory. You might override your work.`);
```
fmvilas: We could improve the message here, with something like "To skip this rule, please make your code a git repository or use the --force-write flag (not recommended).".

Not sure "rule" is the right word here and above tho.

lib/generator.js
derberg: actually I removed it in my last commit because if not specified, it will anyway be `undefined` and will still work as expected. Condition will not pass until `true` anyway

lib/generator.js
derberg: look on my previous comment, I had it but removed on purpose. What you think?

lib/generator.js
derberg: `rule` sounds good to me

derberg:
Not so rusty after all 😌 
I made all the changes you suggested. Let me know what you think about my comment to the topic of default value for the force flag

lib/generator.js
fmvilas: Yes, but then [this documentation](https://github.com/asyncapi/generator/pull/223/files#diff-88333ba11114e014e87198d04680144dR33) will not render that the default value is `false`. Just because `undefined` casts to `false` when checking on conditions it doesn't mean `undefined` is correct, right? Think about it as if we had types. Some day we'll have them 😉

derberg: I like `undefined`, don't treat it like this, it is enough that all those Java devs are mocking it 😄 
I will revert to get default back in 😄

derberg: @fmvilas default added, check out my last commit

fmvilas: LOL, I like it too. I just don't think this is its right place. Thanks man. Resolving.

PR #226, opened 2020-02-21
Add support for gitignore to git protection feature

derberg:
I started using this feature because I started working on latest master for another issue....and I got annoyed that there is no support for case:
_I have a git repo and my output directory is in .gitignore which mean no harm if generator generates files there_

### DIFF
--- a/lib/generator.js
+++ b/lib/generator.js
@@ -700,19 +700,29 @@ class Generator {
   }
 
   /**
-   * Check if given directory is a git repo with unstaged changes or is not empty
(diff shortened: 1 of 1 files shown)


fmvilas:
Wouldn't it be similar to the case when the directory is not a git repo? I mean, should we check if the directory isn't empty?

derberg:
@fmvilas I don't think so, it is like another case. User knows what he has in gitignore meaning what is not handled by git. My dev experience is, I do not want to have this protection

PR #227, opened 2020-02-24
Add support for server variables in Nodejs template

### LINKED ISSUE #182: Support server URL variables
Currently, the generator doesn't support server URL variables. For instance, when we have a server like this:

```yaml
servers:
  production:
    url: test.mosquitto.org:{port}
    protocol: mqtt
    description: Test broker
    variables:
      port:
        description: Secure connection (TLS) is available through port 8883.
        default: '1883'
        enum:
          - '1883'
          - '8883'
```

The generated code tries to connect to `mqtt://test.mosquitto.org:{port}`. It would be great if it could replace `port` by a generated value, i.e., the default value or a valid one, based on the variable definition.

derberg:
Fixes #182 

- Updates docs for creating templates in area of hooks and partials
- Extends the test doc to have a server variable
- Adds support for variables in the `servers` object to the Node.js template. An example is based on the `port` variable

I could test MQTT protocol as it is already in the streetlights.yml. After my changes I generated the Nodejs example and started it. It received a message sent to `mqtt pub -t 'smartylighting/streetlights/1/0/event/{streetlightId}/lighting/measured' -h 'test.mosquitto.org' -p 1883 -m '{"id": 1"}'`

Important notes to how the `replaceVariablesWithValues` filter works:
- if variable is in URL but not declared, it stays
- if variable is in URL and declared, it picks up default, otherwise first enum element, otherwise is stays unchanged
It is pretty generic, works on any string, not just url of course.

I need some suggestions on how to test if my changes work with other brokers. How you do it? should I just grab Kafka and others on my local and play or?

So far I made sure that in `common.yml` config whenever `url()` is grabbed and filter (that gets a string and created new URL object from it) is used, I add the variable parsing filter first to not get errors. If filters are shared among other templates, I'm afraid there might be some issues in other templates...:thinking: I ran generation for all of them and got no errors, but I don't know if it means all is fine

### DIFF
--- a/templates/nodejs/.filters/all.js
+++ b/templates/nodejs/.filters/all.js
@@ -193,4 +193,55 @@ module.exports = ({ Nunjucks }) => {
   Nunjucks.addFilter('filenamify', (string, options) => {
     return filenamify(string, options || { replacement: '-', maxLength: 255 });
   });
+
+  /**
+   * Replaces variables in the server url with its declared values. Default or first enum in case of default is not declared
+   * Replace is performed only if there are variables in the URL and they are declared for a server
+   * @private
+   * @param {String} url The server url value.
+   * @param {Object} serverVariables object containins server variables.
+   * @return {String}
+   */
+  function replaceVariablesWithValues(url, serverVariables) {
+    const urlVariables = getVariablesNamesFromUrl(url);
+    const isVariableDeclared =
+      urlVariables.filter(el => serverVariables.hasOwnProperty(el[1])) !== 0;
+
+    if (urlVariables.length !== 0 && isVariableDeclared) {
+      let value;
+      let newUrl = url;
+
+      urlVariables.forEach(el => {
+        value = getVariableValue(serverVariables, el[1]);
+
+        if (value) {
+          newUrl = newUrl.replace(el[0], value);
+        }
+      });
+      return newUrl;
+    }
+    return url;
+
+    function getVariablesNamesFromUrl(url) {
+      let result = [],
+        array;
+      const regEx = /{([^}]+)}/g;
+  
+      while ((array = regEx.exec(url)) !== null) {
+        result.push([array[0], array[1]]);
+      }
+  
+      return result;
+    }
+  
+    function getVariableValue(object, variable) {
+      const keyValue = object[variable]._json;
+  
+      if (keyValue) return keyValue.default || (keyValue.enum && keyValue.enum[0]);
+    }
+  }
+
+  Nunjucks.addFilter('replaceVariablesWithValues', (string, options) => {
+    return replaceVariablesWithValues(string, options);
+  });
 };
--- a/templates/nodejs/.filters/all.js
+++ b/templates/nodejs/.filters/all.js
@@ -204,10 +204,14 @@ module.exports = ({ Nunjucks }) => {
    */
   function replaceVariablesWithValues(url, serverVariables) {
     const urlVariables = getVariablesNamesFromUrl(url);
-    const isVariableDeclared =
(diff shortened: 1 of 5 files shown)


templates/nodejs/.filters/all.js
```diff
@@ -192,4 +192,55 @@ module.exports = ({ Nunjucks, _ }) => {
   Nunjucks.addFilter('filenamify', (string, options) => {
     return filenamify(string, options || { replacement: '-', maxLength: 255 });
   });
+
+  /**
+   * Replaces variables in the server url with its declared values. Default or first enum in case of default is not declared
+   * Replace is performed only if there are variables in the URL and they are declared for a server
+   * @private
+   * @param {String} url The server url value.
+   * @param {Object} serverVariables object containins server variables.
```
fmvilas:
```suggestion
   * @param {Object} serverVariables object containing server variables.
```

templates/nodejs/.filters/all.js
```diff
@@ -192,4 +192,55 @@ module.exports = ({ Nunjucks, _ }) => {
...
+   * @return {String}
+   */
+  function replaceVariablesWithValues(url, serverVariables) {
+    const urlVariables = getVariablesNamesFromUrl(url);
+    const isVariableDeclared =
+      urlVariables.filter(el => serverVariables.hasOwnProperty(el[1])) !== 0;
```
fmvilas: Did you mean this?

```suggestion
      urlVariables.filter(el => serverVariables.hasOwnProperty(el[1])).length !== 0;
```

derberg: nope, but it is still a bug, thanks for pointing out. `hasOwnProperty` returns Boolean. I fixed it and renamed the variable to reflect the reality

derberg:
@fmvilas take a look at my comment. Fixes done
I can do the setup, it is fine, but if you point me to specific tools that I can use, that would be great, things that you are using when working on a template. For Kafka that should be easy to find something, but for amqp? that would be rabbitmq?

btw `"supportedProtocols": ["amqp", "mqtt", "kafka", "ws"],` I guess ws is not supported by this template? this is why we have Node.js-ws template?

docker-compose is a must-have long term as it will be needed for CI testing anyway. But it is not in scope of this PR imho

fmvilas:
WebSockets is supported by Node.js template but it's not using Socket.io and it's a bit different than the other. I'll do a cleanup as soon as I find time. Support for WS in Node.js template was added after I created the nodejs-ws template. That's why both exist.

> docker-compose is a must-have long term as it will be needed for CI testing anyway. But it is not in scope of this PR imho

Yup, totally out of scope. Should we create an issue not to forget about it?

derberg:
@fmvilas I will create one for sure

I will not merge until I test with Kafka and RabbitMQ to make sure I did not break anything.
Do you have some sample spec I could use? I recall from one of the conferences that you for sure had a Kafka example. I never played with those, will be fun

fmvilas:
Yeah, RabbitMQ would be ok for AMQP. As per the examples, just change the protocol and everything else should work the same way.

templates/nodejs/config/common.yml
```diff
@@ -16,7 +16,7 @@ default:
       exchange:
       username:
       password:
-      host: {{ asyncapi.server(params.server).url() | host }}
+      host: amqp://{{ asyncapi.server(params.server).url() | replaceVariablesWithValues(asyncapi.server(params.server).variables())  | host }}
```
fmvilas: It shouldn't have `amqp://` on the `host`. Hosts don't contain protocol information. You probably have to add `amqp://` where this value is being used (if needed).

derberg: you are right, works well without it......makes me thing why I added it as I don't remember

PR #233, opened 2020-03-02
Fix link to recipes in the readme

derberg:


derberg:
@jonaslagoni I should have too, and actually, proper GitHub Action in place should do it for both of us

PR #244, opened 2020-03-09
Node 7 no longer maintained. Moved to 12 (LTS)

jorge07:


derberg:
@jorge07 thanks for your contribution! I'm pretty sure it works fine on 12 as I used it on 13, but I assume you bumped to 12 because LTS for 13 is not yet launched.

jorge07:
@derberg yes, for regenerated code I think LTS works better. 
https://nodejs.org/en/about/releases/

PR #247, opened 2020-03-11
Register optional schema parsers

fmvilas:
After [extracting RAML and OpenAPI schema parsers](https://github.com/asyncapi/parser-js/pull/44) from the JS parser, we must register them explicitly now.

### DIFF
--- a/lib/generator.js
+++ b/lib/generator.js
@@ -4,7 +4,10 @@ const util = require('util');
 const xfs = require('fs.extra');
 const walkSync = require('klaw-sync');
 const minimatch = require('minimatch');
-const { parse, AsyncAPIDocument } = require('asyncapi-parser');
+const parser = require('asyncapi-parser');
+const { parse, AsyncAPIDocument } = parser;
+const ramlDtParser = require('@asyncapi/raml-dt-schema-parser');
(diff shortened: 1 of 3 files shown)


derberg:
@fmvilas I guess we are waiting with this one for `parser-js` release and you will have to bump version of it in generator?

fmvilas:
There's no need to wait, @derberg. The schema parsers are the same and registering twice doesn't make any harm.

derberg:
@fmvilas I have mixed feelings about it, we are not in a rush and can wait, or?

fmvilas:
We can wait for merging but not necessarily approving 😄 

The parser PR doesn't introduce any breaking change in the API. What we're doing here is comparable to:

```js
let a;
a = 10
...
a = 10
```

You see `a` is assigned twice but it doesn't really make any difference :)

derberg:
Sure, approval, of course, once the parser version is bumped 😈

fmvilas:
Can we approve this now? 😄 👉 https://github.com/asyncapi/parser-js/releases/tag/v0.15.0

fmvilas:
Thanks!

PR #248, opened 2020-03-11
Updated nodejs version.

damaru-inc:
I changed the nodejs version in the Dockerfile to 12.16.1 (the current version 12 in Docker Hub) and I changed the version in the README requirements section to that version.

This allows template writers to use JavaScript E6 features such as classes.

Slack discussion: https://asyncapi.slack.com/archives/CQVJXFNQL/p1583949648004600?thread_ts=1583869811.002400&cid=CQVJXFNQL

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM node:12.16.1-alpine3.11
+FROM node:12-alpine
 
 WORKDIR /app
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM node:13.8.0-alpine3.11
+FROM node:12.16.1-alpine3.11
 
 WORKDIR /app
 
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM node:13.8.0-alpine3.11
+FROM node:12.16.1-alpine3.11
```
fmvilas: I'd recommend using:

```suggestion
FROM node:12-alpine
```

README.md
```diff
@@ -81,7 +81,7 @@ See [authoring templates](docs/authoring.md) and the [list of templates recipes]
 
 ## Requirements
 
-* Node.js v8.5+
+* Node.js v12.16.1+
```
fmvilas: What about this?

```suggestion
* Node.js v12.16+
```

damaru-inc: Sure, I'll make those changes.

PR #261, opened 2020-03-20
feat: Add support for remote templates

### LINKED ISSUE #251: Make generator support installing remote templates
Changes this issue must introduce:

1. The syntax must now be something like `ag [options] <asyncapi> <templatePath|templateUrl|templatePackageName|templateGitUrl>`
1. All templates are now treated as "remote".
1. Rename the `-w, --watch` flag to `--watch-template`. Let's leave `-w, --watch` for "watching" the AsyncAPI file. Show a warning when the template directory under `node_modules` is not a symlink (i.e., it's a remote template).
1. Slightly change the meaning of `--force-install`. It will now install the template **and** its dependencies.
1. Remove the `-t, --templates <templateDir>` flag. Keep reading to know why.

Installation is handled by the `npm install` command, which supports the following syntax:
```
npm install [<@scope>/]<name>
npm install [<@scope>/]<name>@<tag>
npm install [<@scope>/]<name>@<version>
npm install [<@scope>/]<name>@<version range>
npm install <git-host>:<git-user>/<repo-name>
npm install <git repo url>
npm install <tarball file>
npm install <tarball url>
npm install <folder>
```

Therefore, the current `-t, --templates <templateDir>` flag can be removed and will always point to the `node_modules` folder of the generator. That will allow us to benefit from the npm cache system too.

### What if I have a local template? Will I have to place it inside `node_modules`?

No, you can keep working as usual. `npm install` will create a symlink to your folder and will place it under `node_modules`.

### Why depending on npm install?

It is much easier for us to rely on npm to do the job. We get a lot for free, including the cache —and a mechanism to skip it. In any case, we already had a dependency on npm when we decided to convert templates to npm packages (#219).

fmvilas:
**Description**

Changes this PR introduces:

1. All templates are now treated as "remote".
1. Renames the `-w, --watch` flag to `--watch-template`. Let's leave `-w, --watch` for "watching" the AsyncAPI file. Shows a warning when the template directory under `node_modules` is not a symlink (i.e., it's a remote template).
1. Slightly changes the meaning and name of `--force-install`. Its name is `-i, --install` now. It will install the template **and** its dependencies.
1. Removes the `-t, --templates <templateDir>` flag. Keep reading to know why.
1. **Removes the "templates" directory 😱**. Don't panic, they have been extracted to their own repo: https://github.com/search?q=topic%3Aasyncapi+topic%3Agenerator+topic%3Atemplate

This is how it feels now on the CLI:

```bash
# Use template from npm (skip installation if already installed)
ag asyncapi.yaml @asyncapi/html-template -o output

# Use template from npm (forces installation)
ag asyncapi.yaml @asyncapi/html-template -i -o output
ag asyncapi.yaml @asyncapi/html-template -io output
ag asyncapi.yaml @asyncapi/html-template -o output --install

# Use template from npm (watches template for changes)
ag asyncapi.yaml @asyncapi/html-template -o output --watch-template

# Use template from filesystem (skip installation if already installed)
ag asyncapi.yaml ../my-templates/asciidoc -o output

# Use template from filesystem (forces installation)
ag asyncapi.yaml ../my-templates/asciidoc -io output
ag asyncapi.yaml ../my-templates/asciidoc -o output --install

# Use template from filesystem (watches template for changes)
ag asyncapi.yaml ../my-templates/asciidoc -o output --watch-template
```

**Related issue(s)**
Fixes #251 

**To do**
- [x] Update docs


cli.js
```diff
@@ -52,14 +53,13 @@ program
...
+  .option('-i, --install', 'installs the template and its dependencies (defaults to false)')
   .option('-n, --no-overwrite <glob>', 'glob or path of the file(s) to skip when regenerating', noOverwriteParser)
+  .option('-o, --output <outputDir>', 'directory where to put the generated files (defaults to current directory)', parseOutput, process.cwd())
   .option('-p, --param <name=value>', 'additional param to pass to templates', paramParser)
-  .option('-t, --templates <templateDir>', 'directory where templates are located (defaults to internal templates directory)', Generator.DEFAULT_TEMPLATES_DIR, path.resolve(__dirname, 'templates'))
-  .option('--force-install', 'forces the installation of the template dependencies. By default, dependencies are installed and this flag is taken into account only if `node_modules` is not in place.')
   .option('--force-write', 'force writing of the generated files to given directory even if it is a git repo with unstaged files or not empty dir (defaults to false)')
+  .option('--watch-template', 'watches the template directory and the AsyncAPI document, and re-generate the files when changes they occur')
```
fmvilas: Also took the opportunity to short params alphabetically ☝️

lib/generator.js
```diff
@@ -275,22 +273,65 @@ class Generator {
...
+
+          if (isFileSystemPath(this.templateName)) {
+            const pkg = require(path.resolve(this.templateName, 'package.json'));
+            installedPkg = require(path.resolve(DEFAULT_TEMPLATES_DIR, pkg.name, 'package.json'));
+            if (installedPkg.name !== pkg.name) throw new Error('Package not installed');
+          } else { // Template is not a filesystem path...
+            const templatePath = path.resolve(DEFAULT_TEMPLATES_DIR, this.templateName);
+            if (await isLocalTemplate(templatePath)) {
+              const { resolvedLink } = await getLocalTemplateDetails(templatePath);
+              console.warn(`WARNING: this template is already installed and pointing to your filesystem at ${resolvedLink}.`);
```
fmvilas: Not sure if this should be a warning or an "info" message.

fmvilas: Also, it should probably say "Skipping installation..." or something similar. What do you think?

jonaslagoni: What about we add a info message before searching locally and then another info message here that it was found locally and you should use the force install flag to update the template package 🤔 ?

README.md
```diff
@@ -34,40 +34,37 @@ asyncapi/generator -o ./output asyncapi.yml markdown
...
 Specify where to put the result:
 ```bash
-ag -o ./docs asyncapi.yaml markdown
+ag asyncapi.yaml @asyncapi/html-template -o ./docs
 ```
 
 Passing parameters to templates:
 ```bash
-ag -o ./docs --param title='Hello from param' asyncapi.yaml markdown
+ag asyncapi.yaml @asyncapi/html-template -o ./docs -p title='Hello from param'
```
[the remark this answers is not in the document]
fmvilas: This must be a problem with your computer. I'm not able to reproduce. I found this issue: https://github.com/npm/cli/issues/613.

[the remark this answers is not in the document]
derberg: what `npm` version you have? I `6.13.6`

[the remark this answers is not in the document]
derberg: when I do `./cli.js ./test/docs/ws.yml @asyncapi/nodejs-ws-template -o ./output -p server=localhost --force-write` it works fine 🤔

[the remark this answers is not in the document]
fmvilas: I'm in node 13.11.0 and npm 6.13.7 (for this test). It works well.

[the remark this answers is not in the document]
fmvilas: Thanks. This has been fixed now. Can you try again?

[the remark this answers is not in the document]
fmvilas: With both empty and non-empty output directory. Can you try again with the fixes I did?

lib/generator.js
```diff
@@ -275,22 +274,65 @@ class Generator {
...
+   * @param {Boolean} [force=false] Whether to force installation (and skip cache) or not.
+   */
+  installTemplate(force = false) {
+    return new Promise(async (resolve, reject) => {
+      if (!force) {
+        try {
+          let installedPkg;
+
+          if (isFileSystemPath(this.templateName)) {
+            const pkg = require(path.resolve(this.templateName, 'package.json'));
```
derberg: you should use `PACKAGE_JSON_FILENAME` here and in few other places

lib/generator.js
```diff
@@ -275,22 +274,65 @@ class Generator {
...
+            installedPkg = require(path.resolve(DEFAULT_TEMPLATES_DIR, pkg.name, 'package.json'));
+            if (installedPkg.name !== pkg.name) throw new Error('Package not installed');
```
derberg: can you explain when it can happen as I could not figure out? and why we throw error instead of info + run installation?

lib/generator.js
derberg: tbh I'm not sure exactly when will I get this warning anyway? How can I have a situation that remote template is actually local?

fmvilas: For instance, imagine we're working locally on our HTML template and you run:

```
ag asyncapi.yaml ~/html-template
```

That template contains a package.json with the name `@asyncapi/html-template` and therefore npm will install it as `node_modules/@asyncapi/html-template`. However, because it's a local template, it will use a soft link pointing to `~/html-template`.

Right afterward, you run:

```
ag asyncapi.yaml @asyncapi/html-template
```

And the generator finds that the template has been already installed **but** it's a symlink and therefore it knows it's a local template, even though the name (`@asyncapi/html-template`) doesn't look like it is. This is that use case 😄

lib/generator.js
fmvilas: Good catch. This is not needed anymore. The reason it was throwing the error is for the parent try/catch to catch the error and avoid calling the `resolve` function a few lines below.

lib/generator.js
derberg: well explained, thanks, makes a lot of sense. This is `info` for me, not a `warning` then

fmvilas:
@derberg all comments have been addressed now.

> when I checked with git and tar, the installation is always triggered, even if I already used the template once:

In order to know the name of the package —and know if it has been installed already— npm has to first download it to look into the package.json file. That's why it's always downloaded :)

PR #290, opened 2020-04-08
fix: node_modules created on current dir

### LINKED ISSUE #289: Do not create node_modules directory
It would be ideal to not leave behind a `node_modules` directory in the directory that you run the "ag" command from to generate an app. I'm not a node dev so I'm not sure if there is an established pattern for things like this but I would expect that it should be in a hidden directory in the user's home or maybe under /tmp?

fmvilas:
**Description**
It was creating a `node_modules` directory on the directory the command was executed. It's fixed now.

**Related issue(s)**
Fixes #289

### DIFF
--- a/lib/generator.js
+++ b/lib/generator.js
@@ -305,6 +305,7 @@ class Generator {
       npmi({
         name: this.templateName,
         install: force,
+        path: ROOT_DIR,
         pkgName: 'dummy value so it does not force installation always',
         npmLoad: {
           loglevel: 'http',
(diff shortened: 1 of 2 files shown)


fmvilas:
FYI @Mrc0113

lib/generator.js
```diff
@@ -305,6 +305,7 @@ class Generator {
       npmi({
         name: this.templateName,
         install: force,
+        path: ROOT_DIR,
```
fmvilas, line 308: To clarify: ROOT_DIR is the directory where the library is currently installed. It's not the system's root dir 😄

derberg: makes sense and `npmi` default is `path: '.'`

PR #298, opened 2020-04-17
feat: support filters from external library

### LINKED ISSUE #212: Avoid code repetition in templates filters
**Description:**

Templating in the generator is based on Nunjucks. One of its features is `filters` that allow you to apply custom functions on variables https://mozilla.github.io/nunjucks/templating.html#filters.

Generic filters/functions should be abstracted to one common location and shared with other filters. There is no specific proposal on how to do it, it can be just a part of the generator but in one common place, specific filters should stay in current locations. 

We also need to take into account users that do not use existing templates located in repository but instead, they keep them privately. We need to have a good solution here, that will allow these users to easily import generic filters to their templates if they want, so they can clean up their filters.

**Reason:**

The current issue is that those filters are not shared but every template have own filters, even the basic ones.

Example: a `log` function is duplicated here https://github.com/asyncapi/generator/blob/master/templates/html/.filters/all.js#L116 and https://github.com/asyncapi/generator/blob/master/templates/markdown/.filters/all.js#L3.

derberg:
**Description**

- support filters from an external library. New template config parameter is added, `filters` that accepts an array of libraries that you can additionally register as filters for your template.
- generator now reads any function and then registers it as a filter so it is easier to write functions for templates and test them
- extracted whole logic related to filters registration to a separate file as `generator.js` is super overwhelming so using the opportunity that I worked on filters I moved them out

How to test:
`./cli.js test/docs/streetlights.yml https://github.com/derberg/html-template/#use-filters-lib -o output --force-write`

## Migration Guide
Your templates will not work with the latest generator. In your template, in filter directory you need to refactor your functions to follow new structure or remove filters that are already extracted to https://github.com/asyncapi/generator-filters

* To use https://github.com/asyncapi/generator-filters you need to add the following property to your template configuration file
```
  "filters": [
    "@asyncapi/generator-filters"
  ]
```
and have this package in `package.json` dependencies
```
 "@asyncapi/generator-filters": "^1.0.0"
```
* To refactor your existing filters just make sure your filters are normal JS functions and you export them as part of the module. Have a look on one of our official templates to see how new filters look like https://github.com/asyncapi/html-template/blob/master/filters/all.js. You do not have to change the logic, they just need to be refactored as they no longer need Nunjucks environment as input, which also means they are much easier to test now

**Related issue(s)**
Fixes #212

### DIFF
--- a/docs/authoring.md
+++ b/docs/authoring.md
@@ -64,7 +64,7 @@ Schema name is 'people' and properties are:
 ```
 ## Filters
 
-A filter is a helper function that you can create to perform complex tasks. They are JavaScript files that register one or many [Nunjuck filters](https://mozilla.github.io/nunjucks/api.html#custom-filters). The generator parses all the files in the `.filters` directory. Functions exported from these files are registered as filters.
+A filter is a helper function that you can create to perform complex tasks. They are JavaScript files that register one or many [Nunjuck filters](https://mozilla.github.io/nunjucks/api.html#custom-filters). The generator parses all the files in the `filters` directory. Functions exported from these files are registered as filters.
 
 You can use the filter function in your template as in the following example:
 
@@ -72,7 +72,7 @@ You can use the filter function in your template as in the following example:
 const {{ channelName | camelCase }} = '{{ channelName }}';
 ```
 
-In case you have more than one template and want to reuse filters, you can put them in a single library. You can configure such a library in the template configuration under `filter` property. You can also use the official AsyncAPI [filters library](https://github.com/asyncapi/generator-filters). To learn how to add such filters to configuration [read more about the configuration file](#configuration-file).
+In case you have more than one template and want to reuse filters, you can put them in a single library. You can configure such a library in the template configuration under `filters` property. You can also use the official AsyncAPI [filters library](https://github.com/asyncapi/generator-filters). To learn how to add such filters to configuration [read more about the configuration file](#configuration-file).
 
 ## Hooks
 
--- a/docs/authoring.md
+++ b/docs/authoring.md
@@ -8,7 +8,7 @@ The AsyncAPI generator has been built with extensibility in mind. The package us
 1. The template can have own dependencies. Just create `package.json` for the template. The generator makes sure to trigger the installation of dependencies.
 1. Templates may contain multiple files. Unless stated otherwise, all files will be rendered.
 1. The template engine is [Nunjucks](https://mozilla.github.io/nunjucks).
-1. Templates may contain `filters` or helper functions. They must be stored in the `.filters` directory under the template directory. [Read more about filters](#filters).
+1. Templates may contain [Nunjucks filters or helper functions](https://mozilla.github.io/nunjucks/templating.html#builtin-filters). [Read more about filters](#filters).
 1. Templates may contain `hooks` that are functions invoked after the generation. They must be stored in the `.hooks` directory under the template directory. [Read more about hooks](#hooks).
 1. Templates may contain `partials` (reusable chunks). They must be stored in the `.partials` directory under the template directory. [Read more about partials](#partials).
 1. Templates may have a configuration file. It must be stored in the template directory and its name must be `.tp-config.json`. [Read more about the configuration file](#configuration-file).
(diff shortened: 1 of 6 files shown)


docs/authoring.md
```diff
@@ -25,41 +25,16 @@ The AsyncAPI generator has been built with extensibility in mind. The package us
    
 ## Filters
 
-A filter is a helper function that you can create to perform complex tasks. They are JavaScript files that register one or many [Nunjuck filters](https://mozilla.github.io/nunjucks/api.html#custom-filters). The generator will parse all the files in the `.filters` directory.
+A filter is a helper function that you can create to perform complex tasks. They are JavaScript files that register one or many [Nunjuck filters](https://mozilla.github.io/nunjucks/api.html#custom-filters). The generator parses all the files in the `.filters` directory. Functions exported from these files are registered as filters.
```
fmvilas, line 28:
```suggestion
A filter is a helper function that you can create to perform complex tasks. They are JavaScript files that register one or many [Nunjuck filters](https://mozilla.github.io/nunjucks/api.html#custom-filters). The generator parses all the files in the `filters` directory. Functions exported from these files are registered as filters.
```

docs/authoring.md
```diff
@@ -25,41 +25,16 @@ The AsyncAPI generator has been built with extensibility in mind. The package us
...
-```
-
-And then you can use the filter in your template as follows:
+You can use the filter function in your template as in the following example:
 
 ```js
 const {{ channelName | camelCase }} = '{{ channelName }}';
 ```
 
+In case you have more than one template and want to reuse filters, you can put them in a single library. You can configure such a library in the template configuration under `filter` property. You can also use the official AsyncAPI [filters library](https://github.com/asyncapi/generator-filters). To learn how to add such filters to configuration [read more about the configuration file](#configuration-file).
```
fmvilas, line 36:
```suggestion
In case you have more than one template and want to reuse filters, you can put them in a single library. You can configure such a library in the template configuration under `filters` property. You can also use the official AsyncAPI [filters library](https://github.com/asyncapi/generator-filters). To learn how to add such filters to configuration [read more about the configuration file](#configuration-file).
```

lib/filtersRegistry.js
```diff
@@ -0,0 +1,78 @@
...
+module.exports.registerFilters = async (nunjucks, templateConfig, templateDir) => {
+  registerConfigFilters(nunjucks, templateConfig);
+  await registerLocalFilters(nunjucks, templateDir);
+};
+/**
+   * Registers the template filters.
+   * @private
+   * @param {Object} nunjucks Nunjucks environment.
+   * @param {String} templateDir Directory where template is located.
+   */
```
fmvilas, line 15:
```suggestion
 * Registers the template filters.
 * @private
 * @param {Object} nunjucks Nunjucks environment.
 * @param {String} templateDir Directory where template is located.
 */
```

derberg:
@fmvilas your suggestions are in, thanks! I also removed one not needed function that was left from my development. What about the config? nothing agains? should I move other config values there with this PR or?

derberg:
@fmvilas forget my last question, I'm removing `node-config` as its purpose if for apps rather than libraries and it doesn't work nice in a scenario when the generator is used as a library, only complicates live. I will bring back global variable `FILTERS_DIRNAME` and pass it to the filters function like others arguments.

lib/filtersRegistry.js
```diff
@@ -0,0 +1,91 @@
+const path = require('path');
+const fs = require('fs');
+const xfs = require('fs.extra');
+const { isLocalTemplate } = require('./utils');
+
+/**
+ * Registers all template filters.
+ * @private
```
fmvilas, line 8: Is it really private? It's being exported below.

lib/filtersRegistry.js
```diff
@@ -0,0 +1,91 @@
...
+    });
+
+    walker.on('end', async () => {
+      resolve();
+    });
+  });
+}
+
+/**
+* Registers the additinally configured filters.
```
fmvilas, line 59:
```suggestion
* Registers the additionally configured filters.
```

Issue #280, opened 2020-04-01
Add mechanism to link templates versions to generator versions


fmvilas:
We should have a mechanism to figure out which versions of the generator are supported by each template.

Proposed solutions:
* Have a metadata field on the template config file indicating the minimum Generator version it works with and probably the maximum version too. It could be a version range or we directly assume it works with the whole major version.
* Have the generator as a dependency of the template. This way, people will install and use the template **from the template** instead of **from the generator**. Not incompatible with the previous approach.

fmvilas:
Added the `generator` param to all the templates (except the python and typescript ones):

1. https://github.com/asyncapi/java-spring-template/pull/23
1. https://github.com/asyncapi/java-spring-cloud-stream-template/pull/21
1. https://github.com/asyncapi/nodejs-template/pull/10
1. https://github.com/asyncapi/nodejs-ws-template/pull/6
1. https://github.com/asyncapi/markdown-template/pull/12
1. https://github.com/asyncapi/html-template/pull/18

PR #309, opened 2020-04-27
feat: add support for template-generator versioning requirements

fmvilas:
**Description**
It allows template owners to specify which version range of the generator they support. For instance, if we find a `.tp-config.json` file with the attribute `"generator": ">=0.36.0 <=1.0.0"` and the current version of the generator is `1.5.0`, it should trigger an error.

The format of the `generator` config property is a string representing semantic versioning ranges. See [semver](https://www.npmjs.com/package/semver) for more details.

**Related issue(s)**
Partially solves #280

### DIFF
--- a/lib/generator.js
+++ b/lib/generator.js
@@ -13,6 +13,8 @@ const Ajv = require('ajv');
 const filenamify = require('filenamify');
 const git = require('simple-git/promise');
 const npmi = require('npmi');
+const semver = require('semver');
+const packageJson = require('../package.json');
 const {
   convertMapToObject,
   isFileSystemPath,
@@ -669,8 +671,12 @@ class Generator {
    * @return {Promise}
    */
   async validateTemplateConfig(asyncapiDocument) {
-    const { parameters, supportedProtocols, conditionalFiles } = this.templateConfig;
+    const { parameters, supportedProtocols, conditionalFiles, generator } = this.templateConfig;
     let server;
+
+    if (typeof generator === 'string' && !semver.satisfies(packageJson.version, generator)) {
+      throw new Error(`This template is not compatible with this version of the generator (${packageJson.version}). It's compatible with the following version range: ${generator}.`);
+    }
 
     const requiredParams = [];
     Object.keys(parameters || {}).forEach(key => {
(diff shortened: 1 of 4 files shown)


lib/generator.js
```diff
@@ -13,6 +13,8 @@ const Ajv = require('ajv');
 const filenamify = require('filenamify');
 const git = require('simple-git/promise');
 const npmi = require('npmi');
+const semver = require('semver');
+const packageJson = require('../package.json');
```
derberg, line 17: I think better would be to read the version from environment variable `process.env.npm_package_version` as we do it here already https://github.com/asyncapi/generator/blob/master/package.json#L18

fmvilas: Why is this approach better than reading from the package.json file? It's not obvious that this environment variable exists. Also, if someone uses Yarn (or another package manager), does this environment variable exist too?

fmvilas:
> It would be a nice DX if we could enrich the info displayed in the console with information what is the latest version of the template compatible with a given version of the generator

This is hard to implement. I can offer the opposite: the latest generator version that works with the template. How do we find the latest version of the template compatible with the current version of the generator? Any ideas?

lib/generator.js
derberg: This is officially supported variable https://docs.npmjs.com/misc/scripts#packagejson-vars

At the end is it alwasy better to not read file if not needed.

Yarn also supports it https://classic.yarnpkg.com/en/docs/cli/version/#toc-version-lifecycle

fmvilas: I don't think this environment variable is available outside `preversion`, `version`, and `postversion`.

Screen Shot 2020-04-28 at 11 15 02

derberg:
@fmvilas I'm not saying it is easy, I'm just saying it is better :) we can have a separate issue for it for further discussion and research. At least official templates should have such functionality

lib/generator.js
fmvilas: Definitely they're not available. Check docs here: https://docs.npmjs.com/misc/scripts#environment. It's only when running a package script.

derberg: ok, forget it, it is too risky, works only if you run code with `npm` script, it will only complicate the development flow

derberg:
@fmvilas just update docs and I approve

fmvilas:
Docs updated 😊

Issue #308, opened 2020-04-25
Add hooks generate:before


Tenischev:
#### Reason/Context
Hooks launched before generation will allow modifying asyncApi before processing the template.
Also they could be used to remove some part of template which is not relevant for protocol in specified asyncapi.

#### Description
In theory, adding of `await this.launchHook('generate:before');` [here](https://github.com/asyncapi/generator/blob/master/lib/generator.js#L147) will solve this.

fmvilas:
I think that makes sense. When I added the "after" hook, I knew at some point we would have to add this one but couldn't come up with a use case. Thanks for raising this issue 🙂

PR #315, opened 2020-04-28
feat: add hooks before

Tenischev:
**Description**

- Add support of hook `generate:before`

Tested with an example of hook from authoring.md and java-spring template

**Related issue(s)**
Resolves: #308


docs/authoring.md
```diff
@@ -101,7 +101,9 @@ const {{ channelName | camelCase }} = '{{ channelName }}';
 
 ## Hooks
 
-Hooks are functions called by the generator on a specific moment in the generation process. For now there is one hook supported called `generate:after` that is called at the very end of the generation. The generator will parse all the files in the `.hooks` directory.
+Hooks are functions called by the generator on a specific moment in the generation process. 
+For now there is two hooks supported called `generate:before` and `generate:after` that are called before processing the template and at the very end of the generation respectively.
```
fmvilas, line 105: I think we can convert this into a table for better readability. Also, I'm sure new hooks will be added in the future. It will make things easier :)

Tenischev: @fmvilas please check new version

docs/authoring.md
```diff
@@ -76,8 +76,16 @@ In case you have more than one template and want to reuse filters, you can put t
...
 ## Hooks
 
-Hooks are functions called by the generator on a specific moment in the generation process. For now there is one hook supported called `generate:after` that is called at the very end of the generation. The generator will parse all the files in the `.hooks` directory.
+Hooks are functions called by the generator on a specific moment in the generation process.
+The following types of hooks are currently supported:
 
+|Hook name|Description|
+|---|---|
+| `generator:before` | Called after registration of all filters and before generator starts processing of the template.|
+| `generator:after` | Called at the very end of the generation. |
```
fmvilas, line 85:
```suggestion
| `generate:before` | Called after registration of all filters and before generator starts processing of the template.|
| `generate:after` | Called at the very end of the generation. |
```

docs/authoring.md
```diff
@@ -100,6 +108,27 @@ module.exports = register => {
   });
 };
 ```
+And here an example Hook that before generation switch publish and subscribe operations for each channel.
```
fmvilas, line 111:
```suggestion
And here an example Hook that before generation switches `publish` and `subscribe` operations for each channel.
```

Tenischev:
Thanks, suggestions committed.

PR #316, opened 2020-04-29
chore: add unit tests

### LINKED ISSUE #158: Introducing testing for the generator and cli

fmvilas:
**Description**
This PR add unit tests to the Generator 🙌 

**Related issue(s)**
Fixes #158


fmvilas:
Sounds good. I'll keep this file only about unit tests.

fmvilas:
Removed snapshot tests and only left unit tests. Added a few more. Let me know if it looks good so far before I continue.

jonaslagoni:
Nice! 

If you add coverage we should add it as a badge such as how [Mocha ](https://github.com/mochajs/mocha)does it with [coveralls ](https://coveralls.io/github/mochajs/mocha) 🤔 ? Would be a nice addition to the parser as well tbh 😄

fmvilas:
Yes, and it should actually be easy to migrate to Jest. Most of the syntax is the same.

fmvilas:
Yeah, in unit tests, you can only test that certain things have been called. We assume that they work as expected because we're not testing inner functionality but rather the surface API. For instance, we don't test if `mkdirp` or `fs.readFile` works. We trust it and don't test them because it's not part of our code. And yeah, snapshot testing will take care of this.

PR #321, opened 2020-04-30
feat: add objectSchema file template

### LINKED ISSUE #239: Schema file templating should return all schemas of type Object
**Reason**
From a templater's point of view it would make sense if the `$$schema$$` file template returned all object type schemas so it is possible to reduce generated code complexity, and making it possible for templates to easily take the object oriented approach.

**Hints**
- Discussion in [slack](https://asyncapi.slack.com/archives/CQVJXFNQL/p1583333853010200) about an issue with the `$$schema$$` file template not generating nested schemas together with a discussion in an [issue in the parser](https://github.com/asyncapi/parser-js/issues/32) triggered this issue.
- This issue is blocked by an [issue in the parser](https://github.com/asyncapi/parser-js/issues/39)

jonaslagoni:
**Description**
Added `$$objectSchema$$` to generate correct class files.

**Related issue(s)**
Fixes #239


lib/generator.js
```diff
@@ -366,11 +366,14 @@ class Generator {
    * @return {Promise}
    */
   generateDirectoryStructure(asyncapiDocument) {
+    const objectMap = {}
+    asyncapiDocument.allSchemas().forEach((schema, schemaId) => {if(schema.type() !== undefined && schema.type() === "object"){objectMap[schemaId] = schema}})
```
fmvilas, line 370:
```suggestion
    asyncapiDocument.allSchemas().forEach((schema, schemaId) => { if(schema.type() === 'object') objectMap[schemaId] = schema })
```

PR #323, opened 2020-04-30
feat: New changeFilename hook for customizing file names for template files 

### LINKED ISSUE #203: Separation of channels into files are not getting a suitable file name
So, when using NATS, and I am guessing other protocols, ones channel name would be something like `smartylighting.streetlights.1.0.action.{streetlightId}.dim` and `smartylighting.streetlights.1.0.dim.*`. However these are not really suitable as file names.

Suggestion 1 is to remove `.`, `*` and parameters so the file names would be `SmartylightingStreetlights10ActionDim` and `SmartylightingStreetlights10Dim`. One problem with this is when defining these two channels `smartylighting.streetlights.1.0.action.{streetlightId}.dim` and `smartylighting.streetlights.1.0.action.dim` they would conflict.

Suggestion 2 is to replace `*` with `All`, replace `{parameter}` with `Parameter` and remove `.`. So for `smartylighting.streetlights.1.0.action.{streetlightId}.dim` and `smartylighting.streetlights.1.0.dim.*` they would become `SmartylightingStreetlights10ActionStreetlightIdDim` and `SmartylightingStreetlights10DimAll`. I cant foresee any problems with this approach :thinking:

jonaslagoni:
**Description**
Changed the behavior of the template hook system so we are able to provide arguments and return types of hooks and added a new hook for customizing file names for template files.

**Related issue(s)**
Fixes #203


lib/generator.js
```diff
@@ -468,7 +468,17 @@ class Generator {
...
     try {
       const relativeBaseDir = path.relative(this.templateContentDir, baseDir);
-      const newFileName = fileName.replace(`\$\$${template}\$\$`, filenamify(name, { replacement: '-', maxLength: 255 }));
+
+      const beforeFileGenerationHookName = 'generate:changeFilename'
+      let filename = name
+      if(this.isHookAvailable(beforeFileGenerationHookName)){
+        let filenamesFromHooks = await this.launchHook(beforeFileGenerationHookName, {"filename": filename})
+        //Use the result of the first hook
+        filename = filenamesFromHooks[0];
```
jonaslagoni, line 477: IMO we only need use the first hook found. Should that somehow be stated in the documentation?

lib/generator.js
```diff
@@ -468,7 +468,17 @@ class Generator {
   async generateSeparateFile(asyncapiDocument, name, component, template, fileName, baseDir) {
     try {
       const relativeBaseDir = path.relative(this.templateContentDir, baseDir);
-      const newFileName = fileName.replace(`\$\$${template}\$\$`, filenamify(name, { replacement: '-', maxLength: 255 }));
+
+      const beforeFileGenerationHookName = 'generate:changeFilename'
```
jonaslagoni, line 472: Let me know if the hook makes sense for you or if it should be changed

docs/authoring.md
```diff
@@ -76,13 +76,14 @@ In case you have more than one template and want to reuse filters, you can put t
...
 
-|Hook name|Description|
-|---|---|
-| `generate:before` | Called after registration of all filters and before generator starts processing of the template.|
-| `generate:after` | Called at the very end of the generation. |
+|Hook name|Description| Return type | Arguments 
+|---|---|---|---|
+| `generate:before` | Called after registration of all filters and before generator starts processing of the template.| void | none
+| `generate:after` | Called at the very end of the generation. | void | none
+| `generate:changeFilename` | Called right before saving a file template. | string : a new filename for the generator to use for the file template | { "filename" : string }
```
jonaslagoni, line 86: does `{ "filename" : string }` makes sense for explaining the argument provided to the hook?

Tenischev, line 86: Maybe a small improve for description?
`Called right before saving a new file generated by [file template](https://github.com/asyncapi/generator/blob/master/docs/authoring.md#file-templates)`
And perhaps `{ "originalFilename" : string }` instead of `{ "filename" : string }` ?

lib/generator.js
```diff
@@ -468,7 +468,17 @@ class Generator {
...
+      let filename = name
+      if(this.isHookAvailable(beforeFileGenerationHookName)){
```
Tenischev, line 474: formatting pls ><

lib/generator.js
```diff
@@ -468,7 +468,17 @@ class Generator {
...
+        let filenamesFromHooks = await this.launchHook(beforeFileGenerationHookName, {"filename": filename})
+        //Use the result of the first hook
+        filename = filenamesFromHooks[0];
+      }else{
```
Tenischev, line 478: formatting pls ><

docs/authoring.md
jonaslagoni: I like it 👍

docs/authoring.md
```diff
@@ -129,6 +130,15 @@ module.exports = register => {
 };
 ```
 
+Example hook for changing the filename of a template file. Replaces all '-' characters with '_'.
+```js
+module.exports = register => {
+	register('generate:changeFilename', (generator, hookArguments) => {
+		const currentFilename = hookArguments.filename;
```
Tenischev, line 137: Should be `hookArguments.originalFilename` now

jonaslagoni: @Tenischev ... Only changed it in my template, good catch 😄

Tenischev:
Thanks, lgfm!

lib/generator.js
derberg: yes, as it kind of indicates that you should have only one hook that modifies names as we cannot assure which one will be triggered first

derberg:
@jonaslagoni I think I only have issue with the name of the hook. I think it should always indicate when it is called, like `generate:before` or `generate:after`. Although I agree that the scope of this hook is pretty limited... `filetemplateGenerate: before` ? Thoughts?

I also left one comment, and also please add info about the hook in **File template** section

fmvilas:
Yeah, name could be more aligned with existing syntax. Maybe `templateFileNameGenerate:before`? The problem I see with this name is that it sounds like you can customize all the template file names but it's not true, only those with $$ syntax can be customized. Maybe `dynamicFileName:generate`? 🤔

jonaslagoni:
Since we are calling it "File templates" in the [authoring templates](https://github.com/asyncapi/generator/blob/master/docs/authoring.md#file-templates) then what about `fileTemplateNameChange`? Since it is a direct hook not related to any state i.e. `before` etc I dont think we should add `:before` or anything else to it.

derberg:
@jonaslagoni I wanted to argue for `:before` but at the end you are right 😄 but it is not about file name change but about setting the name, so `setFileTemplateName` ?

jonaslagoni:
@derberg we can do that :+1: @fmvilas @tenischev any objections?

fmvilas:
Nope. Sounds good to me too.

Tenischev:
Ok for me as well.

jonaslagoni:
@fmvilas @derberg done.

jonaslagoni:
Had to solve a conflict, and then it merged all the code from master.. Not sure if it should be reverted?

derberg:
@jonaslagoni what do you mean? it shows only your changes? all good from my side, or you have somethign specific in mind?

jonaslagoni:
@derberg last night it showed all the changes from master as well. I guess it got corrected 😄

Issue #326, opened 2020-05-01
Improve error stack trace for errors in filters.


jonaslagoni:
#### Reason/Context
When developing templates sometimes you make an implementation mistake in a filter could be a simple `cannot call method of undefined`, however when that happens we receive no information about where in the filter the error occurred. It sometimes takes a lot of time to figure out where the error is really located, and sometimes the only approach is step by step console log and comment out functionality. 

As an example with the [html template](https://github.com/asyncapi/html-template) if you[ add a check to the following line](https://github.com/asyncapi/html-template/blob/6151bd4f16af558dd5da09adc4889c91266680ad/filters/all.js#L5 ) such as 
`obj.oneOf().length > 0 ||` it gives you the stack trace:
```
Something went wrong:
Template render error: (PATH@asyncapi\generator\node_modules\@asyncapi\html-template\template\index.html)
  Template render error: (PATH@asyncapi\generator\node_modules\@asyncapi\html-template\partials\content.html)
  Template render error: (PATH@asyncapi\generator\node_modules\@asyncapi\html-template\partials\operations.html)
  TypeError: Cannot read property 'length' of null
    at Object._prettifyError (PATH@asyncapi\generator\node_modules\nunjucks\src\lib.js:36:11)
    at PATH@asyncapi\generator\node_modules\nunjucks\src\environment.js:561:19
    at eval (eval at _compile (PATH@asyncapi\generator\node_modules\nunjucks\src\environment.js:631:18), <anonymous>:49:11)
    at PATH@asyncapi\generator\node_modules\nunjucks\src\environment.js:569:11
    at eval (eval at _compile (PATH@asyncapi\generator\node_modules\nunjucks\src\environment.js:631:18), <anonymous>:64:12)
    at PATH@asyncapi\generator\node_modules\nunjucks\src\environment.js:569:11
    at eval (eval at _compile (PATH@asyncapi\generator\node_modules\nunjucks\src\environment.js:631:18), <anonymous>:10:11)
    at PATH@asyncapi\generator\node_modules\nunjucks\src\environment.js:611:9
    at eval (eval at _compile (PATH@asyncapi\generator\node_modules\nunjucks\src\environment.js:631:18), <anonymous>:54:12)
    at PATH@asyncapi\generator\node_modules\nunjucks\src\environment.js:611:9
    at eval (eval at _compile (PATH@asyncapi\generator\node_modules\nunjucks\src\environment.js:631:18), <anonymous>:10:11)
[...]

PR #327, opened 2020-05-03
feat: add debug flag to enable Nunjucks dev flag for more filter errors

derberg:
**Description**

- add debug flag to enable Nunjucks dev flag for more filter errors

**Related issue(s)**
See also #326. This PR is not solving linked issues yet as there is a pending discussion in Nunjucks.

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,6 +46,7 @@ Usage: cli [options] <asyncapi> <template>
 Options:
   -V, --version                  output the version number
   -d, --disable-hook <hookName>  disable a specific hook
+  --debug                        enable more specific errors about filters. Keep in mind that as a result errors about template are less descriptive
   -i, --install                  installs the template and its dependencies (defaults to false)
   -n, --no-overwrite <glob>      glob or path of the file(s) to skip when regenerating
   -o, --output <outputDir>       directory where to put the generated files (defaults to current directory)
--- a/README.md
+++ b/README.md
@@ -46,7 +46,7 @@ Usage: cli [options] <asyncapi> <template>
 Options:
   -V, --version                  output the version number
   -d, --disable-hook <hookName>  disable a specific hook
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -46,6 +46,7 @@ Usage: cli [options] <asyncapi> <template>
 Options:
   -V, --version                  output the version number
   -d, --disable-hook <hookName>  disable a specific hook
+  --debug                        enable more specific errors about filters. Keep in mind that as a result errors about template are less descriptive
```
fmvilas, line 49: I like that we have a `--debug` option but it shouldn't be **only** about filters. We can either remove the "specificity" on the description or call it `--debug-filters`. I'm more inclined to the first option and create an issue to follow up on improving debugging. Probably using https://www.npmjs.com/package/debug.

derberg: How about `enable more specific errors in the console. At the moment it only shows specific errors about filters. Keep in mind that as a result errors about template are less descriptive`

I'm with option to leave `--debug` and then extend it with more use cases.

fmvilas: Sounds good for now. What worries me is this "for now". We soon should be adding error handling to the generator because logging an error in the console is fine when you use it as a CLI but it's not great when you use it as a library. I'd just remove any mention to filters here and would probably put this information somewhere else but I'm also concerned that people will not see it if it's not here. It's just weird to me that a "debug" option makes explicit mention to one of the features.

derberg: the description reflects reality, and I think we need it there so there are no other expectations. Once we add more debugging options we should simply adjust the description to make it more generic, right? It is strange that it is for filters only but this is because it is just a first iteration

fmvilas: Ok. Go for it! 👍

derberg: @fmvilas description updated

PR #334, opened 2020-05-12
feat: add support for configurable hooks libraries

### LINKED ISSUE #318: Avoid hooks repetition similar to what we did with filters
#### Reason/Context
We duplicate code across templates and we should have a central library with reusable hooks

#### Description
Check out how we did it with filters and we probably should do the same to hooks https://github.com/asyncapi/generator/issues/212

derberg:
**Description**

- `generate:before` and others become a hook type. Hook name is the name of hook function. you do not have to always provide it but it is a must have for hooks libararies and you enable a hook by referring it's name
- Way of writing hook function changed. `register` function is no longer needed as part of the hook which enables easier testing. You write a regular functon and expose it grouped by hook type:
```
module.exports = {
  "generate:after": [myHook1, myHook2],
  "generate:before": myHook3
}
```
- hooks registration extracted to separate `js` file and local hooks traversing refactored to use `xfs` for async walking as we use in other parts of the generator
- test and docs
- I tried to have more sophisticated tests for hooks registration where I can test different scenarios when passing different configs with errors or not but I think I'm too stupid to do it with jest. I wasted a few hours trying to mock files with modules that could be used in the registration but first I stumbled upon an issue with `xfs` that didn't work well with `mock-fs`. Then with quick refactor to stop using `xfs` in favor of `klaw` things were getting better but this would require larger refactor to kind out `xfs` from the project. 

**Related issue(s)**
See also #318


README.md
```diff
@@ -45,7 +45,7 @@ Usage: cli [options] <asyncapi> <template>
 
 Options:
   -V, --version                  output the version number
-  -d, --disable-hook <hookName>  disable a specific hook
+  -d, --disable-hook <hookName>  disable a specific hook type
```
Tenischev, line 48: `<hookType>` or `<hookTypeName>` instead of `<hookName>`?

cli.js
```diff
@@ -53,7 +53,7 @@ program
     asyncapiFile = path.resolve(asyncAPIPath);
     template = tmpl;
   })
-  .option('-d, --disable-hook <hookName>', 'disable a specific hook', disableHooksParser)
+  .option('-d, --disable-hook <hookName>', 'disable a specific hook type', disableHooksParser)
```
Tenischev, line 56: `<hookType>` or `<hookTypeName>` instead of `<hookName>`?

docs/authoring.md
```diff
@@ -127,17 +140,17 @@ module.exports = register => {
...
+  };
 };
 ```
 
 Example hook for changing the filename of a template file. Replaces all '-' characters with '_'.
 ```js
-module.exports = register => {
-	register('setFileTemplateName ', (generator, hookArguments) => {
+module.exports = {
+	'setFileTemplateName': generator, hookArguments) => {
```
Tenischev, line 150: `(` missed in `generator, hookArguments)`

docs/authoring.md
```diff
@@ -209,7 +223,10 @@ The `.tp-config.json` file contains a JSON object that may have the following in
   "generator": "<2.0.0",
   "filters": [
     "@asyncapi/generator-filters"
-  ]
+  ],
+  "hooks": {
+    "@asyncapi/generator-hooks": "hookFunctionName"
```
Tenischev, line 228: Will it be better to provide upper in example section at least one hook with defined name?

docs/authoring.md
```diff
@@ -80,22 +80,35 @@ In case you have more than one template and want to reuse filters, you can put t
...
+- All modules listed in the template configuration and triggers only hooks that names were added to the config. You can use the official AsyncAPI [hooks library](https://github.com/asyncapi/generator-hooks). To learn how to add hooks to configuration [read more about the configuration file](#configuration-file).
+
 #### Examples
+
+> Some examples have name of functions provided and some not. Keep in mind that hook functions kept in template in default location do not require a name.
+
+Most basic module with hooks looke like this:
+```js
+module.exports = {
+  'generate:after': generator => {console.log(generator);}
```
fmvilas, line 100:
```suggestion
  'generate:after': generator => console.log('This runs after generation is complete')
```

README.md
derberg: will do `<hookType>`

docs/authoring.md
derberg: What do you mean? in example section we provide examples of hooks that you rather put locally in your template, writing own hooks library is like advanced usage in my opinion

this config example shows usage of our official hooks library, you do not have to add to config your local hooks

Tenischev: Yes, but assume i'm newbie. I just read Hooks section, now read this section and see `specify exact name of the hook`. And now i ask, what exactly `hook name` is? This has not been mentioned before.
So, sure, you do not need to give names to local hooks, but example of hook with name will increase understanding of the concept, from my point of view.

derberg: how about now? I added some more content https://github.com/asyncapi/generator/pull/334/commits/87a9dfd4f170a5ac8927c01f0e789f9fc13bdde9

Tenischev: I still think that one example is better than long description 😄 , but nevertheless questions are described now in docs. So it's fine.

PR #337, opened 2020-05-13
fix: watcher ignores output and regeneration in loop

### LINKED ISSUE #185: Generator with watcher keeps regenerating if AsyncAPI file gets overwritten but contains the same data
Given you generate the `nodejs` template in the same directory where the AsyncAPI file is located the watcher will keep regenerate in an endless loop.

We could save the hash of the file/folder to see if any changes occurred before regenerating.

derberg:
**Description**

- watcher now watches explicitly the location of the template specified by the user. Before it was watching location where the template was installed in the generator global location
```
/.../.nvm/versions/node/v12.16.1/lib/node_modules/@asyncapi/generator/node_modules/@asyncapi/nodejs-ws-template/output/src/api/index.js was changed
```
- watcher now by default ignores output directory
- watcher supports flow when you trigger generation with watcher in the root of the template
- bump chokidar to latest

**Related issue(s)**
Fixes #185

### DIFF
--- a/cli.js
+++ b/cli.js
@@ -60,7 +60,7 @@ program
   .option('-o, --output <outputDir>', 'directory where to put the generated files (defaults to current directory)', parseOutput, process.cwd())
   .option('-p, --param <name=value>', 'additional param to pass to templates', paramParser)
   .option('--force-write', 'force writing of the generated files to given directory even if it is a git repo with unstaged files or not empty dir (defaults to false)')
-  .option('--watch-template', 'watches the template directory and the AsyncAPI document, and re-generate the files when changes occur')
+  .option('--watch-template', 'watches the template directory and the AsyncAPI document, and re-generate the files when changes occur. Ignores the output directory')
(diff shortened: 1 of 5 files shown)


cli.js
```diff
@@ -78,14 +78,19 @@ xfs.mkdirp(program.output, async err => {
 
   // If we want to watch for changes do that
   if (program.watchTemplate) {
-    const watchDir = path.resolve(Generator.DEFAULT_TEMPLATES_DIR, template);
+    const watchDir = path.resolve(template);
+    // Template name is needed as it is not always a part of the cli commad
+    // There is a use case that you run generator from a root of the template with `./` path
+    const templateName = require(path.resolve(watchDir,'./package.json')).name;
```
fmvilas, line 84: Not sure if the slash there will create problems on Windows. Just in case, you can do this:

```suggestion
    const templateName = require(path.resolve(watchDir,'package.json')).name;
```

PR #341, opened 2020-05-14
fix: watcher not reloading changes done to local filters and hooks

### LINKED ISSUE #204: Watcher does not reload filters when they are changed
When using the watcher and you change the filter it does recognize the change, however they are never applied.

derberg:
**Description**

- clear module cache each time hooks and filters are required

**Related issue(s)**
Resolves #204


derberg:
@jonaslagoni as you reported the issue, would be cool if you could check it out. I tried on local and when I add console log to hook or filter during watching, the changes are picked up

jonaslagoni:
> @jonaslagoni as you reported the issue, would be cool if you could check it out. I tried on local and when I add console log to hook or filter during watching, the changes are picked up

I'll review it later tonight if I have the time 😄

derberg:
@jonaslagoni I'm planning to sleep tonight :)

PR #343, opened 2020-05-15
feat: add default value for template parameters

### LINKED ISSUE #339: Enable template developer to declare default value for a parameter in the configuration file
#### Reason/Context

Discussion started here https://github.com/asyncapi/generator-hooks/pull/1#discussion_r424954487 when we were about to merge new hooks handling and ability to reuse hooks across templates and use parameters inside hooks.

Why `parameters[param].default` in config would be useful?
- much clearer for the template user do see directly what are the default values for parameters
- template developers, when they use some generic hook and its parameter, could specify a default value for the parameter that suits the purpose of their template

#### Description

Add support for `parameters[param].default` in the `.tp-config.json` file

Tenischev:
**Description**

- Now you as template author could define a default value for template properties. If user pass parameter value in start command, for sure, it will be used as value of parameter. 
- Tested with https://github.com/asyncapi/java-spring-template/pull/37

**Related issue(s)**
Resolves #339

### DIFF
--- a/docs/authoring.md
+++ b/docs/authoring.md
@@ -183,7 +183,7 @@ The `.tp-config.json` file contains a JSON object that may have the following in
 |`supportedProtocols`| [String] | A list with all the protocols this template supports.
 |`parameters`| Object[String, Object] | An object with all the parameters that can be passed when generating the template. When using the command line, it's done by indicating `--param name=value` or `-p name=value`.
 |`parameters[param].description`| String | A user-friendly description about the parameter.
-|`parameters[param].default`| String | Default value of this parameter if not set in the start command.
+|`parameters[param].default`| Any | Default value of the parameter if not specified.
 |`parameters[param].required`| Boolean | Whether the parameter is required or not.
 |`conditionalFiles`| Object[String, Object] | An object containing all the file paths that should be conditionally rendered. Each key represents a file path and each value must be an object with the keys `subject` and `validation`.
 |`conditionalFiles[filePath].subject`| String | The `subject` is a [JMESPath](http://jmespath.org/) query to grab the value you want to apply the condition to. It queries an object with the whole AsyncAPI document and, when specified, the given server. The object looks like this: `{ asyncapi: { ... }, server: { ... } }`.
@@ -201,8 +201,12 @@ The `.tp-config.json` file contains a JSON object that may have the following in
   "parameters": {
     "server": {
       "description": "The server you want to use in the code.",
-      "default": "localhost",
       "required": true
+    },
+    "pathToApiFile": {
+      "description": "The path to store original AsyncAPI file.",
+      "default": "./api",
+      "required": false
     }
   },
   "conditionalFiles": {
--- a/docs/authoring.md
+++ b/docs/authoring.md
@@ -203,9 +203,9 @@ The `.tp-config.json` file contains a JSON object that may have the following in
       "description": "The server you want to use in the code.",
       "required": true
     },
-    "asyncapiFileDir": {
-      "description": "The path to store original AsyncAPI file.",
-      "default": "./api",
+    "dummyParameter": {
+      "description": "Example of parameter with default value.",
+      "default": "just a string",
       "required": false
     }
   },
@@ -244,4 +244,3 @@ There are some template parameters that have a special meaning:
 |Name|Description|
 |---|---|
 |`server`| It is used to let the template know which server you want to use. In some cases, this may be required. For instance, when generating code that connects to a specific server. If your template need your users to specify a server, use this parameter.
-|`asyncapiFileDir`| It is used by `@asyncapi/generator-hooks#`[`createAsyncapiFile`](https://github.com/asyncapi/generator-hooks/blob/62c50cac6156110e1325152cb2599d916a310055/src/index.js#L5) to specify directory where original AsyncAPI file will be stored. Defining a default parameter value allows template maintainers to pre-set this location. |
--- a/docs/authoring.md
+++ b/docs/authoring.md
@@ -183,6 +183,7 @@ The `.tp-config.json` file contains a JSON object that may have the following in
 |`supportedProtocols`| [String] | A list with all the protocols this template supports.
 |`parameters`| Object[String, Object] | An object with all the parameters that can be passed when generating the template. When using the command line, it's done by indicating `--param name=value` or `-p name=value`.
 |`parameters[param].description`| String | A user-friendly description about the parameter.
+|`parameters[param].default`| String | Default value of this parameter if not set in the start command.
 |`parameters[param].required`| Boolean | Whether the parameter is required or not.
 |`conditionalFiles`| Object[String, Object] | An object containing all the file paths that should be conditionally rendered. Each key represents a file path and each value must be an object with the keys `subject` and `validation`.
 |`conditionalFiles[filePath].subject`| String | The `subject` is a [JMESPath](http://jmespath.org/) query to grab the value you want to apply the condition to. It queries an object with the whole AsyncAPI document and, when specified, the given server. The object looks like this: `{ asyncapi: { ... }, server: { ... } }`.
@@ -200,6 +201,7 @@ The `.tp-config.json` file contains a JSON object that may have the following in
   "parameters": {
     "server": {
       "description": "The server you want to use in the code.",
+      "default": "localhost",
       "required": true
     }
   },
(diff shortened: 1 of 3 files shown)


docs/authoring.md
```diff
@@ -183,6 +183,7 @@ The `.tp-config.json` file contains a JSON object that may have the following in
 |`supportedProtocols`| [String] | A list with all the protocols this template supports.
 |`parameters`| Object[String, Object] | An object with all the parameters that can be passed when generating the template. When using the command line, it's done by indicating `--param name=value` or `-p name=value`.
 |`parameters[param].description`| String | A user-friendly description about the parameter.
+|`parameters[param].default`| String | Default value of this parameter if not set in the start command.
```
Tenischev, line 186: I'm not sure about type `String`. Actually `String`, `Integer`, `Boolean` will work. But from CLI all parameter values are `String`, so i set here only `String`

fmvilas, line 186: What is "start command"? Also, from CLI everything may be String but when using generator as a module it can be anything. I'd put Any as the type.

lib/generator.js
```diff
@@ -642,9 +642,36 @@ class Generator {
...
 
+  /**
+   * Loads default values of parameters from template config. If value was already set as parameter it will not be
+   * overrided.
+   * @private
+   */
+  async loadDefaultValues() {
+    const parameters = this.templateConfig.parameters;
+    const defaultValues = [];
+    Object.keys(parameters || {}).forEach(key => {
```
fmvilas, line 657: We are already iterating over the parameters keys here: https://github.com/asyncapi/generator/blob/master/lib/generator.js#L664.

Can we reuse this loop instead of iterating again?

test/generator.test.js
```diff
@@ -383,4 +383,59 @@ describe('Generator', () => {
...
   });
+
+  describe('#loadDefaultValues', () => {
+    it('default value of parameter is set', async () => {
+      const gen = new Generator('testTemplate', __dirname, {
+        templateParams: {
+          test: true
+        }
+      });
+      const json = '{"parameters": {' +
```
derberg, line 394: JS supports template literals and I recommend it here instead of concatenating strings like this

derberg: anyway why string?

lib/generator.js
```diff
@@ -642,9 +642,36 @@ class Generator {
     } catch (e) {
       this.templateConfig = {};
     }
+    await this.loadDefaultValues();
```
derberg, line 645: please check some other tests where we check if `validateTemplateConfig` was called, and make sure they are extended to check if `loadDefaultValues` is called

docs/authoring.md
Tenischev: I agree that type is `Any`, didn't know how to properly name it, thanks!
"Start command" is `ag asyncapi.yaml @asyncapi/html-template`, how to name it to be more clear?

lib/generator.js
Tenischev: It's in `validateTemplateConfig` method, different purposes. Btw i do not want to complicate the `validateTemplateConfig` method even more.

lib/generator.js
```diff
@@ -642,9 +642,36 @@ class Generator {
...
+  async loadDefaultValues() {
+    const parameters = this.templateConfig.parameters;
+    const defaultValues = [];
+    Object.keys(parameters || {}).forEach(key => {
+      if (parameters[key].default) {
+        defaultValues.push(key);
+      }
+    });
+
+    if (Array.isArray(defaultValues)) {
```
derberg, line 663: because of `const defaultValues = [];` it is always array and always `true` here

docs/authoring.md
derberg: the thing is that as @fmvilas mentioned, it is not only used in CLI but generator can be used as a library too. Just call it `Default value of the parameter if not specified.` maybe?

lib/generator.js
Tenischev: Ok, will fix, but who say this to author of these lines https://github.com/asyncapi/generator/blob/230e6652e48f0215eb6834b53fc4eb86c6819677/lib/generator.js#L663-L668 ? 😃

lib/generator.js
Tenischev: Everywhere where you have check for call of `validateTemplateConfig` the `loadTemplateConfig` is mocked. So in previous tests `loadDefaultValues` method is unreachable.

derberg: 🤦 you're right

docs/authoring.md
Tenischev: Agree, used.

lib/generator.js
Tenischev: code updated

docs/authoring.md
```diff
@@ -200,6 +201,7 @@ The `.tp-config.json` file contains a JSON object that may have the following in
   "parameters": {
     "server": {
       "description": "The server you want to use in the code.",
+      "default": "localhost",
```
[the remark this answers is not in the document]
Tenischev: another parameter added

test/generator.test.js
Tenischev: done

lib/generator.js
Tenischev: fixed

docs/authoring.md
```diff
@@ -201,6 +202,11 @@ The `.tp-config.json` file contains a JSON object that may have the following in
     "server": {
       "description": "The server you want to use in the code.",
       "required": true
+    },
+    "pathToApiFile": {
```
derberg, line 206: please add some more explanation to it:
- that when you access this param in your template with {{params.pathToApiFile}} you get the default value if other not provided
- that it will never work with params that are marked as `required` as those you always have to provide explicitly

It is not that obvious when looking at the example

derberg: you can just extend the description in the table where default param is described and it will be fine

derberg:
@Tenischev I've just added one last comment to docs, the rest looks awesome. I also played with it locally and lgtm

do you plan to extend this PR https://github.com/asyncapi/java-spring-template/pull/37 to use https://github.com/asyncapi/generator-hooks/blob/master/src/index.js ? as not it should be possible

docs/authoring.md
Tenischev: Seems I was a little dizzy when wrote this 😃 
Everything make sense, will fix

docs/authoring.md
```diff
@@ -244,3 +244,4 @@ There are some template parameters that have a special meaning:
 |Name|Description|
 |---|---|
 |`server`| It is used to let the template know which server you want to use. In some cases, this may be required. For instance, when generating code that connects to a specific server. If your template need your users to specify a server, use this parameter.
+|`asyncapiFileDir`| It is used by `@asyncapi/generator-hooks#`[`createAsyncapiFile`](https://github.com/asyncapi/generator-hooks/blob/62c50cac6156110e1325152cb2599d916a310055/src/index.js#L5) to specify directory where original AsyncAPI file will be stored. Defining a default parameter value allows template maintainers to pre-set this location. |
```
derberg, line 247: @Tenischev you went too far with docs here, `asyncapiFileDir` is not a special parameter, it is not part of generator code, it is in generic hooks but not mandatory and if someone what's his own hook that supports parameter with the same name, this is all fine

Tenischev: I would say, it depends on how you consider this sentence 
> There are some template parameters that have a special meaning

Because this is only one sentence that describe these parameters, no information that "it must be a part of generator code" or "each template must mandatory have it" e.t.c.

From my point of view, the phrase "It is used by `@asyncapi/generator-hooks#createAsyncapiFile`" clearly explain by whom used this parameter.

Of course, i could move this explanation to the description `asyncapiFileDir.description`, but then it will not be a good `example` of `.tp-config.json` file, from my point of view

derberg: That is a philosophical dispute 😄 


the example will be just fine if you call it `dummyParameter`, it doesn't have to be a super real param example. `asyncapiFileDir` is simply not special at all, it is just a random parameter anyone can define without using generator-hooks library, while `server` is super special one https://github.com/asyncapi/generator/blob/master/lib/generator.js#L685

Tenischev: removed

Tenischev:
Also updated https://github.com/asyncapi/java-spring-template/pull/37

lib/generator.js
fmvilas: Awesome! Thanks for taking care!

fmvilas:
Agree. This feature is really cool 👏

PR #346, opened 2020-05-15
feat: fetching asyncapi file from server in cli and library

### LINKED ISSUE #266: Support remote AsyncAPI files
After adding support for remote templates (#261), it makes total sense now to support remote AsyncAPI files. This should be useful especially on CI systems.

derberg:
**Description**

- new function exposed -> `generateFromURL`
- support in CLI so we can do `ag https://raw.githubusercontent.com/asyncapi/asyncapi/master/examples/2.0.0/streetlights.yml @asyncapi/html-template`

validation if path is filesystem path or url is very basic here, not based on any sophisticated regex but just using `url.parse`. Let me know if you think it is bad approach. I don't see a use case that someone uses this feature as a library and on a server passes relative path to fetch the file from the same server where the generator is running, I think it is always about fetching from different server (hostname)

**Related issue(s)**
Resolves #266

### DIFF
--- a/cli.js
+++ b/cli.js
@@ -64,7 +64,7 @@ program
   .parse(process.argv);
 
 if (!asyncapiDocPath) {
-  console.error(red('> Path or Link to AsyncAPI file not provided.'));
+  console.error(red('> Path or URL to AsyncAPI file not provided.'));
   program.help(); // This exits the process
 }
 const isAsyncapiDocLocal = isFilePath(asyncapiDocPath);
--- a/cli.js
+++ b/cli.js
@@ -7,14 +7,14 @@ const xfs = require('fs.extra');
 const packageInfo = require('./package.json');
 const Generator = require('./lib/generator');
 const Watcher = require('./lib/watcher');
-const { isLocalTemplate } = require('./lib/utils');
+const { isLocalTemplate, isFilePath } = require('./lib/utils');
 
 const red = text => `\x1b[31m${text}\x1b[0m`;
 const magenta = text => `\x1b[35m${text}\x1b[0m`;
 const yellow = text => `\x1b[33m${text}\x1b[0m`;
 const green = text => `\x1b[32m${text}\x1b[0m`;
 
-let asyncapiFile;
+let asyncapiDocPath;
 let template;
 const params = {};
 const noOverwriteGlobs = [];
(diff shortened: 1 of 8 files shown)


cli.js
```diff
@@ -63,10 +63,11 @@ program
   .option('--watch-template', 'watches the template directory and the AsyncAPI document, and re-generate the files when changes occur. Ignores the output directory. This flag should be used only for template development.')
   .parse(process.argv);
 
-if (!asyncapiFile) {
-  console.error(red('> Path to AsyncAPI file not provided.'));
+if (!asyncapiDocPath) {
+  console.error(red('> Path or Link to AsyncAPI file not provided.'));
```
fmvilas, line 67:
```suggestion
  console.error(red('> Path or URL to AsyncAPI file not provided.'));
```

cli.js
```diff
@@ -78,19 +79,25 @@ xfs.mkdirp(program.output, async err => {
...
     // Template name is needed as it is not always a part of the cli commad
     // There is a use case that you run generator from a root of the template with `./` path
     const templateName = require(path.resolve(watchDir,'package.json')).name;
 
-    console.log(`[WATCHER] Watching for changes in the template directory ${magenta(watchDir)} and in the AsyncAPI file ${magenta(asyncapiFile)}`);
-
+    if (isAsyncapiDocLocal) {
+      console.log(`[WATCHER] Watching for changes in the template directory ${magenta(watchDir)} and in the AsyncAPI file ${magenta(asyncapiDocPath)}`);
+      watcher = new Watcher([asyncapiDocPath, watchDir], outputPath);
+    } else {
```
fmvilas, line 92: This just reminded me this "else" should always be the case. The param is called `watchTemplate` and we're watching the AsyncAPI file too. We should have a different param for watching the AsyncAPI file, like `--watch` (thinking from a user perspective, watching the template is not interesting except for template developers). Should we create a follow-up issue?

derberg: I'm not sure. With my previous bug fixes I explicitly specified that `--watch-template` is for development only and I was thinking if there might be a use case that someone uses a template with a watch enabled while editing asyncapi file....and I could not imagine someone doing it really, that is what AsyncAPI Hub is for 😄

derberg: I'll create an issue for it for further discussion, I don't think generator is a tool that should solve a need for nice AsyncAPI document development experience, this is something for AsyncAPI CLI that we do not have yet :D

derberg: https://github.com/asyncapi/generator/issues/352

PR #349, opened 2020-05-20
feat: move template config to package.json

### LINKED ISSUE #228: Get rid of .tp-config.json and use package.json
Now that [templates are packages](https://github.com/asyncapi/generator/pull/219), we could use a custom key on the package.json file instead of having our own entire file. For instance:

```js
// package.json
{
  "name": "my-template",
  "version": "0.1.0",
  "asyncapi-generator": {
    "supportedProtocols": ["amqp", "mqtt", "kafka", "ws"],
    "parameters": {
      "server": {
        "description": "The server you want to use in the code.",
        "required": true
      }
    },
    "nonRenderableFiles": [
      "src/api/middlewares/*.*",
      "src/lib/config.js"
    ]
  },
  "dependencies": {
    // ...
  }
}
```

Thoughts?

derberg:
**Description**

- `.tp-config.json` is removed and config should be now done under `generator` props in the `package.json` file
- linter didn't like `validateTemplateConfig` as too complex function. Because I was working on config here so I took the liberty to refactor it and move to a separate file + add tests that cover every validation option

**MIGRATION**
This is a breaking change. Migrate by removing current config file and moving its content to `package.json` and bumping version of supported generator to the one released with this PR.
Example -> https://github.com/asyncapi/nodejs-template/pull/15

**Related issue(s)**
Resolves #228

### DIFF
--- a/lib/generator.js
+++ b/lib/generator.js
@@ -8,12 +8,10 @@ const ramlDtParser = require('@asyncapi/raml-dt-schema-parser');
 const openapiSchemaParser = require('@asyncapi/openapi-schema-parser');
 const Nunjucks = require('nunjucks');
 const jmespath = require('jmespath');
-const Ajv = require('ajv');
 const filenamify = require('filenamify');
 const git = require('simple-git/promise');
 const npmi = require('npmi');
-const semver = require('semver');
-const packageJson = require('../package.json');
+const { validateTemplateConfig } = require('./templateConfigValidator');
 const {
   convertMapToObject,
   isFileSystemPath,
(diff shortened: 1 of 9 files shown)


lib/generator.js
```diff
@@ -166,9 +162,9 @@ class Generator {
     this.templateContentDir = path.resolve(this.templateDir, TEMPLATE_CONTENT_DIRNAME);
     this.configNunjucks();
     await this.loadTemplateConfig();
+    validateTemplateConfig(this.templateConfig, this.templateParams, asyncapiDocument);
```
derberg, line 165: I changed the order of calling validation and removed it from here https://github.com/asyncapi/generator/pull/349/files#diff-b58e2f1cee2acff55355343774e6d576L646 as I could not come up with an explanation why we were calling validation twice.

also, it is no longer async as I think it should never be

PR #355, opened 2020-05-21
docs: update readme with toc and list of templates

### LINKED ISSUE #285: Requesting README Updates
The getting started experience could be made a bit easier with some relatively simple updates to the README.md. 

I suggest the following: 
* A link for how to install npm (useful for non-node devs)
* Should have a list of available templates since they are now in other repos
* The example command assume that you know _markdown_ is a template, this should be clarified
* Usage should say `AsyncAPI Document` or `AsyncAPI Specification File` instead of just `asyncapi`. It should also specify if it needs to be yaml, json or either
* While not really a README thing, it would be ideal if the generator would print out the parameters that were passed in so it's clear that the generator read them in properly from the command line.

derberg:
**Description**

- add list of templates
- add toc
- add npx instruction
- add overview
- add more clarification on cli usage
- add contributors
- custom GH action that informs us that list of templates is outdated
- workflow that makes sure that bot updates API and TOC docs when there is a push of just `docs:` commit

TODO: if we go with list of templates, we need it also in tools on asyncapi website too

**Related issue(s)**
Resolves #285

### DIFF
--- a/.github/templates-list-validator/index.js
+++ b/.github/templates-list-validator/index.js
@@ -9,7 +9,7 @@ async function run() {
     const templatesListContent = getReadmeContent();
     const officialTemplates = await searchOfficialTemplates(token);
 
-    const missingTemplates = officialTemplates.filter(str => templatesListContent.indexOf(str) === -1);
+    const missingTemplates = officialTemplates.filter(str => !templatesListContent.includes(str));
 
     if (missingTemplates) core.setFailed(
         `The following templates are not in the README.md: ${missingTemplates}. Make sure missing templates are added to the list in the README of this repository and also to the website to the list of offecial tools: https://github.com/asyncapi/website/blob/master/content/docs/tooling.md`
--- a/.github/templates-list-validator/index.js
+++ b/.github/templates-list-validator/index.js
@@ -1,0 +1,71 @@
+const github = require('@actions/github');
+const core = require('@actions/core');
+const { readFileSync } = require('fs');
+
+async function run() {
+  try {
+    const token = process.env.GITHUB_TOKEN || core.getInput('token', { required: true });
+    //TODO we should not only check the content of the readme but also the content of the website listing from https://github.com/asyncapi/website/blob/master/content/docs/tooling.md
+    const templatesListContent = getReadmeContent();
+    const officialTemplates = await searchOfficialTemplates(token);
+
+    const missingTemplates = officialTemplates.filter(str => templatesListContent.indexOf(str) === -1);
+
+    if (missingTemplates) core.setFailed(
+        `The following templates are not in the README.md: ${missingTemplates}. Make sure missing templates are added to the list in the README of this repository and also to the website to the list of offecial tools: https://github.com/asyncapi/website/blob/master/content/docs/tooling.md`
+    );
+
+  } catch (error) {
+    core.setFailed(error.message);
+  }
+}
+
+/**
+   * Gets the markdown table from the readme of the Generator
+   *
+   * @private
+   * @return {String}
+   */
+function getReadmeContent() {
+  const readmeContent = readFileSync('../../README.md', 'utf8');
+  const startingTag = '<!-- TEMPLATES-LIST:';
+  const closingTag = '-->';
+  const startOfOpeningTagIndex = readmeContent.indexOf(`${startingTag}START`);
+  const endOfOpeningTagIndex = readmeContent.indexOf(closingTag, startOfOpeningTagIndex);
+  const endOfClosingTagIndex = readmeContent.indexOf(`${startingTag}END ${closingTag}`);
+
+  if (endOfOpeningTagIndex === -1 || endOfClosingTagIndex === -1) {
+    core.setFailed();
+    throw new Error('Templates list is missing or someone removed markers that indicate the list');
+  }
+
+  return readmeContent.slice(endOfOpeningTagIndex + closingTag.length, endOfClosingTagIndex);
+}
+
+/**
+   * Gets array with list of names of all the official templates
+   *
+   * @private
+   * @param {String} Token to authorize with GitHub
+   * @return {Array[String]}
+   */
+async function searchOfficialTemplates(token) {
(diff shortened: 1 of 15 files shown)


derberg:
@jonaslagoni you might like updates coming with this PR -> https://github.com/asyncapi/generator/pull/355/files#diff-4ef181fda0284a9e7f5db8fbad0d21b8

.github/templates-list-validator/index.js
```diff
@@ -0,0 +1,71 @@
...
+  const readmeContent = readFileSync('../../README.md', 'utf8');
+  const startingTag = '<!-- TEMPLATES-LIST:';
+  const closingTag = '-->';
+  const startOfOpeningTagIndex = readmeContent.indexOf(`${startingTag}START`);
+  const endOfOpeningTagIndex = readmeContent.indexOf(closingTag, startOfOpeningTagIndex);
+  const endOfClosingTagIndex = readmeContent.indexOf(`${startingTag}END ${closingTag}`);
+
+  if (endOfOpeningTagIndex === -1 || endOfClosingTagIndex === -1) {
+    core.setFailed();
+    throw new Error('Templates list is missing or someone removed markers that indicate the list');
```
fmvilas, line 39: I'd give a more thorough explanation of this error. I'm imagining us in 2 years from now seeing this message and wondering what the fuck is happening 😄

.github/templates-list-validator/index.js
```diff
@@ -0,0 +1,71 @@
...
+const { readFileSync } = require('fs');
+
+async function run() {
+  try {
+    const token = process.env.GITHUB_TOKEN || core.getInput('token', { required: true });
+    //TODO we should not only check the content of the readme but also the content of the website listing from https://github.com/asyncapi/website/blob/master/content/docs/tooling.md
+    const templatesListContent = getReadmeContent();
+    const officialTemplates = await searchOfficialTemplates(token);
+
+    const missingTemplates = officialTemplates.filter(str => templatesListContent.indexOf(str) === -1);
```
fmvilas, line 12: The following suggestions are much more readable in my opinion:

```suggestion
    const missingTemplates = officialTemplates.filter(str => templatesListContent.includes(str) === false);
```

or

```suggestion
    const missingTemplates = officialTemplates.filter(str => !templatesListContent.includes(str));
```

But these are just tastes :)

derberg: I like the 2nd one, it is clean, didn't know -1 evaluates to false, makes sense

fmvilas: It doesn't evaluate to false. I changed the method to `includes`, which returns a boolean.

.github/templates-list-validator/index.js
derberg: done

derberg:
@fmvilas yeap about the tools on website, I will just leave this TODO here https://github.com/asyncapi/generator/pull/355/files#diff-62076ba232f6d377838d5936746b8194R8 and you remember that on a site, where you will list tools, also explicitly list the templates ;)

and, I would forget, when you will work on a new website, please please give more love to the way how "edit document" functionality is exposed, some visible button top right and not so lame as I did it as a text at the bottom of the doc ;)

.github/templates-list-validator/index.js
derberg: 🤦

fmvilas:
Ha! Ok, I'll try to make it more prominent.

Mrc0113:
@derberg looks much better! I learned something new with `npx` too :) 

I suggest adding two more things: 
1) How to update the generator to the latest since changes are made quite often
2) How to run the generator against a specific template version

derberg:
@Mrc0113 good points, versioning is not that clear

derberg:
@Mrc0113 what do you think https://github.com/asyncapi/generator/pull/355/commits/53b42ca2f43101400d9dcf76d74daa0af403c918 ?

@fmvilas I've added above after your approval, you might want to have a look to make sure you agree

README.md
```diff
@@ -16,33 +68,37 @@
...
-# Example that you can run inside generator directory after cloning this repository. First you specify mount in location of your AsyncAPI file and then you mount in directory where generation result should be saved.
-docker run --rm -it \
--v ${PWD}/test/docs/streetlights.yml:/app/asyncapi.yml \
--v ${PWD}/output:/app/output \
-asyncapi/generator -o ./output asyncapi.yml @asyncapi/html-template --force-write
-```
+You might want to update your local installation of generator for different reasons:
 
-## Usage
+* You want latest generator to have it's latest features. Perform usual installation and in case you had generator installed already, it will upgrade to latest available:
```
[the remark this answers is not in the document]
fmvilas: I'd also note here that depending on your npm installation, you might have to force installation as follows: `npm install -g --force @asyncapi/generator`. I think it recently happened to some users.

README.md
```diff
@@ -16,33 +68,37 @@
...
+    ```bash
+    npm install -g @asyncapi/generator
+    ```
+* You want specific version of the generator because your template might not be compatible with latest generator. Check [what version you need](https://github.com/asyncapi/generator/releases) and perform installation, specifying exact version with `@` character:
```
fmvilas, line 89:
```suggestion
* You want a specific version of the generator because your template might not be compatible with the latest generator. Check [what version you need](https://github.com/asyncapi/generator/releases) and perform installation, specifying the exact version with the `@` character:
```

README.md
```diff
@@ -16,33 +68,37 @@
...
+    ```bash
+    npm install -g @asyncapi/generator@0.50.0
+    ```
 
-### From the command-line interface (CLI)
+### Options
```
fmvilas, line 94:
```suggestion
### CLI usage
```

derberg:
@fmvilas I pushed your suggestions and also added a note about --force flag

derberg:
@Mrc0113 since I already added both topics you pointed out in the review and @fmvilas already approved, I'll merge this one and any future improvements can follow in separate issue or even directly PR

PR #359, opened 2020-05-23
fix: update cli Generator forceInstall parameter to install

sebastian-palma:
This is a very small change. As the cli instantiates a new `Generator` object with the given parameters, it uses `forceInstall` as the option to install the given template or not. But in the generator file the constructor definition it's done by deconstructing the third argument and the argument for installing the template is `forceInstall` not `install`.

So, no matter if you specify you want to install the template, `Generator.install` is always `false`.

I could've modified all the matches to `forceInstall` to make it clear you want to force the installation of the template, but I don't know if that's the case - as `forceWrite` is indeed forcing to write the changes. I better modified a single line to make the `forceInstall` option to be just `install`.

I noticed this "failed" silently as the cli file isn't tested, which makes me ask, would it be better to add full coverage to that file in order to merge this? Or, would it be easier to add a guard clause in the `Generator` constructor body to return if it receives an unexpected parameter (I think it'd be easier)?

By the way, thanks for this amazing tool.

### DIFF
--- a/lib/generator.js
+++ b/lib/generator.js
@@ -23,7 +23,8 @@ const {
   writeFile,
   copyFile,
   exists,
-  fetchSpec
+  fetchSpec,
+  getInvalidOptions
 } = require('./utils');
 const { registerFilters } = require('./filtersRegistry');
 const { registerHooks } = require('./hooksRegistry');
@@ -45,6 +46,7 @@ const PACKAGE_JSON_FILENAME = 'package.json';
 const ROOT_DIR = path.resolve(__dirname, '..');
 const DEFAULT_TEMPLATES_DIR = path.resolve(ROOT_DIR, 'node_modules');
 const TEMPLATE_CONTENT_DIRNAME = 'template';
+const GENERATOR_OPTIONS = ['debug', 'disabledHooks', 'entrypoint', 'forceWrite', 'install', 'noOverwriteGlobs', 'output', 'templateParams'];
 
 const shouldIgnoreFile = filePath =>
   filePath.startsWith(`.git${path.sep}`);
@@ -82,10 +84,12 @@ class Generator {
    * @param {Boolean} [options.debug=false] Enable more specific errors in the console. At the moment it only shows specific errors about filters. Keep in mind that as a result errors about template are less descriptive.
    */
   constructor(templateName, targetDir, { templateParams = {}, entrypoint, noOverwriteGlobs, disabledHooks, output = 'fs', forceWrite = false, install = false, debug = false } = {}) {
+    const invalidOptions = getInvalidOptions(GENERATOR_OPTIONS, arguments[arguments.length - 1] || []);
+    if (invalidOptions.length) throw new Error(`These options are not supported by the generator: ${invalidOptions.join(', ')}`);
     if (!templateName) throw new Error('No template name has been specified.');
     if (!entrypoint && !targetDir) throw new Error('No target directory has been specified.');
     if (!['fs', 'string'].includes(output)) throw new Error(`Invalid output type ${output}. Valid values are 'fs' and 'string'.`);
-    
+
     /** @type {String} Name of the template to generate. */
     this.templateName = templateName;
     /** @type {String} Path to the directory where the files will be generated. */
--- a/cli.js
+++ b/cli.js
@@ -142,7 +142,7 @@ function generate(targetDir) {
         noOverwriteGlobs,
         disabledHooks,
         forceWrite: program.forceWrite,
-        forceInstall: program.install,
+        install: program.install,
         debug: program.debug
       });
 
--- a/lib/__mocks__/utils.js
+++ b/lib/__mocks__/utils.js
@@ -1,4 +1,7 @@
 const utils = jest.genMockFromModule('../utils');
+const { getInvalidOptions } = require.requireActual('../utils');
+
+utils.getInvalidOptions = getInvalidOptions;
 
 utils.__files = {};
 utils.readFile = jest.fn(async (filePath) => {
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -119,3 +119,13 @@ utils.getGeneratorVersion = () => {
   return packageJson.version;
 };
 
+/**
+ * Filters out the Generator invalid options given
+ *
+ * @param {Array}
+ * @returns {Array}
+ */
+utils.getInvalidOptions = (generatorOptions, options) => {
+  if (typeof options !== 'object') return [];
+  return Object.keys(options).filter(param => !generatorOptions.includes(param));
+};
--- a/test/generator.test.js
+++ b/test/generator.test.js
@@ -53,6 +53,36 @@ describe('Generator', () => {
       expect(gen.templateParams.test).toStrictEqual(true);
     });
 
+    it('throws an error indicating an unexpected param was given', () => {
+      const t = () => new Generator('testTemplate', __dirname, {
+        entrypoint: 'test-entrypoint',
+        noOverwriteGlobs: ['test-globs'],
+        disabledHooks: ['test-hooks'],
+        output: 'string',
+        forceWrite: true,
+        forceInstall: true,
+        templateParams: {
+          test: true,
+        }
+      });
+      expect(t).toThrow('These options are not supported by the generator: forceInstall');
+    });
+
+    it('throws an error indicating multiple unexpected params were given', () => {
+      const t = () => new Generator('testTemplate', __dirname, {
+        entrypoint: 'test-entrypoint',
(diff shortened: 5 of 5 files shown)


derberg:
@sebastian-palma great catch! both solutions proposed by you are good but better is the one that validates incoming parameters, it's adding protection in case we make errors in cli.js and nice validation for those using the generator as a library. Would you be willing to add it to this PR or separate one?

sebastian-palma:
Cool @derberg, I can keep working in this PR.

sebastian-palma:
Hey @derberg, I just committed some changes. Would you mind helping me here to see if everything is okay?

lib/generator.js
```diff
@@ -53,6 +54,14 @@ const shouldIgnoreDir = dirPath =>
...
   || dirPath.startsWith(`.git${path.sep}`);
 
+const validateOptions = options => {
+  const unexpectedOptions = Object.keys(options).filter(param => !GENERATOR_OPTIONS.includes(param));
+  const readableUnexpectedOptions = unexpectedOptions.join('\', \'');
+  if (unexpectedOptions.length === 0) return '';
+  if (unexpectedOptions.length === 1) return `An unexpected param was given: '${readableUnexpectedOptions}'`;
+  return `Unexpected params were given: '${readableUnexpectedOptions}'`;
+};
+
```
derberg, line 64:
```suggestion
const validateOptions = options => {
  const unexpectedOptions = Object.keys(options).filter(param => !GENERATOR_OPTIONS.includes(param));
  if (!unexpectedOptions.length) return;
  const readableUnexpectedOptions = unexpectedOptions.join(', ');
  
  return `These options are not supported by the generator: ${readableUnexpectedOptions}`;
};
```

- when we console log other errors, we do not put them in between `'` this is why I suggest above to remove those
- 0 is always falsy so no need to do `=== 0`. You also do not have to return empty string as it is not needed anywhere
- are you sure we need 2 different log messages for plural and singular? in other places we just approach it like this `Error: This template requires the following missing params: server.` and I think that form DX perspective this is pretty fine, and as a result one line of code to maintain as then you can write it as I wrote it in my suggestion
- I proposed different error message, especially that for us `params` can be mistaken with params passed with `templateParams`. So at the end these are options

derberg: I think that the best would be if you move this function to utils, let us try to make everything possible to not grow generator.js to much 😅

lib/generator.js
```diff
@@ -82,10 +91,12 @@ class Generator {
    * @param {Boolean} [options.debug=false] Enable more specific errors in the console. At the moment it only shows specific errors about filters. Keep in mind that as a result errors about template are less descriptive.
    */
   constructor(templateName, targetDir, { templateParams = {}, entrypoint, noOverwriteGlobs, disabledHooks, output = 'fs', forceWrite = false, install = false, debug = false } = {}) {
+    const unexpectedOptionsError = validateOptions(arguments[arguments.length - 1] || []);
+    if (arguments.length === 3 && unexpectedOptionsError) throw new Error(unexpectedOptionsError);
```
derberg, line 95: damn I lost one comment, let we write again :( 

what do you think about splitting responsibility of `validateOptions` so it is only responsible to `getInvalidOptions` and the building of the error string stays in generator? then you can avoid things like `arguments.length === 3`

something like:
```
const invalidOptions = getInvalidOptions(arguments[arguments.length - 1] || [])
if (invalidOptions.length) throw new Error(`These options are not supported by the generator: ${invalidOptions.join(', ')}`);
```

I'm super curious what do you think about it

lib/generator.js
sebastian-palma: Thanks @derberg! Just added the `return ''` to return the same type of object in any case, but it's completely understandable, for all the other cases as well.

_"are you sure we need 2 different log messages for plural and singular?"_, not really, it was just to make it more fancy 😁.

I agree about moving out this added logic to utils, so I'm gonna give it a try.

lib/generator.js
sebastian-palma: Nice one, I thought the same.

As per the `arguments.length === 3`, it was there to avoid throwing a new error in case a third argument isn't given, as `Object.keys()` is receiving the last argument always and when a string is given it returns an array of strings corresponding to its length.

I'm gonna figure out how to return just the invalid options, which would make it easier 👌.

sebastian-palma:
Hey @derberg, just moved some things. Looking forward to another round of review. I'm wondering if it's okay with the guard clauses and your style guide.

lib/generator.js
```diff
@@ -82,10 +83,12 @@ class Generator {
    * @param {Boolean} [options.debug=false] Enable more specific errors in the console. At the moment it only shows specific errors about filters. Keep in mind that as a result errors about template are less descriptive.
    */
   constructor(templateName, targetDir, { templateParams = {}, entrypoint, noOverwriteGlobs, disabledHooks, output = 'fs', forceWrite = false, install = false, debug = false } = {}) {
+    const invalidOptionsError = invalidOptions(arguments[arguments.length - 1] || []);
+    if (invalidOptionsError) throw new Error(invalidOptionsError);
```
derberg, line 87: we kind of still have it here an error message hidden in the variable which is not the best as while reading this and next line I want to have a clear answer what error is thrown when

best experience would be to have
```
const invalidOptions = getInvalidOptions(GENERATOR_OPTIONS, arguments[arguments.length - 1])
if (getInvalidOptions.length) throw new Error(`These options are not supported by the generator: ${invalidOptions.join(', ')}`);
```

why:
- `GENERATOR_OPTIONS` belongs to a file where the generator constructor is, for maintenance reasons 
- error message reason mentioned at the beginning

then just merge `invalidOptions` with `getInvalidOptions` function that should not be marked as private.

---

Despite my comment above I still have a big concern about how we get `GENERATOR_OPTIONS` up to date, as if I forget to put new options there, no tests will fail. You noticed that in the initial implementation you forgot about `debug`

derberg: https://github.com/dsheiko/bycontract looks interesting but I haven't tried it, but it talks about jsdoc that we at least always keep up to date

sebastian-palma: Remove `invalidOptions`, really I don't know why both were marked as private, maybe too much time looking to the same piece of code 🙂

> Despite my comment above I still have a big concern about how we get `GENERATOR_OPTIONS` up to date, as if I forget to put new options there, no tests will fail. You noticed that in the initial implementation you forgot about `debug`

That's a good question. It'd be easy if JS provides keyword arguments, so we use them to reference the mandatory ones one and everything else can be passed as an optional object, but it doesn't.

Or we could define the generator options within the class, and dynamically build the constructor by iterating through its element, so it acts as the single source of truth. E.g:

```javascript
class Generator {
  constructor(templateName, targetDir, options) {
    const GENERATOR_OPTIONS = ['debug', 'disabledHooks', 'entrypoint', 'forceWrite', 'install', 'noOverwriteGlobs', 'output', 'templateParams'];
    GENERATOR_OPTIONS.forEach(option => {
      this[option] = options[option];
    });
  }
}

let gen = new Generator('foo', 'bar', { debug: 'a', disabledHooks: 'b', entrypoint: 'c' });
gen.debug            // 'a'
gen.disabledHooks    // 'b'
gen.entrypoint       // 'c'
gen.forceWrite       // undefined
gen.install          // undefined
gen.noOverwriteGlobs // undefined
gen.output           // undefined
gen.templateParams   // undefined
```

derberg: looks nice but then you lose what is nice currently, that options are destructured in params and we can use defaults there. I'm afraid now adding to your suggesting defaults, would complicate stuff 🤔 
did you try the library I mentioned? sounds like the perfect choice

sebastian-palma: nono I haven't tried it still, I wanted to know if we have to chance to do it without libraries.

Would you like to try the library and incorporate it to this PR or do a separate one for that?

derberg: nah, let's leave it, I was just curious, to use this type of decorators we would probably have to use babel or something anyway.

Just please move `GENERATOR_OPTIONS` to `generator.js` so we maintain the list where the constructor is, and just extend `getInvalidOptions` with an additional argument.

I think that whatever option we choose here to have "future errors secure implementation" will just add complexity that I'm not sure is worth it. Let's ignore my initial concern and react properly in the future where we see it is really a problem

derberg: @sebastian-palma we want to release 1.0.0-rc1 and it would be great to have this fix for the CLI in there. Do you agree with my comment and still want to work on it, or did you loose your patience because of me being picky 😅

sebastian-palma: Hey @derberg, just committed and squashed the changes. Sorry for the delay 😁

derberg:
@all-contributors please add @sebastian-palma for code

PR #361, opened 2020-05-26
fix: sorting issue for suggested params

derberg:
**Description**

- by default `sort()` takes number into string, so in case 3 and 11, 11 is first. I added custom sort comparison and extended tests

### DIFF
--- a/lib/templateConfigValidator.js
+++ b/lib/templateConfigValidator.js
@@ -65,9 +65,17 @@ function isRequiredParamProvided(configParams, templateParams) {
  * @param {Object} configParams Parameters specified in template configuration
  */
 function provideAdvice(wrongParams, configParams) {
+  const suggestion = (p) => {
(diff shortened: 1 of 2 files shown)


lib/templateConfigValidator.js
```diff
@@ -65,9 +65,17 @@ function isRequiredParamProvided(configParams, templateParams) {
  * @param {Object} configParams Parameters specified in template configuration
  */
 function provideAdvice(wrongParams, configParams) {
+  const suggestion = (p) => {
+    const sortInt = (a, b) => {
+      return a[0] - b[0];
+    };
```
fmvilas, line 71: Just a suggestion, you might like this syntax:
```suggestion
    const sortInt = (a, b) => a[0] - b[0];
```

Tenischev:
You will be never boring with reading of documentation in JS

Issue #358, opened 2020-05-23
conditionalFiles only works if you are lucky


Tenischev:
#### Describe the bug
1. Description of `conditionalFiles` in [docs](docs/authoring.md) is incomplete and example is actually pitfall.
2. If `subject` is not resolved in AsyncAPI, then file will be rendered because of [this condition](https://github.com/asyncapi/generator/blob/fc4dee9af622dfb198471a2cf0da53f5a3c7c7ae/lib/generator.js#L565)

#### How to Reproduce
Just try to use provided example

#### Expected behavior
1. Add note that path of file should be relative to `tempalte` folder
2. If `subject` is not found in AsyncAPI, then file is not rendered
3. Add note that `validation` is not any `JSON Schema Draft 07 object`, it must be from [`JSON Schema Validation`](https://tools.ietf.org/html/draft-handrews-json-schema-validation-01)
4. Provide a valid example:
Correct `subject` path in term of AsyncAPI spec to server protocols will be `servers.*.protocol`, but result of this (and any other) resolution will be an array, you should take it into account. Since it's not possible to predict how much servers will be in user AsyncAPI, usage of `const` is not possible as is. Perhaps, `contains` should be used always to check against result array. 
Here an example of `validation` that will correctly check for the `mqtt` protocol `{ "contains": { "type": "string", "const": "mqtt" }}`. 
Or another approach where filtering delegated to `subject`  - `"subject": "servers.*.{protocol: protocol} | [?protocol=='mqtt']", "validation": { "const": [ { "protocol": "mqtt" } ] }`

#### Hint
Perhaps we have a bug in [nodejs template](https://github.com/asyncapi/nodejs-template/blob/master/package.json) because it's exact copy of example from cuurent doc

derberg:
One thing sure here is that current conditionalFiles in nodejs-template do not make sense as files listed there do not exist in the tempalate anyway https://github.com/asyncapi/nodejs-template/blob/master/package.json#L88

derberg:
After some investigation:
- current configuration of `conditionalFiles` from nodejs-template can be removed https://github.com/asyncapi/nodejs-template/blob/master/package.json#L88 as files like `amqp.js` or `mqtt.js` do not exist in the template anymore
- I think the feature and the example work pretty well, just tested it in nodejs-template and it worked like a charm (my spec files has 3 servers). I added this:
```
"conditionalFiles": {
      "config/common.yml": {
        "subject": "server.protocol",
        "validation": {
[...]

PR #363, opened 2020-05-27
feat: support files listed in nonRenderableFiles with conditionalFiles

derberg:
**Description**

- nonRenderableFiles are also supported by conditionalFiles.
- extend some details in authoring.md

**Related issue(s)**
Resolves https://github.com/asyncapi/generator/issues/358


derberg:
@fmvilas 

I added support for nonrenderablefiles as agreed in the related issue

I tested against:
```
"nonRenderableFiles": [
      "src/api/middlewares/*.*",
      "src/lib/config.js"
    ],
    "conditionalFiles": {
      "src/lib/config.js": {
        "subject": "server.protocol",
        "validation": {
          "const": "amqp"
        }
      }
    }
```
and works as expected, `src/lib/config.js` is not getting through

ready for review

Tenischev:
@derberg i do not agree that this will resolve #358
First of all, example of "subject": "server.protocol", is not sufficient since there is no hint that every time when you wish to generate template, in the fact, you **must** specify `server` paramter otherwise `conditionalFiles` will no work.

derberg:
@Tenischev it is not yet merged, so for sure it doesn't resolve yet. But the goal is that it does 😉 

> First of all, example of "subject": "server.protocol", is not sufficient since there is no hint that every time when you wish to generate template, in the fact, you must specify server paramter otherwise conditionalFiles will no work.

- there are 2 examples now
- descriptions of server param is enriched

Do you like it 😄 
If not I need some hints on how to make you happy 😄

Tenischev:
I ❤️  new `server` parameter description 👏 😄 

But there still issue about these 
> if `subject` is not resolved in AsyncAPI, then file will be rendered because of [this condition](https://github.com/asyncapi/generator/blob/fc4dee9af622dfb198471a2cf0da53f5a3c7c7ae/lib/generator.js#L565)

derberg:
@Tenischev oh, I know what you mean, that actually sucks, you are right, in this new example I added about `info.contact`, if entire `contact` is not provided, `support.html` doc will be created

so I made a fix but to fix it nice there would have to be a nice log message. But we have too many of those already, so I added another use case to our new `debug` flag.

example of log messages for different cases, where there is debug on or off
```
$ ./cli.js ../nodejs-template/asyncapi.yml ../nodejs-template -o output --force-write -p server=dummy-mqtt --debug
docs/support.html was not generated because info.contact specified in template configuration in conditionalFiles was not found in provided AsyncAPI specification file

Done! ✨
Check out your shiny new generated files at /Users/wookiee/Documents/sources/generator/output.

$ ./cli.js ../nodejs-template/asyncapi.yml ../nodejs-template -o output --force-write -p server=dummy-mqtt --debug

docs/support.html was not generated because condition specified for this file in template configuration in conditionalFiles matched

Done! ✨
Check out your shiny new generated files at /Users/wookiee/Documents/sources/generator/output.

$ ./cli.js ../nodejs-template/asyncapi.yml ../nodejs-template -o output --force-write -p server=dummy-mqtt


Done! ✨
Check out your shiny new generated files at /Users/wookiee/Documents/sources/generator/output.
```

docs/authoring.md
```diff
@@ -211,17 +211,17 @@ The `generator` property from `package.json` file must contain a JSON object tha
     }
   },
   "conditionalFiles": {
-    "src/api/adapters/amqp.js": {
+    "template/dir/relative/path/to/file/test-amqp.js": {
```
Tenischev, line 214: Without `template` in the beginning, because of https://github.com/asyncapi/generator/blob/736bfa0108d158da07d86acaffbc57438c402f73/lib/generator.js#L546

docs/authoring.md
```diff
@@ -211,17 +211,17 @@ The `generator` property from `package.json` file must contain a JSON object tha
...
       "validation": {
         "const": "amqp"
       }
     },
-    "src/api/adapters/mqtt.js": {
-      "subject": "server.protocol",
-      "validation": {
-        "const": "mqtt"
-      }
+    "template/dir/relative/path/to/file/support.html": {
```
Tenischev, line 220: Without `template` in the beginning, because of https://github.com/asyncapi/generator/blob/736bfa0108d158da07d86acaffbc57438c402f73/lib/generator.js#L546

lib/generator.js
```diff
@@ -158,6 +159,8 @@ class Generator {
       throw new Error('Parameter entrypoint is required when using output = "string"');
     }
 
+    if (this.debug) log.setLevel('debug');
```
Tenischev, line 162: Nice idea! Could we pass debug flag into template, so in template we also could write some debug logs. (Ofc this suggestion is for separate issue)

derberg: you for sure have access to it in hooks out of the box, but filters and Nunjucks, that would be awesome, please write an issue for it

docs/authoring.md
derberg: ok, I think I made it a bit too confusing, I meant here in example that path must be relative to template dir. I'll fix

derberg:
@Tenischev how about https://github.com/asyncapi/generator/pull/363/commits/1c0dccd4774ff15506e58a731b9d7127beabb051

PR #364, opened 2020-05-27
refactor: remove streetlight example for testing

### LINKED ISSUE #238: Create a dummy-like asyncapi document for testing templates functionality
**Reason:**
At the moment, we manually test templates on a streetlight.yml example that we also modify from time to time. At some point of time, it will become a Frankenstein, and that is not the purpose of streetlight example. We should have a very dummy example that is easy to extend with all possible use cases. Later we could share it as an official test document that other templates should be tested against

**Description:**
- Take current streetlight and replace with dummy yaml
- Make sure it works with current templates
- remove streetlights example
- updated contribution docs with info that every enhancement in templates should extend dummy example if needed

derberg:
**Description**

- remove streetlights yaml file
- add new dummy file
- update docs

Long term when we go more and more functional in templates, I believe there will be a need to separate dummy file into: dummy-ws, dummy-amqp, dummy-n files becase there might be a case in future we add support for more extensions and bindings and simply having all in one will probably not be possible (that is just an assumption)

**Related issue(s)**
Resolves #238


test/templateConfigValidator.test.js
```diff
@@ -1,7 +1,7 @@
 const { validateTemplateConfig } = require('../lib/templateConfigValidator');
 const fs = require('fs');
 const path = require('path');
-const streetlightYAML = fs.readFileSync(path.resolve(__dirname, './docs/streetlights.yml'), 'utf8');
+const streetlightYAML = fs.readFileSync(path.resolve(__dirname, './docs/dummy.yml'), 'utf8');
```
jonaslagoni, line 4: Maybe rename to `dummyYAML` since streetlight is no longer a thing in the generator :thinking:?

derberg: of course, you are right, great catch, on it

derberg:
@fmvilas I need your another approval, I removed streetlight named variables from tests as per @jonaslagoni suggestion

PR #380, opened 2020-06-10
Rename {.gitignore} and {.npmignore}

fmvilas:
Since we're using npm to install templates, files like `.gitignore` and `.npmignore` are never downloaded (that's npm behavior we can't change).

Therefore, the only solution is to have templates rename `.gitignore` and `.npmignore` files to something different that npm will not understand. I chose `{.gitignore}` and `{.npmignore}` respectively. It's now the job of the Generator to rename them to become `.gitignore` and `.npmignore`.

### DIFF
--- a/lib/generator.js
+++ b/lib/generator.js
@@ -44,6 +44,8 @@ const FILTERS_DIRNAME = 'filters';
 const HOOKS_DIRNAME = 'hooks';
 const CONFIG_FILENAME = 'package.json';
 const PACKAGE_JSON_FILENAME = 'package.json';
+const GIT_IGNORE_FILENAME = '{.gitignore}';
+const NPM_IGNORE_FILENAME = '{.npmignore}';
 const ROOT_DIR = path.resolve(__dirname, '..');
 const DEFAULT_TEMPLATES_DIR = path.resolve(ROOT_DIR, 'node_modules');
 const TEMPLATE_CONTENT_DIRNAME = 'template';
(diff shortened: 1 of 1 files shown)


lib/generator.js
```diff
@@ -548,7 +550,8 @@ class Generator {
   async generateFile(asyncapiDocument, fileName, baseDir) {
     const sourceFile = path.resolve(baseDir, fileName);
     const relativeSourceFile = path.relative(this.templateContentDir, sourceFile);
-    const targetFile = path.resolve(this.targetDir, relativeSourceFile);
+    const targetFile = path.resolve(this.targetDir, this.maybeRenameSourceFile(relativeSourceFile));
+    console.log(targetFile);
```
derberg, line 554: remove this console.log

lib/generator.js
```diff
@@ -580,6 +583,21 @@ class Generator {
     await writeFile(targetFile, parsedContent, { encoding: 'utf8', mode: stats.mode });
   }
 
+  /**
+   * It may rename the source file name in cases where special names are not supported, like .gitignore or .npmignore.
```
derberg, line 587: would be great to have clearly mentioned here what we have in pr description, that it is here because of npm

lib/generator.js
fmvilas: LOL thanks!

fmvilas: BTW, this would be a great use case for the debug param. I want to see the name of the target file but of course I don't want to see the rest of the log messages.

PR #421, opened 2020-10-02
refactor: Use official release in Dockerfile instead of sources

### LINKED ISSUE #399: Improve docker image to contain npm global installation only instead of generator sources
#### Reason/Context
At the moment we provide official docker image that contains generator sources https://github.com/asyncapi/generator/blob/master/Dockerfile#L18-L20
So in the docker image, we do not enable users to use official release but sources of the release. The official release is what goes to npm and this is what Docker should use. This means we should do something like this in the docker image https://github.com/Mermade/oas-kit/blob/master/packages/swagger2openapi/Dockerfile#L8

Another reason to do it is to make sure users never get this issue https://github.com/asyncapi/generator/issues/388#issuecomment-675360044

#### Description
- improve Dockerfile
- make sure docs instructions are still ok

mowies:
**Description**

- Adjusted the Dockerfile so that it uses the official npm package instead of building from local source code.
- Updated the README accordingly (+ found a small inconsistency in a code example for the cli)

**Related issue(s)**
Resolves #399

### DIFF
--- a/README.md
+++ b/README.md
@@ -98,7 +98,7 @@ You might want to update your local installation of generator for different reas
 ### CLI usage
 
 ```bash
-Usage: cli [options] <asyncapi> <template>
+Usage: ag [options] <asyncapi> <template>
 
 - <asyncapi>: Local path or URL pointing to AsyncAPI specification file
 - <template>: Name of the generator template like for example @asyncapi/html-template or https://github.com/asyncapi/html-template
--- a/README.md
+++ b/README.md
@@ -178,11 +178,12 @@ docker run --rm -it \
 -v [ASYNCAPI SPEC FILE LOCATION]:/app/asyncapi.yml \
 -v [GENERATED FILES LOCATION]:/app/output \
 asyncapi/generator [COMMAND HERE]
+
 # Example that you can run inside generator directory after cloning this repository. First you specify mount in location of your AsyncAPI specification file and then you mount in directory where generation result should be saved.
 docker run --rm -it \
 -v ${PWD}/test/docs/dummy.yml:/app/asyncapi.yml \
 -v ${PWD}/output:/app/output \
-asyncapi/generator -o ./output asyncapi.yml @asyncapi/html-template --force-write
+asyncapi/generator -o /app/output /app/asyncapi.yml @asyncapi/html-template --force-write
 ```
 
 ### CLI usage with npx instead of npm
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -178,11 +178,12 @@ docker run --rm -it \
 -v [ASYNCAPI SPEC FILE LOCATION]:/app/asyncapi.yml \
 -v [GENERATED FILES LOCATION]:/app/output \
 asyncapi/generator [COMMAND HERE]
+
 # Example that you can run inside generator directory after cloning this repository. First you specify mount in location of your AsyncAPI specification file and then you mount in directory where generation result should be saved.
 docker run --rm -it \
 -v ${PWD}/test/docs/dummy.yml:/app/asyncapi.yml \
 -v ${PWD}/output:/app/output \
-asyncapi/generator -o ./output asyncapi.yml @asyncapi/html-template --force-write
+asyncapi/generator -o /app/output /app/asyncapi.yml @asyncapi/html-template --force-write
```
derberg, line 186: you think it is better to do `/app/asyncapi.yml` instead of `asyncapi.yml` in docs? will it be less confusing? both ways work as `app` is a workdir

derberg:
@mowies lgtm, I just added one question to make sure what were your intentions

I also changed the PR title prefix as it is not something that should trigger release because the Docker image still works as before for users. Now it is just better to play with it if someone knows `ag` and want to interact with image differently by overriding the entrypoint like this for example `--entrypoint sh`

README.md
mowies: Yes, I think it makes the example a little bit clearer. You see immediately that the volumes you just mounted in `/app/asyncapi.yml` and `/app/output/` are used directly in the command you pass to the CLI.
Also, the docker image workdir can only be seen if a user actually looks at the `Dockerfile`. So in my opinion, this approach makes it a bit easier for docker beginners who just want to run the thing but don't care about the details.

derberg (approved):

derberg:
@mowies thanks a lot for your contribution. In case you are interested with more Hacktoberfest issues from AsyncAPI then please have a look at [this](https://www.asyncapi.com/blog/hacktoberfest-2020) blog post. Among other things, this blog post mentions a link to excel sheet where we maintain a list of all the issues that you can work on ;)

derberg:
@mowies would you mind sharing with me how did you find out about AsyncAPI and this issue? we made some effort into Hacktoberfest and want to identify what efforts were the best for contributors and how they learned about us, so next Hacktoberfest is even better

mowies:
I found the issue through GitHub's search. I specifically looked for docker related Hacktoberfest issues on there and stumbled upon your ticket

derberg:
@mowies  thanks a lot for the prompt reply 🙏

Issue #416, opened 2020-09-18
Move function from watcher before the for loop


derberg:
**Context**

SonarCloud info:

> Defining a function inside of a loop can yield unexpected results. Such a function keeps references to the variables which are defined in outer scopes. All function instances created inside the loop therefore see the same values for these variables, which is probably not expected.

So yeah, never create new instance of a function inside every loop, unless this is explicitly required by some specific library

**Description**

Declare [this](https://github.com/asyncapi/generator/blob/master/lib/watcher.js#L33) anonymous function before the `for` loop.﻿

PR #422, opened 2020-10-02
refactor: Moved function from watcher before the for loop (#416)

slavakr:
**Description**
This should fix SonarCloud issue:

`Defining a function inside of a loop can yield unexpected results. Such a function keeps references to the variables which are 
defined in outer scopes. All function instances created inside the loop therefore see the same values for these variables, which is 
probably not expected.`

**Related issue(s)**
#416

### DIFF
--- a/lib/watcher.js
+++ b/lib/watcher.js
@@ -23,15 +23,25 @@ class Watcher {
   }
 
   /**
+   * Initiates watch on a path.
+   * @param {*} path The path the watcher is listening on.
+   * @param {*} changeCallback Callback to call when changed occur.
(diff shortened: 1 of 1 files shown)


derberg:
Thanks @slavakr! In case you are interested in more hacktoberfest related issues from AsyncAPI, they are all listed here https://www.asyncapi.com/blog/hacktoberfest-2020

derberg (approved):

derberg:
@slavakr would you mind sharing with me how did you find out about AsyncAPI and this issue? we made some effort into Hacktoberfest and want to identify what efforts were the best for contributors and how they learned about us, so next Hacktoberfest is even better

Issue #391, opened 2020-07-24
The generate does not install template modules correctly if local 


jonaslagoni:
#### Describe the bug
If you use the generator with a local path for a template i.e. `ag "./streetlight.json" "./ts-nats-template"` it does install all the modules however it still says missing module when moving to the generation phase.

```
npm http fetch GET 304 https://registry.npmjs.org/is-url 489ms (from cache)
npm http fetch GET 304 https://registry.npmjs.org/lodash 503ms (from cache)
npm http fetch GET 304 https://registry.npmjs.org/quicktype-core 501ms (from cache)
npm http fetch GET 304 https://registry.npmjs.org/@asyncapi%2fgenerator-filters 1359ms (from cache)
npm http fetch GET 304 https://registry.npmjs.org/markdown-it 53ms (from cache)
npm http fetch GET 304 https://registry.npmjs.org/openapi-sampler 55ms (from cache)
npm http fetch GET 304 https://registry.npmjs.org/entities 40ms (from cache)
npm http fetch GET 304 https://registry.npmjs.org/json-pointer 52ms (from cache)
npm http fetch GET 304 https://registry.npmjs.org/foreach 45ms (from cache)
npm http fetch GET 304 https://registry.npmjs.org/collection-utils 66ms (from cache)
npm http fetch GET 304 https://registry.npmjs.org/js-base64 68ms (from cache)
npm http fetch GET 304 https://registry.npmjs.org/pluralize 82ms (from cache)
npm http fetch GET 304 https://registry.npmjs.org/browser-or-node 72ms (from cache)
npm http fetch GET 304 https://registry.npmjs.org/wordwrap 147ms (from cache)
npm http fetch GET 304 https://registry.npmjs.org/urijs 149ms (from cache)
npm http fetch GET 304 https://registry.npmjs.org/pako 153ms (from cache)
npm http fetch GET 304 https://registry.npmjs.org/readable-stream 152ms (from cache)
npm http fetch GET 304 https://registry.npmjs.org/yaml 156ms (from cache)
npm http fetch GET 304 https://registry.npmjs.org/isomorphic-fetch 158ms (from cache)
npm http fetch GET 304 https://registry.npmjs.org/@mark.probst%2funicode-properties 936ms (from cache)
npm http fetch GET 304 https://registry.npmjs.org/unicode-trie 80ms (from cache)
npm http fetch GET 304 https://registry.npmjs.org/brfs 92ms (from cache)
[...]

PR #426, opened 2020-10-12
Use require algorithm to resolve templates

### LINKED ISSUE #425: The generator does not use templates installed in node_modules
#### The bug

This installs  @asyncapi/html-template twice.
```
npm i -D @asyncapi/generator  @asyncapi/html-template
./node_modules/.bin/ag src/simple.yaml @asyncapi/html-template --force-write 
```

repro: https://github.com/laat/repro-asyncapi-generator

#### Expected behavior
It should reuse installed templates. This lets me pin templates at a specific version

#### The repro has a patch

This solves the issue for me. Essentially `require.resolve` does the heavy lifting and should support the intended uses of the replaced code.

```patch
diff --git a/node_modules/@asyncapi/generator/lib/generator.js b/node_modules/@asyncapi/generator/lib/generator.js
index 841af20..dc62139 100644
--- a/node_modules/@asyncapi/generator/lib/generator.js
+++ b/node_modules/@asyncapi/generator/lib/generator.js
@@ -320,31 +320,12 @@ class Generator {
     return new Promise(async (resolve, reject) => {
       if (!force) {
         try {
-          let installedPkg;
-
-          if (isFileSystemPath(this.templateName)) {
-            const pkg = require(path.resolve(this.templateName, PACKAGE_JSON_FILENAME));
-            installedPkg = require(path.resolve(DEFAULT_TEMPLATES_DIR, pkg.name, PACKAGE_JSON_FILENAME));
-          } else { // Template is not a filesystem path...
-            const templatePath = path.resolve(DEFAULT_TEMPLATES_DIR, this.templateName);
-            if (await isLocalTemplate(templatePath)) {
-              // This "if" is covering the following workflow:
-              // ag asyncapi.yaml ../html-template
-              // The previous command installs a template called @asyncapi/html-template
-              // And now we run the command again but with the resolved name:
-              // ag asyncapi.yaml @asyncapi/html-template
-              // The template name doesn't look like a file system path but we find
-              // that the package is already installed and it's a symbolic link.
-              const { resolvedLink } = await getLocalTemplateDetails(templatePath);
-              log.debug(`This template has already been installed and it's pointing to your filesystem at ${resolvedLink}.`);
-            }
-            installedPkg = require(path.resolve(templatePath, PACKAGE_JSON_FILENAME));
-          }
-
+          const pkgJsonPath = require.resolve(path.join(this.templateName, PACKAGE_JSON_FILENAME))
+          const installedPkg = require(pkgJsonPath)
           return resolve({
             name: installedPkg.name,
             version: installedPkg.version,
-            path: path.resolve(DEFAULT_TEMPLATES_DIR, installedPkg.name),
+            path: path.dirname(pkgJsonPath),
           });
         } catch (e) {
           // We did our best. Proceed with installation...
```

laat:
fixes #425

An alternative solution could be to use something like https://github.com/sindresorhus/resolve-pkg

--
I'm unsure how we should test this, maybe some end to end test that actually generates from templates at different locations?

### DIFF
--- a/lib/__mocks__/utils.js
+++ b/lib/__mocks__/utils.js
@@ -13,9 +13,6 @@ utils.fetchSpec = jest.fn(async (fileUrl) => {
   return utils.__contentOfFetchedFile;
 });
 
-utils.__isLocalTemplateValue = false;
-utils.isLocalTemplate = jest.fn(async () => utils.__isLocalTemplateValue);
-
 utils.__generatorVersion = '';
 utils.getGeneratorVersion = jest.fn(() => utils.__generatorVersion);
 
--- a/lib/__mocks__/utils.js
+++ b/lib/__mocks__/utils.js
@@ -13,15 +13,8 @@ utils.fetchSpec = jest.fn(async (fileUrl) => {
   return utils.__contentOfFetchedFile;
 });
 
-utils.__isFileSystemPathValue = false;
-utils.isFileSystemPath = jest.fn(() => utils.__isFileSystemPathValue);
-
 utils.__isLocalTemplateValue = false;
 utils.isLocalTemplate = jest.fn(async () => utils.__isLocalTemplateValue);
-
-utils.getLocalTemplateDetails = jest.fn(async () => ({
-  resolvedLink: utils.__getLocalTemplateDetailsResolvedLinkValue || '',
-}));
 
 utils.__generatorVersion = '';
 utils.getGeneratorVersion = jest.fn(() => utils.__generatorVersion);
--- a/lib/__mocks__/utils.js
+++ b/lib/__mocks__/utils.js
@@ -13,6 +13,9 @@ utils.fetchSpec = jest.fn(async (fileUrl) => {
   return utils.__contentOfFetchedFile;
 });
 
+utils.__isLocalTemplateValue = false;
+utils.isLocalTemplate = jest.fn(async () => utils.__isLocalTemplateValue);
+
 utils.__generatorVersion = '';
 utils.getGeneratorVersion = jest.fn(() => utils.__generatorVersion);
 
--- a/lib/filtersRegistry.js
+++ b/lib/filtersRegistry.js
@@ -1,7 +1,7 @@
 const path = require('path');
 const fs = require('fs');
 const xfs = require('fs.extra');
-const { isLocalTemplate } = require('./utils');
+const resolvePkg = require('resolve-pkg');
 
 /**
  * Registers all template filters.
@@ -12,7 +12,7 @@ const { isLocalTemplate } = require('./utils');
  */
 module.exports.registerFilters = async (nunjucks, templateConfig, templateDir, filtersDir) => {
   await registerLocalFilters(nunjucks, templateDir, filtersDir);
-  await registerConfigFilters(nunjucks, templateDir, templateConfig);
+  registerConfigFilters(nunjucks, templateDir, templateConfig);
 };
 
 /**
@@ -40,7 +40,7 @@ function registerLocalFilters(nunjucks, templateDir, filtersDir) {
         const mod = require(filePath);
 
         addFilters(nunjucks, mod);
-        
+
         next();
       } catch (e) {
         reject(e);
@@ -64,19 +64,15 @@ function registerLocalFilters(nunjucks, templateDir, filtersDir) {
 * @param {String} templateDir Directory where template is located.
 * @param {Object} templateConfig Template configuration.
 */
-async function registerConfigFilters(nunjucks, templateDir, templateConfig) {
-  const confFilters = templateConfig.filters; 
-  const DEFAULT_MODULES_DIR = 'node_modules';
+function registerConfigFilters(nunjucks, templateDir, templateConfig) {
+  const confFilters = templateConfig.filters;
   if (!Array.isArray(confFilters)) return;
 
-  const promises = confFilters.map(async el => {
-    const modulePath = await isLocalTemplate(templateDir) ? path.resolve(templateDir, DEFAULT_MODULES_DIR, el) : el;
+  confFilters.forEach(async el => {
+    const modulePath = resolvePkg(el, { cwd: templateDir });
     const mod = require(modulePath);
-
     addFilters(nunjucks, mod);
   });
-
-  await Promise.all(promises);
 }
 
 /**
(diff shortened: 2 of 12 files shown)


laat:
Actually, I think https://github.com/sindresorhus/resolve-pkg is better. Because `require.resolve()` locates relative paths from `__dirname` instead of `process.cwd()`. I think the relative paths given to the `ag` cli is ment to be cwd-relative.

laat:
@derberg updated the PR to use `resolve-pkg` instead. It fixed some issue I had when installing globally

derberg (changes requested):
Excelent work. I have few remarks so far:

- `isLocalTemplate` function that was removed is also used in the `cli.js` within the watcher (I'm surprised that neither linter nor SourceCloud didn't spot it)
- why did you remove all those tests? those test cases are still valid right?
- in place where you do `const pkgPath = resolvePkg(this.templateName);` it would be super useful to do `console.info` of the path if `debug` flag is passed as I was surprised at the beginning what kind of magic is happening here 😄  something like 
   ```
   if (pkgPath && debug) console.info(`Template sources taken from ${pkgPath}`)
   ```


So far I see the following scenario fails. I have template locally and it doesn't have deps installed there and it obviously fails (even though while doing the test, I had html-template installed in node-modules of the generator):
```
./cli.js https://bit.ly/asyncapi-dummy ../html-template -o output --force-write
Something went wrong:
Error: Cannot find module 'rimraf'
Require stack:
- /Users/wookiee/sources/html-template/hooks/01_removeNotRelevantParts.js
- /Users/wookiee/sources/generator/lib/hooksRegistry.js
- /Users/wookiee/sources/generator/lib/generator.js
- /Users/wookiee/sources/generator/cli.js
    at Function.Module._resolveFilename (internal/modules/cjs/loader.js:965:15)
    at Function.Module._load (internal/modules/cjs/loader.js:841:27)
    at Module.require (internal/modules/cjs/loader.js:1025:19)
    at require (internal/modules/cjs/helpers.js:72:18)
    at Object.<anonymous> (/Users/wookiee/sources/html-template/hooks/01_removeNotRelevantParts.js:3:16)
    at Module._compile (internal/modules/cjs/loader.js:1137:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:1157:10)
    at Module.load (internal/modules/cjs/loader.js:985:32)
    at Function.Module._load (internal/modules/cjs/loader.js:878:14)
    at Module.require (internal/modules/cjs/loader.js:1025:19)
```

I will later do some more test in docker in a more isolated environment but you already have some comments to work on

laat:
> `isLocalTemplate` function that was removed is also used in the `cli.js` within the watcher (I'm surprised that neither linter nor SourceCloud didn't spot it)

I should have noticed...

> why did you remove all those tests? those test cases are still valid right?

Inside `Generator.installTemplate` I had to remove calls to `isFileSystemPath`, and the tests I deleted all relied on mocking of that call. I do not know of a better way to test this than to test the CLI in some sort of end 2 end test.

Tests like this perhaps:
```bash
#!/usr/bin/env bash
node cli.js api.yaml @asyncapi/html-template
```

> So far I see the following scenario fails.

I do not think this scenario has ever worked, because when we do a `require('rimraf')` in a file called `/apps/my-app/foobar.json` it looks for rimraf in `/apps/my-app/node_modules`, then `/apps/node_modules` and lastly `/node_modules` and nowhere else. Since the dependencies needed for `../html-template` is not installed in a `node_modules` folder that it can reach, the require fails.

I do not think we should support that use case.

laat:
> > So far I see the following scenario fails.
> 
> I do not think this scenario has ever worked

Well, it might have. I'm not sure. But it's strange to `npm install` stuff in other folders outside the working directory. Is this something we must support?

derberg:
@laat it is a valid use case, but it doesn't work as you think (that npm install ran in html-template folder) but now it works this way that the local template ends up in node-modules for local generator. On mac it always worked perfectly, only for windows looks like it is a problem https://github.com/asyncapi/generator/issues/391 and one has to manually install deps for the local template.

Regarding tests, you are totally right, best would be integrations tests. Generator can be used as a library. I have tests like this (template integration tests, using generator) in here https://github.com/asyncapi/template-for-generator-templates/pull/1/files#diff-952a085e44dea8bc915968669bccf8fc5340f8b42d7b13fe66b85a8d5d293b1f. So you can basically run generator in a test and even evaluate if the resulting file did not change.

laat:
I'm unable to test this properly, and do not even know which test cases that must be supported.

derberg:
@laat the only test case that doesn't work is the one I described before, you can test it manually, just get a template on your local and run generation, instead of pointing to template package name, point to template filesystem location

laat:
@derberg I will make a new attempt at a PR for this, because it became difficult to fix the remaining issue. I did not test the watch use case, which also must work.

My thinking right now is this:

* Without this cli argument, we should not npm install anything.
  ```-i, --install                  installs the template and its dependencies```
* With the `-i` argument and the dependency is available remotely (git, npm). Install it into the current working directory `node_modules`.
* With the `-i` argument and the dependency is a local directory `./some/path` relative to the working directory. Install the dependencies into `node_modules`
* With the `-i` argument and the dependency is not local `../some/path` relative to the working directory. Fail with a error message stating that it cannot install, and that the user should either `npm link` it or do `npm install` in that directory. I believe this should be outside the scope of what it should do because it makes everything more complex. 

  This is because changing the installation directory from `/node_modules/@asyncapi/generator/node_modules` to `{working_directory}/node_modules` makes resolution of dependencies difficult. I believe modifying anything inside `/node_modules/@asyncapi/generator` after installation should be avoided.
* `--watch` should watch the directory that is resolved by `resolve-pkg` (not handled by this PR afaik)

This should fix #425 and hopefully #391

It will take some time to do this, as I have little time at the moment :)

derberg:
@laat these assumptions are not good as it basically means that one, will always have to pass `-i` when running `ag asyncapi.yml @asyncapi/html-template` or did I misunderstood something?

laat:
@derberg How about extracting the current template resolution into a separate class? A strategy pattern for template resolution.

So move current implementation as `DefaultTemplateResolution` class and the Generator constructor just defaults to using that strategy? Then I can work on changing the resolution strategy in my projects because I use the Generator class programmatically, and send a PR when I have found something that works.

PR #437, opened 2020-10-28
chore: add chromium to docker for html-template

derberg:
**Description**

- add chromium to docker for html-template

**Related issue(s)**
Fixes https://github.com/asyncapi/html-template/issues/112

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,8 +2,13 @@ FROM node:12-alpine
 
 WORKDIR /app
 
+# Since 0.14.0 release of html-template chromium is needed for pdf generation
+ENV PUPPETEER_EXECUTABLE_PATH /usr/bin/chromium-browser
+ENV PUPPETEER_SKIP_CHROMIUM_DOWNLOAD true
+
 # Since 0.30.0 release Git is supported and required as a dependency
-RUN apk --update add git && \
+# Since 0.14.0 release of html-template chromium is needed for pdf generation
+RUN apk --update add git chromium && \
     rm -rf /var/lib/apt/lists/* && \
     rm /var/cache/apk/*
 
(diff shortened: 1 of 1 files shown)


jonaslagoni (commented):
@derberg is it not possible to get around adding a global package to the dockerfile to accommodate 1 template? It seems... Wrong 😄

derberg:
@jonaslagoni yeah, I know what you mean, but there is no other way really. At the end, html-template is an official template, so I do not see this as a big problem.

jonaslagoni:
> @jonaslagoni yeah, I know what you mean, but there is no other way really. At the end, html-template is an official template, so I do not see this as a big problem.

Can't we imagine this is something everyone could benefit from, adding a feature which controls the docker environment per template? I can understand if we just want to make this a temporary fix, however there should be an issue where this is raised IMO.

derberg:
@jonaslagoni tbh I have no idea how else this could be solved, other than saying `create your own image using generator image as a base`. I also cannot really imagine any other template needing such a custom approach. Puppeteer is just unique here needing dedicated chromium instance for alpine. What I'm trying to say is that any followup issue will just be food for the stale bot 😄 as we do not have a use case for such extensibility. Maybe I add a comment that in case more customizations are needed for other templates, we should consider reworking the image?

jonaslagoni:
> What I'm trying to say is that any followup issue will just be food for the stale bot 😄 as we do not have a use case for such extensibility. Maybe I add a comment that in case more customizations are needed for other templates, we should consider reworking the image?

I mean, you just had the use case 😄 If 1 template requires custom dependencies, others will probably too, even if we can't think of a reason atm. I mean there are different methods to make it require dependencies, extending the generator docker image, make a framework for shell files, **make** files etc. But I agree atm this is too much, but I still think an issue would be valid. Ill accept the PR on the notion that this is temporary 😄

jonaslagoni (approved):
see comment.

derberg:
@jonaslagoni yo, can you approve again, I added a comment to dockerfile

jonaslagoni (approved):

Issue #333, opened 2020-05-12
Add new flag to disable specific hook function


derberg:
#### Reason/Context
Once this is merged https://github.com/asyncapi/generator/issues/318 we have hooks types (`generate:after`) and hook names (`createAsyncAPIFile`).

You can now use flag `--disable-hook` to disable hook type, but there is no option to disable only a specific hook.

#### Description

- Introduce new flag/option where user of a particular template can disable only one specific hook
- Come up with suggestion on how info about hooks should be documented in the template, so folks know what hooks can be disabled without reading the code

magicmatatjahu:
@derberg Could I take it? I have an idea to use `--disable-hook` flag for this and extend the flag's argument to something like

```bash
--disable-hook <hookType>#<hookName>
```

Then generator only skip hook with given name :)

I know that hookName is not required, so at the moment it should works only for common hooks like `createAsyncApiFile` from `@asyncapi/generator-hooks`.

I am not a expert in generator and its source code, so probably I don't see problems with this approach. What do you think?

> Come up with suggestion on how info about hooks should be documented in the template, so folks know what hooks can be disabled without reading the code

Simple list of used hooks with names, like:

|Type|Name|Desc|
|-|-|-|
|`generate:after`|`createAsyncApiFile`|`foobar`|

eventually in `desc` points to desc in another repo, if hook is not written in template itself.

Then user should known which hooks can disable.

derberg:
> --disable-hook <hookType>#<hookName>

love it but as log as `--disable-hook <hookType>` will still work as before

> I know that hookName is not required, 

what do you mean?

> I am not a expert in generator and its source code, so probably I don't see problems with this approach. What do you think?

I don't remember implementation details but should be fine

---
table looks good, but `name` should be first imho. Even if hook comes from external lib, it should still be described in the repo where it is used imho

Aynway, take it man

magicmatatjahu:
> love it but as log as --disable-hook <hookType> will still work as before 

Yes, If user put only `--disable-hook <hookType>` then whole hooks of given type will be skipped, but if put `--disable-hook <hookType>#<hookName>`, then only particular hook (with <hookName> name) of <hookType> will be disabled.

> >I know that hookName is not required,
>
> what do you mean?

[...]

PR #442, opened 2020-11-02
chore: disable specific hook function(s)

magicmatatjahu:
**Description**

- Extend `--disable-hook` flag to disable specific hook function(s). Example usage:

```text
--disable-hook generate:before generate:after=foo,bar
```

then generator will disable all hooks of `generate:before` type and `foo`, `bar` hooks of `generate:after` type.

However, the logic in the code overrides previous values for given type, so using flag such as: `--disable-hook generate:before generate:before=foo,bar` disables only `foo`, `bar` hooks of `generate:before` type, not whole type.

- update Readme.md
- update tests

**Related issue(s)**
Resolves https://github.com/asyncapi/generator/issues/333

### DIFF
--- a/lib/generator.js
+++ b/lib/generator.js
@@ -754,11 +754,11 @@ class Generator {
     const disabledHooks = this.disabledHooks[hookName] || [];
     if (disabledHooks === true) return;
 
-    const hookNames = this.hooks[hookName];
-    if (!Array.isArray(hookNames)) return;
-    const promises = hookNames.map(async (hook) => {
+    const hooks = this.hooks[hookName];
+    if (!Array.isArray(hooks)) return;
+    const promises = hooks.map(async (hook) => {
       if (typeof hook !== 'function') return;
-      if (disabledHooks.includes(hook)) return;
+      if (disabledHooks.includes(hook.name)) return;
       return await hook(this, hookArguments);
     }).filter(Boolean);
 
--- a/lib/generator.js
+++ b/lib/generator.js
@@ -75,7 +75,7 @@ class Generator {
    * @param {String} [options.templateParams]   Optional parameters to pass to the template. Each template define their own params.
    * @param {String} [options.entrypoint]       Name of the file to use as the entry point for the rendering process. Use in case you want to use only a specific template file. Note: this potentially avoids rendering every file in the template.
    * @param {String[]} [options.noOverwriteGlobs] List of globs to skip when regenerating the template.
-   * @param {Object<String, Boolean | String | String[]>} [options.disabledHooks] Object with hooks to disable. The key is a hook type. If key has "true" value, then the generator skips all hooks from the given type. If the key is a string with the name of a single hook, then the generator skips only this single hook name. If the key is an array of strings, then the generator skips only hooks from the array.
+   * @param {Object<String, Boolean | String | String[]>} [options.disabledHooks] Object with hooks to disable. The key is a hook type. If key has "true" value, then the generator skips all hooks from the given type. If the value associated with a key is a string with the name of a single hook, then the generator skips only this single hook name. If the value associated with a key is an array of strings, then the generator skips only hooks from the array.
    * @param {String} [options.output='fs'] Type of output. Can be either 'fs' (default) or 'string'. Only available when entrypoint is set.
    * @param {Boolean} [options.forceWrite=false] Force writing of the generated files to given directory even if it is a git repo with unstaged files or not empty dir. Default is set to false.
    * @param {Boolean} [options.install=false] Install the template and its dependencies, even when the template has already been installed.
@@ -96,7 +96,7 @@ class Generator {
     this.entrypoint = entrypoint;
     /** @type {String[]} List of globs to skip when regenerating the template. */
     this.noOverwriteGlobs = noOverwriteGlobs || [];
-    /** @type {Object<String, Boolean | String | String[]>} Object with hooks to disable. The key is a hook type. If key has "true" value, then the generator skips all hooks from the given type. If the key is a string with the name of a single hook, then the generator skips only this single hook name. If the key is an array of strings, then the generator skips only hooks from the array. */
+    /** @type {Object<String, Boolean | String | String[]>} Object with hooks to disable. The key is a hook type. If key has "true" value, then the generator skips all hooks from the given type. If the value associated with a key is a string with the name of a single hook, then the generator skips only this single hook name. If the value associated with a key is an array of strings, then the generator skips only hooks from the array. */
     this.disabledHooks = disabledHooks || {};
     /** @type {String} Type of output. Can be either 'fs' (default) or 'string'. Only available when entrypoint is set. */
     this.output = output;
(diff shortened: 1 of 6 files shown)


lib/generator.js
```diff
@@ -75,7 +75,7 @@ class Generator {
    * @param {String} [options.templateParams]   Optional parameters to pass to the template. Each template define their own params.
    * @param {String} [options.entrypoint]       Name of the file to use as the entry point for the rendering process. Use in case you want to use only a specific template file. Note: this potentially avoids rendering every file in the template.
    * @param {String[]} [options.noOverwriteGlobs] List of globs to skip when regenerating the template.
-   * @param {String[]} [options.disabledHooks] List of hook types to disable.
+   * @param {Object<String, Boolean | String[]>} [options.disabledHooks] Object with hooks to disable. The field is a hook type. If field has "true" value, then generator skips all hooks from given type. If field is array of strings, then generator skips only hooks from array.
```
derberg, line 78: Best would be if implementation support user to provide just one hook name as string, not array

```suggestion
   * @param {Object<String, Boolean | String | String[]>} [options.disabledHooks] Object with hooks to disable. The key is a hook type. If key has "true" value, then the generator skips all hooks from the given type. If the key is a string with the name of a single hook, then the generator skips only this single hook name. If the key is an array of strings, then the generator skips only hooks from the array. 
```

derberg (changes requested):
Great work. It is clear how to use it. I only suggested a slight change to the API

lib/generator.js
magicmatatjahu: I changed logic, but I have problem with `isHookAvailable` function (https://github.com/asyncapi/generator/blob/master/lib/generator.js#L770). Should it has extended signature `isHookAvailable(hookType, hookName)`? If user passes only hookType then function will behave as before, but when user gives hookName it will check a more extended form?

derberg: > If user passes only hookType then function will behave as before, but when user gives hookName it will check a more extended form

what do you mean? it is a private function, the user will not use it directly. 
`setFileTemplateName` is a special hook that lets you interact with the process of [file templates](https://github.com/asyncapi/generator/blob/master/docs/authoring.md#file-templates) generation, so you can influence the file name. So it is important to know if `setFileTemplateName` hook is available, not specific name of the hook. Maybe `isHookAvailable` should become `isHookTypeAvailable` or at least jsdoc extended?

magicmatatjahu: This function is public :) In JS when you want make method/property private then signature will be `#isHookAvailable`

magicmatatjahu: You can check tests here https://github.com/asyncapi/generator/pull/442/files#diff-d2f4f8516ea7f901574dcfd6ad677f6a3bcf85de3606ab5fd2a6a0be241147b3R520

derberg: Well, it is not public as it is marked as private and not visible in [API docs](https://github.com/asyncapi/generator/blob/master/docs/api.md)

magicmatatjahu: Even if this function should behave as private, I think that adapting the function to the new functionality is good idea :)

lib/generator.js
```diff
@@ -751,8 +751,11 @@ class Generator {
    * @private
    */
   async launchHook(hookName, hookArguments) {
-    const disabledHooks = this.disabledHooks[hookName] || [];
+    let disabledHooks = this.disabledHooks[hookName] || [];
     if (disabledHooks === true) return;
+    if (typeof disabledHooks === "string") {
+      disabledHooks = [disabledHooks];
+    }
```
derberg: I know it is just 3 lines of code, you could even do 1 with `if (typeof disabledHooks === "string") disabledHooks = [disabledHooks];`


but did you consider that `if (disabledHooks.includes(hook.name)) return;` could work without the above code because string also has `includes` function. Or am I overlooking something?

lib/generator.js
derberg: sure, just tell me what is the use case for this new functionality 😄

lib/generator.js
magicmatatjahu: Yes, string has also `includes` function, but includes for string is something like a wrapper for regexp so `disabledHooks = [disabledHooks];` is easier way to disable hook -> e.g. if disabledHooks will have `foobar` name  and then single hook function will have `foo` name, then condition `if (disabledHooks.includes(hook.name))` will return true

derberg: @magicmatatjahu did you rebase? strange that this comment got outdated? please don't rebase

Anyway, valid point, I agree with you

lib/generator.js
magicmatatjahu: If someone wants to use the generator as a library (not cli) ) then after reading hooks into the registry, person can check if a hook with a given name exists (I know a function is private in JSDoc, but someone can always know about this function). If this use case is not important for you, I can remove the extra code :)

lib/generator.js
magicmatatjahu: This is outdated because, as you suggested, change 3 lines to 1 line :P

lib/generator.js
fmvilas:
```
* @param {Object<String, Boolean | String | String[]>} [options.disabledHooks] Object with hooks to disable. The key is a hook type. If key has "true" value, then the generator skips all hooks from the given type. If the key is a string with the name of a single hook, then the generator skips only this single hook name. If the key is an array of strings, then the generator skips only hooks from the array. 
```

I'm not sure if I'm understanding correctly but seems to me this explanation is wrong. It says: `If the key is a string with the name of a single hook, then the generator skips only this single hook name.  If the key is an array of strings, then the generator skips only hooks from the array. `. I think it should be: `If the value associated with a key is a string with the name of a single hook, then the generator skips only this single hook name.  If the value associated with a key is an array of strings, then the generator skips only hooks from the array. `

Aside from that, I can't really understand what's the use case. @magicmatatjahu you described **how** you would use this feature but I'm more interested in the **why** or **what for**.

magicmatatjahu: @fmvilas Thanks! Your description is more verbose. Some people may be confused on reading `If the key is a string with the name of a single hook, then the generator skips only this single hook name.`

I have only one use case in my mind for new signature of `isHookAvailable`. When I will use generator as a library, I would like to check if the specific hook of given hook type is available in current template executing. What does it mean? E.g: I'm using html template, but before render template, I want check if `generatePdf` hook is available (also not disabled). So for this I have `isHookAvailable` function. If hook is disabled/doesn't exist, then I would like perform some additional logic, if yes another logic etc. Of course as I wrote, if this use case is not important I can revert changes to old body of function.

derberg: > If hook is disabled/doesn't exist, then I would like perform some additional logic, if yes another logic etc.

What logic woudl it be? I think the case here is that in general, better is to add logic that one can back with real world example, otherwise, it is definitely better to not add additional code when not needed

jonaslagoni: I don't see a use case for it being a public function. If you are using the generator as a library instead of the cli you would have access to the provided generator options and therefore the disabled hook right? 😄

magicmatatjahu: YOU WIN!!! 😄 I changed to old arguments, but I had to change body to adjust for new `disabledHooks` type -> from array to object.

derberg (dismissed):
I did some manual tests, works like a charm.
Please bump package.json and package-lock.json to rc 15 as this is the version I will manually release once we merge this one

derberg (approved):

Issue #441, opened 2020-10-30
Unable to generate html with array of arrays schema input


rez10191:
#### Describe the bug
I've encountered the following error when trying to create an HTML document page for my AsyncAPI 2.0.0 schema.  I believe the schema to be valid by cross checking with [spectral lint](https://github.com/stoplightio/spectral).  This schema did work in earlier versions of `asynapi/generator` (last I'm aware of for sure was `asyncapi-generator@0.7.10`).

#### How to Reproduce
Create a docker container with the latest version from this repo (`1.0.0-rc.13`).
Use the following schema (save as `test.json`)
```
{
    "asyncapi": "2.0.0",
    "id": "https://www.example.com",
    "info": {
        "title": "Test",
        "version": "0.0.1",
        "description": "Test of array of arrays",
        "termsOfService": "https://www.example.com",
        "contact": {
            "name": "API Support",
            "url": "https://www.example.com",
            "email": "support@example.com"
        },
        "license": {
            "name": "License",
            "url": "https://www.example.com"
        }
    },
    "tags": [],
    "servers": {
        "demo": {
            "url": "wss://example.com/...",
            "description": "",
            "protocol": "wss",
            "protocolVersion": "13",
            "security": [
               {
                  "http": []
               }
            ]
        }
    },
    "channels":{
        "/": {
            "subscribe": {
                "operationId": "onMessage",
                "summary": "",
                "description": "A data message.",
                "message": {
                    "name": "lower_triangular",
                    "title": "Lower Triangular",
                    "description": "",
                    "contentType": "application/json",
                    "payload": {
                        "type": "object",
                        "required": ["payload"],
                        "properties": {
                            "payload": {
                                "$ref": "#/components/schemas/matrix"
                            }
                        },
                        "additionalProperties": false
                    }
                }
            },
            "bindings": {
                "ws": {
                    "method": "GET",
                    "query": {
                        "type": "object",
                        "properties": {
                            "access_token": {
                                "description": "",
                                "type": "string"
                            }
                        },
                        "required": ["access_token"]
                    }
                }
            }
        }
    },
    "components":{
        "schemas": {
            "matrix": {
               "type": "array",
               "description": "A lower right triangular matrix.",
               "example": [[11.0], [21.0,22.0], [31.0,32.0,33.0], [41.0,42.0,43.0,44.0], [51.0,52.0,53.0,54.0,55.0], [61.0,62.0,63.0,64.0,65.0,66.0] ],
               "items": [
                   {
[...]

PR #443, opened 2020-11-09
docs: extend dummy example with array that has array in items

derberg:
**Description**

- extend dummy example with array that has array in items

See also https://github.com/asyncapi/generator/issues/441


jonaslagoni (approved):
👍

PR #445, opened 2020-11-12
chore(release): 1.0.0 release :rocket:

derberg:
**Description**

Tested it with an interactive tutorial on katacoda. Tested locally with latest template for templates. All looks good.

- bump 1.0.0
- latest parser 1.1.0

🚀 🍻 🥇

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@asyncapi/generator",
-  "version": "1.0.0-rc.15",
+  "version": "1.0.0",
   "description": "The AsyncAPI generator. It can generate documentation, code, anything!",
   "main": "./lib/generator.js",
   "bin": {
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -49,10 +49,9 @@
   "dependencies": {
     "@asyncapi/avro-schema-parser": "^0.2.0",
     "@asyncapi/openapi-schema-parser": "^2.0.0",
-    "@asyncapi/parser": "v1.0.1",
+    "@asyncapi/parser": "1.1.0",
```
fmvilas, line 52: Shouldn't we do this in a separate release before? To make sure it works.

derberg: yeah, I thought about this, but I don't really think it is needed, 1.1.0 just added one additional validation of the AsyncAPI files, the query params in channels. This should not have any effect on the generator itself, even if there are bugs, they are not breaking the compatibility of the generator

fmvilas (approved):

fmvilas:
🚀  Ship it!

PR #446, opened 2020-11-13
fix: use parser 1.1.1 that fixes channel names validation in the parser

derberg:
**Description**

- use parser 1.1.1 that fixes channel names validation in the parser

See also https://github.com/asyncapi/parser-js/pull/196


WaleedAshraf (approved):
🍰

PR #448, opened 2020-11-13
docs: add reference to template for generator templates

derberg:
**Description**

- add reference to template for generator templates

### DIFF
--- a/README.md
+++ b/README.md
@@ -268,7 +268,10 @@ Before using newer versions of the template, always look at the [changelog](http
 
 ## How to create a template
 
-To create your own template, for example code generator for some specific language and technology, read [authoring templates](docs/authoring.md) and the [list of templates recipes](docs/templates-recipes.md).
+To create your own template, for example code generator for some specific language and technology, learn from the following resources:
+- Documentation about [authoring templates](docs/authoring.md),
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -268,7 +268,10 @@ Before using newer versions of the template, always look at the [changelog](http
 
 ## How to create a template
 
-To create your own template, for example code generator for some specific language and technology, read [authoring templates](docs/authoring.md) and the [list of templates recipes](docs/templates-recipes.md).
+To create your own template, for example code generator for some specific language and technology, learn from the following resources:
+- Documentation about [authoring templates](docs/authoring.md),
+- Use [Template for Generator Templates](https://github.com/asyncapi/template-for-generator-templates) that showcases Generator features,
+- Check the [list of templates recipes](docs/templates-recipes.md)
```
bszwarc, line 274: To make the list consistent, you have two solutions:

*  Each line starts with a verb, so `Read the documentation...`. `Use Template..` or
* remove the verbs and just leave `Documentation`, `Template for...`, `List of template recipes..` 

Because the preceding line ends with `resources` I would rather  go with the second solution because you list them as nouns. 
Also you can get rid of the commas at the end :)

bszwarc (commented):
A small comment :)

bszwarc (approved):

PR #452, opened 2020-11-25
chore(ci): create PR when receiving release event from parser-js

fmvilas:
**Description**

Creates a PR updating package.json and package-lock.json whenever Github Actions receives a release event from `asyncapi/parser-js`.

It also changes the release workflow a little bit. Instead of automerging PRs created by our bots it will only merge PRs with label `automerge`.

**Related issue(s)**
This is the counterpart on the JS Parser: https://github.com/asyncapi/parser-js/pull/205.

### DIFF
--- a/.github/workflows/automerge.yml
+++ b/.github/workflows/automerge.yml
@@ -36,11 +36,9 @@ jobs:
     steps:
       - name: Automerging
         uses: pascalgn/automerge-action@v0.7.5
-        if: github.actor == 'asyncapi-bot' || github.actor == 'dependabot[bot]' || github.actor == 'dependabot-preview[bot]'
         env:
           GITHUB_TOKEN: "${{ secrets.GH_TOKEN }}"
           GITHUB_LOGIN: asyncapi-bot
-          MERGE_LABELS: ""
           MERGE_METHOD: "squash"
           MERGE_COMMIT_MESSAGE: "pull-request-title"
           MERGE_RETRIES: "10"
--- a/.github/workflows/update-deps.yml
+++ b/.github/workflows/update-deps.yml
@@ -1,0 +1,32 @@
+name: Update dependencies
+
+on:
+  repository_dispatch:
+    types: [asyncapi/parser-js/release]
+
+jobs:
+  update:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout repo
+        uses: actions/checkout@v2
+
+      - name: Setup Node.js
+        uses: actions/setup-node@v1
+        with:
+          node-version: 14
+
+      - name: Update dependency version
+        run: npm i ${{ github.event.client_payload.name }}@${{ github.event.client_payload.version }}
+
+      - name: Create Pull Request with updated package files
+        uses: peter-evans/create-pull-request@v3
+        with:
+          base: master
+          token: ${{ secrets.GH_TOKEN }}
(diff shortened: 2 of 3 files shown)


fmvilas:
I think CI is failing because there's no GH_TOKEN in my fork. @derberg do you know if that's the case?

.github/workflows/automerge.yml
```diff
@@ -36,11 +36,9 @@ jobs:
     steps:
       - name: Automerging
         uses: pascalgn/automerge-action@v0.7.5
-        if: github.actor == 'asyncapi-bot' || github.actor == 'dependabot[bot]' || github.actor == 'dependabot-preview[bot]'
         env:
           GITHUB_TOKEN: "${{ secrets.GH_TOKEN }}"
           GITHUB_LOGIN: asyncapi-bot
-          MERGE_LABELS: ""
```
derberg, line 43: why removed? docs say `When an empty string ("") is given, all pull requests will be merged.`
shouldn't you do `MERGE_LABELS: automerge`?

anyway I don't like the change as now we have issue with dependabot again and also the action I just finished yesteday is not supporting this approach. Why is this change needed anyway? why can't we merge automatically once `Update dependencies` is successfully completed?

.github/workflows/update-deps.yml
```diff
@@ -0,0 +1,32 @@
...
+        uses: actions/checkout@v2
+
+      - name: Setup Node.js
+        uses: actions/setup-node@v1
+        with:
+          node-version: 14
+
+      - name: Update dependency version
+        run: npm i ${{ github.event.client_payload.name }}@${{ github.event.client_payload.version }}
+
```
derberg, line 21: we should run tests here, or make sure https://github.com/asyncapi/generator/blob/master/.github/workflows/pull-request-testing.yml is mandatory and if fails, automerge should not be performed....but actually we need both, as what is the point to create a PR anyway if tests fail

.github/workflows/update-deps.yml
```diff
@@ -0,0 +1,32 @@
+name: Update dependencies
```
derberg, line 1: maybe extend the name and file name to clarify that it is about "internal" dependencies?

derberg (changes requested):

.github/workflows/automerge.yml
fmvilas: We can't because it will also merge breaking changes automatically. For instance, if it receives an update to a major version it will get updated on the package.json but the API probably changes and therefore it will fail. 🤔 but now that I think about it, tests would fail 🤔

.github/workflows/update-deps.yml
fmvilas: Yeah, tests should pass. Is there an easy way to do this? I'm confused right now.

.github/workflows/automerge.yml
derberg: exactly, they should fail, and if didn't, not a breaking change, 
parser 2.0 might not require to release generator 2.0 (in theory, it will anyway fail because of tests)

.github/workflows/update-deps.yml
derberg: just another step to run `npm test`, but I think we can survive with a PR that is created that later it cannot be merged because of failing tests, most important is to set in settings in branch protection requirement to make PR testing a mandatory check

derberg (changes requested):
@fmvilas tbh I'm confused why it is failing, PR runs in scope of upstream, so token should be fine 🤔

PR #453, opened 2020-11-26
feat: integrate react SDK for rendering react templates

jonaslagoni:
**Description**
Introduces the [generator React SDK](https://github.com/asyncapi/generator-react-sdk) which enables the rendering of React templates. @magicmatatjahu have provided a rewrite of the [markdown-template](https://github.com/asyncapi/markdown-template/pull/43) to react which can be tried with this PR.

What this PR introduces:

- New validation rule so only `react` or `nunjucks` are allowed as renderer configurations (or none).
- Updated `authoring.md` to include the new react renderer.
- Moved any specific nunjucks code into separate utility functions and made a common function for each renderer. This introduced the functions `renderFile`, `configureTemplate` and `renderAndWriteToFile` which each call either react or nunjucks specific code.

**TODO before merge**
- [x] Release the [generator React SDK](https://github.com/asyncapi/generator-react-sdk) and update the dependency from file to library.
- [x] Add some more documentation regarding VSCode debugging.
- [x] Figure out if we can remove the requirement of `source-map-support` and `react`. 

**TODO after merge**

- The `generator.js` is way to coupled and needs to be broken down. Just look at how problematic it is to separate react and nunjucks code. 

**Related issue(s)**
https://github.com/asyncapi/shape-up-process/issues/35


lib/generator.js
```diff
@@ -548,19 +549,54 @@ class Generator {
...
+      let filePath = outputpath;
+      let permission = 0o666;
+      const content = renderContent.content;
+      if (content === undefined || typeof content !== 'string') throw Error(`Template file not rendered correctly, was rendered as ${content}`);
+      if (renderContent.metadata !== undefined) {
+        const newFileName = renderContent.metadata.name;
+        if (renderContent.metadata.permission !== undefined) {
+          permission = renderContent.metadata.permission;
+        }
+        if (newFileName && newFileName.trim() !== '') {
```
fmvilas, line 588: I'd do `trim()` above, on line 584, so the name is clean since the beginning.

lib/generator.js
```diff
@@ -548,19 +549,54 @@ class Generator {
...
+  async renderAndWriteToFile(asyncapiDocument, templateFilePath, outputpath, extraTemplateData) {
+    const renderContent =  await this.renderFile(asyncapiDocument, templateFilePath, extraTemplateData);
+    console.log(renderContent);
+    if (this.templateConfig && this.templateConfig.type === 'react') {
+      let filePath = outputpath;
+      let permission = 0o666;
+      const content = renderContent.content;
+      if (content === undefined || typeof content !== 'string') throw Error(`Template file not rendered correctly, was rendered as ${content}`);
+      if (renderContent.metadata !== undefined) {
+        const newFileName = renderContent.metadata.name;
```
fmvilas, line 584: What about:

```suggestion
        const newFileName = renderContent.metadata.fileName;
```

Just to be a bit more precise.

lib/generator.js
```diff
@@ -548,19 +549,54 @@ class Generator {
...
+   * @param {String} templateFilePath Path to the file being rendered.
+   * @param {String} outputpath Path to the file being rendered.
+   * @param {Object} [extraTemplateData] Extra data to pass to the template.
+   */
+  async renderAndWriteToFile(asyncapiDocument, templateFilePath, outputpath, extraTemplateData) {
+    const renderContent =  await this.renderFile(asyncapiDocument, templateFilePath, extraTemplateData);
+    console.log(renderContent);
+    if (this.templateConfig && this.templateConfig.type === 'react') {
+      let filePath = outputpath;
+      let permission = 0o666;
```
fmvilas, line 580: Should it be `permissions` instead of `permission`?

lib/generator.js
```diff
@@ -627,18 +661,17 @@ class Generator {
...
    *
    * @private
    * @param {AsyncAPIDocument} asyncapiDocument AsyncAPI document to pass to the template.
-   * @param {String} templateString String containing the template.
    * @param {String} filePath Path to the file being rendered.
    * @param {Object} [extraTemplateData] Extra data to pass to the template.
    * @return {Promise}
    */
-  renderString(asyncapiDocument, templateString, filePath, extraTemplateData = {}) {
+  renderNunjucks(asyncapiDocument, tempalteString, filePath, extraTemplateData = {}) {
```
fmvilas, line 672:
```suggestion
  renderNunjucks(asyncapiDocument, templateString, filePath, extraTemplateData = {}) {
```

lib/generator.js
```diff
@@ -627,18 +661,17 @@ class Generator {
...
     return new Promise((resolve, reject) => {
-      this.nunjucks.renderString(templateString, {
+      this.nunjucks.renderString(tempalteString, {
```
fmvilas, line 674:
```suggestion
      this.nunjucks.renderString(templateString, {
```

lib/generator.js
```diff
@@ -657,11 +698,14 @@ class Generator {
    * @param {AsyncAPIDocument} asyncapiDocument AsyncAPI document to pass to the template.
    * @param {String} filePath Path to the file you want to render.
    * @param {Object} [extraTemplateData] Extra data to pass to the template.
-   * @return {Promise}
+   * @return {Promise<string>}
    */
   async renderFile(asyncapiDocument, filePath, extraTemplateData = {}) {
-    const content = await readFile(filePath, 'utf8');
-    return this.renderString(asyncapiDocument, content, filePath, extraTemplateData);
+    const tempalteString = await readFile(filePath, 'utf8');
```
fmvilas, line 704:
```suggestion
    const templateString = await readFile(filePath, 'utf8');
```

lib/generator.js
```diff
@@ -657,11 +698,14 @@ class Generator {
...
+    if (this.templateConfig && this.templateConfig.type === 'react') {
+      return await this.renderReact(asyncapiDocument, filePath, extraTemplateData);
+    }
+    return this.renderNunjucks(asyncapiDocument, tempalteString, filePath, extraTemplateData);
```
fmvilas, line 708:
```suggestion
    return this.renderNunjucks(asyncapiDocument, templateString, filePath, extraTemplateData);
```

lib/generator.js
```diff
@@ -548,19 +549,54 @@ class Generator {
...
     const temp = {};
     const key = template === 'everySchema' || template === 'objectSchema' ? 'schema' : template;
     temp[`${key}Name`] = name;
     temp[key] = component;
-    const content = await this.renderFile(asyncapiDocument, path.resolve(baseDir, fileName), temp);
+    await this.renderAndWriteToFile(asyncapiDocument, path.resolve(baseDir, fileName), targetFile, temp);
+  }
+
+  /**
+   * Render and write to file
```
fmvilas, line 567:
```suggestion
   * Renders a template and writes the result into a file.
```

lib/generator.js
jonaslagoni: changed

lib/generator.js
jonaslagoni: Should be for an individual file not multiple, so 1 permission right? :thinking:

jonaslagoni: never mind you are right.

lib/generator.js
jonaslagoni: :+1:

docs/authoring.md
```diff
@@ -259,3 +225,85 @@ There are some template parameters that have a special meaning:
...
+* It enables the possiblity of [debugging](#debugging-react-template) your template (this is not possible with Nunjucks).
+* It provides better stack traces once something is not working.
+* Provides better support for separating code into more manageable chunks.
+* The readability of the template are much better then Nunjucks syntax.
+* Better tool support for development.
+
+When writing react templates you decide whether to use CommonJS, ES5 or ES6 modules since everything is bundles together before the rendering process takes over. We use our own React renderer which can be found in the [Generator React SDK](https://github.com/asyncapi/generator-react-sdk). 
+There you can find information about how the renderer works or how we transpile your template files.
+
+Your React template always require 2 dependencies `react` and `@asyncapi/generator-react-sdk`. The `react` dependency is pretty self explanitory, it allows us to render the template. `@asyncapi/generator-react-sdk` is required since react files does not export default function where the root component is not of type `File` will not be rendered, furthermore it provide some support components to easy your development.
```
magicmatatjahu, line 240:
```md
...The `react` dependency is pretty self explanitory, it allows us to render the template...
```

The react is required because before transpilation you have `<Component {...props} >` but after transpilation you have `React.createElement(Component, props)` so without `import React from 'react'` then your code is broken (React will be undefined). It's a magic but it's necessary :) Also react doesn't render our template - it does by our custom renderer in `@asyncapi/generator-react-sdk` package by exported `render` method.

I would rather to change whole paragraph to something like:

```md
Your React template always require 2 dependencies `react` and `@asyncapi/generator-react-sdk`. The `react` dependency is required to declare React components inside template. `@asyncapi/generator-react-sdk` is required to import and use as a root component of your template the `File` component - it provides required metadata and without it your template won't be rendered. Furthermore it provides some common components to make your development easier, like `Text` or `Indent`
```

docs/authoring.md
```diff
@@ -259,3 +225,85 @@ There are some template parameters that have a special meaning:
...
+import React from "react";
+import { File, Text } from "@asyncapi/generator-react-sdk";
+export default function({ asyncapi, params, originalAsyncAPI}) {
+  return (
+    <File name={'asyncapi.md'}>
+      <Text>Some text that should render as is</Text>
+    </File>
+  );
+}
+```
```
magicmatatjahu, line 253: Maybe we should change this example to this https://github.com/asyncapi/generator-react-sdk/blob/4ed0acffc9c674cdf2f3a4d67feac4a5fb53ca3b/README.md#example to show how to using the custom components? 😄

docs/authoring.md
```diff
@@ -259,3 +225,85 @@ There are some template parameters that have a special meaning:
...
+The react component returns a `File` component as root component which the React renderer uses to figure out what what file should be generated. In our case we overwrite the default functionality of saving the file as `MyTemplate.js` but instead use the filename `asyncapi.md`. It is then specified that we should render `Some text that should render as is` within that file.
```
magicmatatjahu, line 254:
```md
The exported by default function returns a `File` component as a root component which the [Generator](https://github.com/asyncapi/generator) uses to figure out what file should be generated. In our case we overwrite the default functionality of saving the file as `MyTemplate.js` but instead use the filename `asyncapi.md`. It is then specified that we should render `Some text that should render as is\n` within that file.
```

New line is very important 😄 and also I changed that Generator figure out metadata not renderer. Renderer only travserses components hierarchy and changes components to string.

magicmatatjahu (changes requested):
Very good! Please refer to my comments :) Additionally, I would like to ask you to add some note that React is recommended, e.g. at the beginning of the document and don't writing about it in several places :)

docs/authoring.md
magicmatatjahu: I mean change the content of `File` component :)

docs/authoring.md
```diff
@@ -259,3 +225,85 @@ There are some template parameters that have a special meaning:
 |Name|Description|
 |---|---|
 |`server`| It is used to let the template know which server from the AsyncAPI specification file you want to use. In some cases, this may be required. For instance, when generating code that connects to a specific server. Use this parameter in case your template relies on users' information about what server from the specification file they want to use during generation. You also need this parameter if you want to use `server.protocol` notation within `conditionalFiles` configuration option. Once you decide to specify this parameter for your template, it is recommended you make it a mandatory parameter otherwise feature like `conditionalFiles` is not going to work if your users do not use this parameter obligatory.
+
+
+## React
+[React](https://reactjs.org) is the newest render engine which we strongly suggest you should use for any new templates. It is only the default render engine to enable backward compatibility.
```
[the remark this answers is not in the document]
jonaslagoni: Yea sounds better. Last section was to describe why it was not the default render engine but yea was missing some words, changed it 😄

docs/authoring.md
```diff
@@ -259,3 +225,85 @@ There are some template parameters that have a special meaning:
...
+      <Text>Some text that should render as is</Text>
+    </File>
+  );
+}
+```
+The react component returns a `File` component as root component which the React renderer uses to figure out what what file should be generated. In our case we overwrite the default functionality of saving the file as `MyTemplate.js` but instead use the filename `asyncapi.md`. It is then specified that we should render `Some text that should render as is` within that file.
+
+For further information about components, props etc. see the [Generator React SDK](https://github.com/asyncapi/generator-react-sdk)
+
+### Debugging React template
```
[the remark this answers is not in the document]
jonaslagoni: Hmm, yea not sure, I like when libraries make it as easy as possible. This ensures that they dont need to figure out how to setup their own actions, but I see your point..

docs/authoring.md
```diff
@@ -259,3 +225,85 @@ There are some template parameters that have a special meaning:
...
+
+
+## React
+[React](https://reactjs.org) is the newest render engine which we strongly suggest you should use for any new templates. It is only the default render engine to enable backward compatibility.
+* It enables the possiblity of [debugging](#debugging-react-template) your template (this is not possible with Nunjucks).
+* It provides better stack traces once something is not working.
+* Provides better support for separating code into more manageable chunks.
+* The readability of the template are much better then Nunjucks syntax.
```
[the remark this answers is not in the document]
jonaslagoni: I think we should keep it because it outlines the issues that is inherit by Nunjucks that they will encounter if they go for that instead. People can make their own decision, but I think it is a good idea to have it 😄

docs/authoring.md
jonaslagoni: Yea you are right, abstracted away from that a bit. Try see the updated readme 😄

docs/authoring.md
jonaslagoni: hmm, I wanted to keep it as simple as possible, also why in the end I refer to the SDK for further information about props, components etc 🤔

docs/authoring.md
jonaslagoni: good point, added.

docs/authoring.md
```diff
@@ -251,7 +251,7 @@ export default function({ asyncapi, params, originalAsyncAPI}) {
   );
 }
 ```
-The react component returns a `File` component as root component which the React renderer uses to figure out what what file should be generated. In our case we overwrite the default functionality of saving the file as `MyTemplate.js` but instead use the filename `asyncapi.md`. It is then specified that we should render `Some text that should render as is` within that file.
+The exported default function returns a `File` component as a root component which the [Generator](https://github.com/asyncapi/generator) uses to figure out what file should be generated. In our case we overwrite the default functionality of saving the file as `MyTemplate.js` but instead use the filename `asyncapi.md`. It is then specified that we should render `Some text that should render as is\n` within that file. Notice the `\n` character at the end, this is something that is automatically added after each children in the `File` and `Indent` component.
```
magicmatatjahu, line 254:
```md
Notice the `\n` character at the end, this is something that is automatically added after each children in the `File` and `Indent` component. 
```

It's added by each `Text` component, not after each children 😄 https://github.com/asyncapi/generator-react-sdk/blob/1e897fc2d2d6b42fb2513f067d88bb5e9b8f2c88/src/components/Text.tsx#L50

docs/authoring.md
```diff
@@ -259,3 +224,112 @@ There are some template parameters that have a special meaning:
...
+}
+```
+The exported default function returns a `File` component as a root component which the [Generator](https://github.com/asyncapi/generator) uses to figure out what file should be generated. In our case we overwrite the default functionality of saving the file as `MyTemplate.js` but instead use the filename `asyncapi.md`. It is then specified that we should render `Some text that should render as is\n` within that file. Notice the `\n` character at the end, this is something that is automatically added after each children in the `File` and `Indent` component. 
+
+For further information about components, props etc. see the [Generator React SDK](https://github.com/asyncapi/generator-react-sdk)
+
+### Common assumptions
+
+1. All files within the template directory are rendered if their root component is of type `File` and the path is not part of the `nonRenderableFiles` configuration.
+1. New lines are automatically added for each child in `File` and `Indent` component.
```
magicmatatjahu, line 261: `New lines are automatically added for each child in `File` and `Indent` component.` is wrong.

New lines are only added in `Text` component. File is w wrapper and Indent adds only indentations (without new lines) :) https://github.com/asyncapi/generator-react-sdk/blob/1e897fc2d2d6b42fb2513f067d88bb5e9b8f2c88/src/utils/withIndendation.ts

docs/authoring.md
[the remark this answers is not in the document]
magicmatatjahu: Ok, I understand, but maybe we should add subheaders for appropriate IDEs? :) `### Debugging React template` then `#### VSC` etc.. What do you think?

docs/authoring.md
```diff
@@ -10,17 +10,10 @@ To work on a template, you need an AsyncAPI specification file that you can use
...
-1. Templates may contain `hooks` that are functions invoked during specific moment of the generation. In the template, they must be stored in the `.hooks` directory under the template directory. They can also be stored in other modules and external libraries and configured inside the template [Read more about hooks](#hooks).
-1. Templates may contain `partials` (reusable chunks). They must be stored in the `.partials` directory under the template directory. [Read more about partials](#partials).
 1. Templates can be configured. Configuration must be stored in the `package.json` file under custom `generator` property. [Read more about the configuration file](#configuration-file).
+1. The template engine is either [Nunjucks](#nunjucks)(default) or [React](#react). This can be controlled with the `renderer` configuration.
+1. Templates may contain `hooks` that are functions invoked during specific moment of the generation. In the template, they must be stored in the `.hooks` directory under the template directory. They can also be stored in other modules and external libraries and configured inside the template [Read more about hooks](#hooks).
```
magicmatatjahu, line 15:
```suggestion
1. Templates may contain `hooks` that are functions invoked during specific moment of the generation. In the template, they must be stored in the `hooks` directory under the template directory. They can also be stored in other modules and external libraries and configured inside the template [Read more about hooks](#hooks).
```

docs/authoring.md
```diff
@@ -10,17 +10,10 @@ To work on a template, you need an AsyncAPI specification file that you can use
...
-1. Templates may contain multiple files. Unless stated otherwise, all files will be rendered.
-1. The template engine is [Nunjucks](https://mozilla.github.io/nunjucks).
-1. Templates may contain [Nunjucks filters or helper functions](https://mozilla.github.io/nunjucks/templating.html#builtin-filters). [Read more about filters](#filters).
-1. Templates may contain `hooks` that are functions invoked during specific moment of the generation. In the template, they must be stored in the `.hooks` directory under the template directory. They can also be stored in other modules and external libraries and configured inside the template [Read more about hooks](#hooks).
-1. Templates may contain `partials` (reusable chunks). They must be stored in the `.partials` directory under the template directory. [Read more about partials](#partials).
 1. Templates can be configured. Configuration must be stored in the `package.json` file under custom `generator` property. [Read more about the configuration file](#configuration-file).
+1. The template engine is either [Nunjucks](#nunjucks)(default) or [React](#react). This can be controlled with the `renderer` configuration.
```
fmvilas, line 14:
```suggestion
1. The template engine can be either [React](#react) (recommended) or [Nunjucks](#nunjucks) (default). This can be controlled with the `renderer` property in the [template configuration](#template-configuration).
```

docs/authoring.md
```diff
@@ -169,31 +150,14 @@ module.exports = {
...
-{{ tags(operation.tags()) }}
-```
 
 ## Configuration File
 
 The `generator` property from `package.json` file must contain a JSON object that may have the following information:
 
 |Name|Type|Description|
 |---|---|---|
+|`renderer`| String | Supports either `react` or `nunjucks`(default).
```
fmvilas, line 160:
```suggestion
|`renderer`| String | Its value can be either `react` or `nunjucks` (default).
```

docs/authoring.md
```diff
@@ -259,3 +224,112 @@ There are some template parameters that have a special meaning:
 |Name|Description|
 |---|---|
 |`server`| It is used to let the template know which server from the AsyncAPI specification file you want to use. In some cases, this may be required. For instance, when generating code that connects to a specific server. Use this parameter in case your template relies on users' information about what server from the specification file they want to use during generation. You also need this parameter if you want to use `server.protocol` notation within `conditionalFiles` configuration option. Once you decide to specify this parameter for your template, it is recommended you make it a mandatory parameter otherwise feature like `conditionalFiles` is not going to work if your users do not use this parameter obligatory.
+
+
+## React
+[React](https://reactjs.org) is the render engine which we strongly suggest you should use for any new templates. The only reason it is not the default render engine is to stay backward compatibility.
```
fmvilas, line 230:
```suggestion
[React](https://reactjs.org) is the render engine that we strongly suggest you should use for any new templates. The only reason it is not the default render engine is to stay backward compatible.
```

docs/authoring.md
```diff
@@ -259,3 +224,112 @@ There are some template parameters that have a special meaning:
...
+* It enables the possiblity of [debugging](#debugging-react-template) your template (this is not possible with Nunjucks).
+* It provides better error stack traces once something is not working.
```
fmvilas, line 232:
```suggestion
* It provides better error stack traces.
```

I think it's redundant to say "once something is not working".

docs/authoring.md
```diff
@@ -259,3 +224,112 @@ There are some template parameters that have a special meaning:
...
+* Provides better support for separating code into more manageable chunks/components.
+* The readability of the template are much better then Nunjucks syntax.
```
fmvilas, line 234:
```suggestion
* The readability of the template is much better compared to Nunjucks syntax.
```

docs/authoring.md
```diff
@@ -259,3 +224,112 @@ There are some template parameters that have a special meaning:
...
+* Better tool support for development.
+
+When writing react templates you decide whether to use CommonJS, ES5 or ES6 modules since everything is bundles together before the rendering process takes over. We use our own React renderer which can be found in the [Generator React SDK](https://github.com/asyncapi/generator-react-sdk).
```
fmvilas, line 237:
```suggestion
When writing React templates you decide whether to use CommonJS, ES5, or ES6 modules since everything is bundled together before the rendering process takes over. We use our own React renderer which can be found in the [Generator React SDK](https://github.com/asyncapi/generator-react-sdk). 
```

docs/authoring.md
```diff
@@ -259,3 +224,112 @@ There are some template parameters that have a special meaning:
...
+* Provides better support for separating code into more manageable chunks/components.
+* The readability of the template are much better then Nunjucks syntax.
+* Better tool support for development.
+
+When writing react templates you decide whether to use CommonJS, ES5 or ES6 modules since everything is bundles together before the rendering process takes over. We use our own React renderer which can be found in the [Generator React SDK](https://github.com/asyncapi/generator-react-sdk). 
+There you can find information about how the renderer works or how we transpile your template files.
+
+Your React template always require 3 dependencies `react`, `source-map-support` and `@asyncapi/generator-react-sdk`. The `react` dependency is required to declare React components inside the template. `source-map-support` ensure that the transpiled files are mapped to the actual template files, this is only to enable correct error stack trace. `@asyncapi/generator-react-sdk` is required to access the `File` component which is required as a root component for a file to be rendered. Furthermore it provides some common components to make your development easier, like `Text` or `Indent`.
```
fmvilas, line 240:
```suggestion
Your React template always requires 3 dependencies `react`, `source-map-support`, and `@asyncapi/generator-react-sdk`. The `react` dependency is required to declare React components inside the template. `source-map-support` ensures that the transpiled files are mapped to the actual template files, this is only to enable correct error stack trace. `@asyncapi/generator-react-sdk` is required to access the `File` component which is required as a root component for a file to be rendered. Furthermore, it provides some common components to make your development easier, like `Text` or `Indent`.
```

> The `react` dependency is required to declare React components inside the template.

Is it? Can't we inject it somehow with Rollup? For instance, when using Next.js it's not necessary. I wonder how they do it.

> `source-map-support` ensures that the transpiled files are mapped to the actual template files, this is only to enable correct error stack trace.

Can't we just include it as part of the `generator-react-sdk`? Just so people don't have to do it.

docs/authoring.md
```diff
@@ -259,3 +224,112 @@ There are some template parameters that have a special meaning:
...
+
+Lets consider a basic react template as the one below called `MyTemplate.js`:
```
fmvilas, line 242:
```suggestion
Let's consider a basic React template as the one below called `MyTemplate.js`:
```

docs/authoring.md
```diff
@@ -259,3 +224,112 @@ There are some template parameters that have a special meaning:
...
+```js
+import React from "react";
+import { File, Text } from "@asyncapi/generator-react-sdk";
+export default function({ asyncapi, params, originalAsyncAPI}) {
+  return (
+    <File name={'asyncapi.md'}>
```
fmvilas, line 248:
```suggestion
    <File name="asyncapi.md">
```

docs/authoring.md
```diff
@@ -259,3 +224,112 @@ There are some template parameters that have a special meaning:
...
+There you can find information about how the renderer works or how we transpile your template files.
+
+Your React template always require 3 dependencies `react`, `source-map-support` and `@asyncapi/generator-react-sdk`. The `react` dependency is required to declare React components inside the template. `source-map-support` ensure that the transpiled files are mapped to the actual template files, this is only to enable correct error stack trace. `@asyncapi/generator-react-sdk` is required to access the `File` component which is required as a root component for a file to be rendered. Furthermore it provides some common components to make your development easier, like `Text` or `Indent`.
+
+Lets consider a basic react template as the one below called `MyTemplate.js`:
+```js
+import React from "react";
+import { File, Text } from "@asyncapi/generator-react-sdk";
+export default function({ asyncapi, params, originalAsyncAPI}) {
```
fmvilas, line 246: > export default function({ asyncapi, params, originalAsyncAPI}) {

I wonder if exporting anonymous functions is the recommended pattern with React. Also, wouldn't it make things harder to debug?

docs/authoring.md
```diff
@@ -259,3 +224,112 @@ There are some template parameters that have a special meaning:
...
+
+For further information about components, props etc. see the [Generator React SDK](https://github.com/asyncapi/generator-react-sdk)
+
+### Common assumptions
+
+1. All files within the template directory are rendered if their root component is of type `File` and the path is not part of the `nonRenderableFiles` configuration.
+1. New lines are automatically added after each `Text` component.
+
+### Debugging React template
+With react it enables you to debug your templates. Below are a section of common idé's and how to enable debugging.
```
fmvilas, line 264:
```suggestion
With React, it enables you to debug your templates. Below is a section of common IDE's and how to enable debugging. 
```

docs/authoring.md
```diff
@@ -259,3 +224,112 @@ There are some template parameters that have a special meaning:
...
+### Common assumptions
+
+1. All files within the template directory are rendered if their root component is of type `File` and the path is not part of the `nonRenderableFiles` configuration.
+1. New lines are automatically added after each `Text` component.
+
+### Debugging React template
+With react it enables you to debug your templates. Below are a section of common idé's and how to enable debugging. 
+
+#### VSCode
+Add the following launch configuration in your workspace:
```
fmvilas, line 267: Where? I'm using VS Code and honestly have no idea what you mean here 😅

docs/authoring.md
```diff
@@ -259,3 +224,112 @@ There are some template parameters that have a special meaning:
...
+      ],
+      "program": "ag"
+    }
+  ]
+}
+```
+Now replace `./asyncapi.yml` with your document of choice. Replace `./template` with the path to your React template. You can now debug your template by adding any breakpoints you want and inspect your code.
+
+## Nunjucks
+[Nunjucks](https://mozilla.github.io/nunjucks) is the default render engine, however we strongly suggest you to consider adapting to the [React](#react) engine.
```
fmvilas, line 294:
```suggestion
[Nunjucks](https://mozilla.github.io/nunjucks) is the default render engine, however, we strongly recommend adopting the [React](#react) engine.
```

fmvilas (changes requested):
Left some comments on the `authoring.md` docs. Great stuff so far!

docs/authoring.md
jonaslagoni: 🤔 It would be optimal, I'll try and see what I can do.

docs/authoring.md
jonaslagoni: Good question, @magicmatatjahu I think we need your input here 😄

docs/authoring.md
jonaslagoni: Fair point, I'll add some more documentation here 😄

docs/authoring.md
magicmatatjahu: At one time, default exports were recommended by the community later that changed 😄 

Return `default` function is crucial for DX and in which way works at this moment the Generator. Why? Because by default export you don't have to name function by `SchemaTemplate` or another predefined name for Generator. We just import the file and then perform some actions to render the template to file. If we enable export by names, then we have a issue: which function we should process? Let's check the example:

```js
export function SomeTemplate({}) {
  // ...implementation
}

export function SomeTemplate2({}) {
  // ...implementation
}
```

which we should choose? As I remember, we didn't discuss on the possibility of rendering several different templates in one file, the difference may be the rendering of an array of templates, which differs only on input data (yes, I am a hypocrite hah 😄).

fmvilas: I don't mean removing `default` but instead adding a function name:

```suggestion
export default function Example({ asyncapi, params, originalAsyncAPI}) {
```

magicmatatjahu: Ahhh, ok. Mea culpa. It's should recommended, so we should change the example, but it's not mandatory.

docs/authoring.md
jonaslagoni: @fmvilas added some more details and link to launch configuration. Do you feel like more details are needed? 🤔

docs/authoring.md
jonaslagoni: I have added all the dependencies to the `generator-react-sdk` library. Everything works fine when its used as a normal dependency. However because we in development use a local dependency we are required to add these dependencies our self because npm installs them locally instead of "Merging" them with the template dependencies. 

Not sure how we should communicate this... Maybe I'll add it to the SDK readme? 🤔

fmvilas: Not sure I follow 😅 What do you mean by "merging"? Why is it not working in development?

jonaslagoni: At the moment in the templates we referer to the SDK with the following: `"@asyncapi/generator-react-sdk": "file:../generator-react-sdk"` i.e. a local dependency. However the way NPM treats this dependency is different from `"@asyncapi/generator-react-sdk": "0.1.0"`.

With a local dependency it installs IT's node_modules inside `../generator-react-sdk/node_modules` i.e. those dependencies are not installed inside `template/node_modules` as it normally would. This means that when the SDK has `source-map-support` dependency the template does not have access to this dependency if we use local dependencies.

Did that make sence in any way? 😅

docs/authoring.md
jonaslagoni: @magicmatatjahu can we mark this as resolved?

docs/authoring.md
fmvilas: Oh! I see what you mean. Why is the `generator-react-sdk` a local dependency? Isn't it going to be another published package? I mean, can't we just replace `file:../generator-react-sdk` with `"@asyncapi/generator-react-sdk": "0.1.0"` and it would work?

If you want to work with a local dependency yourself while developing the `generator-react-sdk` package you can use [npm link](https://docs.npmjs.com/cli/v6/commands/npm-link).

jonaslagoni: Yes thats what we should have done but didnt :D So it should fix itself :)

docs/authoring.md
```diff
@@ -259,3 +224,112 @@ There are some template parameters that have a special meaning:
...
+### Common assumptions
+
+1. All files within the template directory are rendered if their root component is of type `File` and the path is not part of the `nonRenderableFiles` configuration.
+1. New lines are automatically added after each `Text` component.
+
+### Debugging React template
+With React, it enables you to debug your templates. Below is a section of common IDE's and how to enable debugging. 
+
+#### VSCode
+For Visual Studio Code we have created a boiler plate [launch configuration](https://code.visualstudio.com/docs/editor/debugging#_launch-configurations) to enable debug in your template. Add the following launch configuration:
```
fmvilas, line 267:
```suggestion
For Visual Studio Code, we have created a boilerplate [launch configuration](https://code.visualstudio.com/docs/editor/debugging#_launch-configurations) to enable debug in your template. Add the following launch configuration:
```

fmvilas (changes requested):

docs/authoring.md
fmvilas: Looks good now! Thanks. Just added a minor syntax suggestion 👍

lib/generator.js
```diff
@@ -548,19 +564,38 @@ class Generator {
...
+  }
 
-    await writeFile(targetFile, content, 'utf8');
+  /**
+   * Renders a template and writes the result into a file.
+   * 
+   * @private
+   * @param {AsyncAPIDocument} asyncapiDocument AsyncAPI document to pass to the template.
+   * @param {String} templateFilePath Path to the file being rendered.
+   * @param {String} outputpath Path to the file being rendered.
```
fmvilas, line 587: The description of these two is the same. Maybe something like this?

```suggestion
   * @param {String} templateFilePath Path to the input file being rendered.
   * @param {String} outputPath Path to the resulting rendered file.
```

lib/templateConfigValidator.js
```diff
@@ -108,6 +107,19 @@ function isServerProtocolSupported(server, supportedProtocols, paramsServerName)
...
 
+/**
+ * Checks if the the provided renderer are supported (no renderer are also supported, defaults to nunjucks)
+ * 
+ * @param  {Object} templateConfig Template configuration.
+ */
+function isProvidedTemplateRendererSupported(templateConfig) {
+  if (templateConfig.renderer === undefined) return;
+  if (templateConfig.renderer === 'react') return;
+  if (templateConfig.renderer === 'nunjucks') return;
```
fmvilas, line 118:
```suggestion
  if ([undefined, 'react', 'nunjucks'].includes(templateConfig.renderer)) return;
```

lib/utils.js
```diff
@@ -72,6 +72,15 @@ utils.isLocalTemplate = async (templatePath) => {
   return stats.isSymbolicLink();
 };
 
+/**
+ * Returns whether or not the template is a react template
+ * 
+ * @param {object} templateConfig
```
fmvilas, line 78:
```suggestion
 * @param {object} templateConfig 
 * @returns {boolean} Whether the template is a React template or not.
```

fmvilas (changes requested):

package.json
```diff
@@ -52,6 +52,7 @@
     "@asyncapi/parser": "^1.1.1",
     "@asyncapi/raml-dt-schema-parser": "^2.0.0",
     "ajv": "^6.10.2",
+    "asyncAPISDK": "file://./../generator-react-sdk",
```
fmvilas, line 55: I think you can stop using `file://` deps. Use npm link: https://docs.npmjs.com/cli/v6/commands/npm-link.

jonaslagoni: Done 👍

lib/templateConfigValidator.js
jonaslagoni: @magicmatatjahu rewrote it.

lib/generator.js
jonaslagoni: Yea good catch 😅

lib/templateConfigValidator.js
magicmatatjahu: Hah 😄 I didn't notice this comment

lib/renderer/react.js
```diff
@@ -0,0 +1,79 @@
+const path = require('path');
+const AsyncReactSDK = require('asyncAPISDK');
```
magicmatatjahu, line 2: Please update dependency :)

fmvilas (dismissed):
Great stuff my friends 👏

lib/renderer/react.js
jonaslagoni: Ahh, good catch!

jonaslagoni:
@magicmatatjahu didnt you change when the `registerFilters` function was called? Tests are failing 😄

fmvilas (dismissed):

magicmatatjahu (approved):

magicmatatjahu (dismissed):

jonaslagoni:
@fmvilas @derberg I am not allowed to merge 🥳

fmvilas (approved):
Last time I approve and you dismiss my review :trollface:

derberg:
@jonaslagoni @fmvilas you mind wait with merge, I'd like to also review, this is a big change

magicmatatjahu (dismissed):

derberg (changes requested):
eh those if-s and else-s 😄 
yeah, I know that for now there is no other way doing it

I left some minor comments. What I'm missing is a kind of comparison of what options we had in nunjucks and what is their equivalent in react, kind of matrix to refer to for people familiar with nunjucks engine. I guess that this can be done in a separate PR

I tried your PR manually with nodejs template:
- this file https://raw.githubusercontent.com/asyncapi/asyncapi/2.0.0/examples/2.0.0/streetlights.yml
- I get this in rendered config file
```
broker:
    mqtt:
      url: mqtt://83
```
- something ain't right here https://github.com/asyncapi/nodejs-template/blob/master/template/config/common.yml#L30

I'll try to debug later on my own what might be wrong, maybe something on my side 🤷🏼‍♂️

jonaslagoni:
> eh those if-s and else-s 😄
> yeah, I know that for now there is no other way doing it
> 
> I left some minor comments. What I'm missing is a kind of comparison of what options we had in nunjucks and what is their equivalent in react, kind of matrix to refer to for people familiar with nunjucks engine. I guess that this can be done in a separate PR
> 
> I tried your PR manually with nodejs template:
> 
>     * this file https://raw.githubusercontent.com/asyncapi/asyncapi/2.0.0/examples/2.0.0/streetlights.yml
> 
>     * I get this in rendered config file
> 
> 
> ```
> broker:
>     mqtt:
>       url: mqtt://83
> ```
> 
>     * something ain't right here https://github.com/asyncapi/nodejs-template/blob/master/template/config/common.yml#L30
> 
> 
> I'll try to debug later on my own what might be wrong, maybe something on my side 🤷🏼‍♂️

What is the full command you are using? I'll try to replicate it 😄

docs/authoring.md
```diff
@@ -259,3 +224,116 @@ There are some template parameters that have a special meaning:
...
+
+### Common assumptions
+
+1. All files within the template directory are rendered if their root component is of type `File` and the path is not part of the `nonRenderableFiles` configuration.
+1. New lines are automatically added after each `Text` component.
+
+### Debugging React template
+With React, it enables you to debug your templates. Below is a section of common IDE's and how to enable debugging. 
+
+#### VSCode
```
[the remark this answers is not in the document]
jonaslagoni: The thought was that it could be introduced as [suggested](https://github.com/asyncapi/generator/pull/453#discussion_r536196752) by @magicmatatjahu. I am neutral about this, I can understand why it should be changed since we only provide one example.

docs/authoring.md
```diff
@@ -259,3 +224,116 @@ There are some template parameters that have a special meaning:
...
+Now replace `./asyncapi.yml` with your document of choice. Replace `./template` with the path to your React template. You can now debug your template by adding any breakpoints you want and inspect your code.
+
+## Nunjucks
+[Nunjucks](https://mozilla.github.io/nunjucks) is the default render engine, however, we strongly recommend adopting the [React](#react) engine.
+### Common assumptions
+
+1. Templates may contain [Nunjucks filters or helper functions](https://mozilla.github.io/nunjucks/templating.html#builtin-filters). [Read more about filters](#filters).
+1. Templates may contain `partials` (reusable chunks). They must be stored in the `.partials` directory under the template directory. [Read more about partials](#partials).
+1. Templates may contain multiple files. Unless stated otherwise, all files will be rendered.
+1. The default variables you have access to in any the template file are the following:
```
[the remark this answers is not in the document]
jonaslagoni: I am guessing its only the variables you refer to. Even though its the same variables they are provided through the props, which kinda needed its own description. Unless you think people will understand the difference ? 🤔

docs/authoring.md
```diff
@@ -259,3 +224,116 @@ There are some template parameters that have a special meaning:
...
+}
+```
+
+The exported default function returns a `File` component as a root component which the [Generator](https://github.com/asyncapi/generator) uses to figure out what file should be generated. In our case we overwrite the default functionality of saving the file as `MyTemplate.js` but instead use the filename `asyncapi.md`. It is then specified that we should render `Some text that should render as is\n` within that file. Notice the `\n` character at the end, this is something that is automatically added after `Text` component. 
+
+For further information about components, props etc. see the [Generator React SDK](https://github.com/asyncapi/generator-react-sdk)
+
+### Common assumptions
+
+1. All files within the template directory are rendered if their root component is of type `File` and the path is not part of the `nonRenderableFiles` configuration.
```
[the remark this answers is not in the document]
jonaslagoni: Yea, much better 😄

docs/authoring.md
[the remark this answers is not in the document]
derberg: exactly, so once we have more, we can introduce structure, now it looks strange unless you have another PR coming already 😄

docs/authoring.md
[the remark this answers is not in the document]
derberg: for me, the meaning behind this point was to explain to template developer what is the context of the nunjucks template, what data I have access to, that it is the spec before and after parsing + config params.

How it works with react? what do I have access to in react template? in my link, I refer to the object argument you pass there to the function

derberg (changes requested):
ok, forget about the issue with nodejs template, it just doesn't work with the official asyncapi example as there is a bug in `stripProtocol` filter that works wrong if the protocol is not part of the url

docs/authoring.md
```diff
@@ -259,3 +224,118 @@ There are some template parameters that have a special meaning:
 |Name|Description|
 |---|---|
 |`server`| It is used to let the template know which server from the AsyncAPI specification file you want to use. In some cases, this may be required. For instance, when generating code that connects to a specific server. Use this parameter in case your template relies on users' information about what server from the specification file they want to use during generation. You also need this parameter if you want to use `server.protocol` notation within `conditionalFiles` configuration option. Once you decide to specify this parameter for your template, it is recommended you make it a mandatory parameter otherwise feature like `conditionalFiles` is not going to work if your users do not use this parameter obligatory.
+
+
+## React
+
+[React](https://reactjs.org) is the render engine that we strongly suggest you should use for any new templates. The only reason it is not the default render engine is to stay backward compatible.
+
```
derberg, line 232: please add also an important point about components reusability that we could not achieve with nunjucks (like share template file across code templates for the same language)

docs/authoring.md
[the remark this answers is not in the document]
jonaslagoni: yea I actually agree, I'll change it.

lib/renderer/react.js
```diff
@@ -0,0 +1,79 @@
...
+      permissions = renderContent.metadata.permissions;
+    }
+    if (renderContent.metadata.fileName !== undefined) {
+      const newFileName = renderContent.metadata.fileName.trim();
+      const basepath = path.basename(filePath);
+      filePath = filePath.replace(basepath, newFileName);
+    }
+  }
+  await writeFile(filePath, content, {
+    mode: permissions
```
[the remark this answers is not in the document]
jonaslagoni: Yea, I just took the active choice of actually stating our default value while not relying on what `fs` states 😄

docs/authoring.md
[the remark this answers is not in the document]
jonaslagoni: That is a very good question.. I actually thought I added that 👻 I guess not 🤷

lib/renderer/react.js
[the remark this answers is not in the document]
derberg: doesn't have to be changed, but for me as a future maintainer it was confusing and I had to check docs because I was interested if we need some special permissions 😄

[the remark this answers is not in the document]
jonaslagoni: But I mean, that is the expected permissions which we want on our generated files, regardless of how `fs` changes in the feature 🤔 But yea I feel like it had to be stated instead of people having to go to do the docs of `fs` to figure it out 😄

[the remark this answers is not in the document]
derberg: yeah, just leave a comment and all be good ☮️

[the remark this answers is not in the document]
jonaslagoni: Done 🎉

docs/authoring.md
jonaslagoni: @derberg but thats also possible with nunjucks right? You just have to point to the direct source in node_modules when installed as a dependency 🤔 ?

docs/authoring.md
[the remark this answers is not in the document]
jonaslagoni: @derberg I have added it to the [common assumptions](https://github.com/asyncapi/generator/blob/be7dc5b83da90986ac7215a206f46341ddb7a2ac/docs/authoring.md#common-assumptions-1), what you think?

docs/authoring.md
derberg: oh, you mean like add a template to dependency and then point to it directly, like when importing macro? Technically it works, but it is error-prone, you kinda hardcode yourself to a template file structure, we would never go this way

derberg: also, components testability improves how we can test template's output

jonaslagoni: Does not have to be a template dependency, could be a library containing nunjucks partials in some directory. Not quite sure how the generator behaves in regards to outside partials. however do you still think we should add the reason since it is already stated right below in the list 🤔 ?

docs/authoring.md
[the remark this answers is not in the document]
derberg: @jonaslagoni isn't it duplicated content? I mean shouldn't we just put it where it was, at the top of the document, if it is for any render engine?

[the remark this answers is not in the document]
jonaslagoni: From my POV it is different ways it gets provided, i.e. nunjucks provide it globally, and React provide it though props. But yea the props themself are the same, the context in which they are provided are different and has to be stated. Not sure what makes most sense when seen from the outside? 🤷

docs/authoring.md
jonaslagoni: > also, components testability improves how we can test template's output

True, this is missing.

derberg: I think so, because anything else with nunjucks would always be dependent on files structure. So if in a possible library of reusable macros, you would change folder structure, that would be a breaking change, something you do not want and it is easy to forget as it doesn't feel natural. This is in my opinion why we never even considered doing this ourselves and promoting such support.

Because we use React, we do not have such a problem.

docs/authoring.md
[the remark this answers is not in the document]
derberg: @jonaslagoni oh yes, sorry, you are right, technically it is a difference as you stated so makes sense to describe it the way you did 👍🏼

docs/authoring.md
jonaslagoni: @derberg added the part about testability also the part about reusability is also stated here -> https://github.com/asyncapi/generator/blob/775d1ea6032fc9e403bb23afd100f317771af9f3/docs/authoring.md#react

derberg (approved):
lgtm
great work!

derberg:
@jonaslagoni you will have to update to the latest master to enable merge

magicmatatjahu (approved):

PR #472, opened 2021-01-04
fix: copy non-JS files in the React templates by default

### LINKED ISSUE #466: Copy non-JS files in the React render engine
At the moment the Generator in the React case doesn't copy non-JS files in template, but skipped them. Workaround is add files to `nonRenderableFiles` array in configuration of template in `package.json`, but for template like https://github.com/asyncapi/nodejs-template/tree/master/template it isn't cool and has big impact to development: Each non-JS file must be added manually to an array.

magicmatatjahu:
**Description**

As in title: copy all non-JS files in the React render engine by default. Problem is described in the #466 issue:

> At the moment the Generator in the React case doesn't copy non-JS files in template, but skipped them. Workaround is add files to `nonRenderableFiles` array in configuration of template in `package.json`, but for template like https://github.com/asyncapi/nodejs-template/tree/master/template it isn't cool and has big impact to development: Each non-JS file must be added manually to an array.

Also if someone doesn't want render some `.js` file then can include it inside `nonRenderableFiles` array.

**Related issue(s)**
Resolves #466

### DIFF
--- a/lib/generator.js
+++ b/lib/generator.js
@@ -639,9 +639,7 @@ class Generator {
       }
     }
 
-    if (this.isNonRenderableFile(relativeSourceFile)) {
-      return await copyFile(sourceFile, targetFile);
-    }
+    if (this.isNonRenderableFile(relativeSourceFile)) return await copyFile(sourceFile, targetFile);
 
     await this.renderAndWriteToFile(asyncapiDocument, sourceFile, targetFile);
   }
--- a/lib/generator.js
+++ b/lib/generator.js
@@ -28,7 +28,8 @@ const {
   exists,
   fetchSpec,
   getInvalidOptions,
-  isReactTemplate
+  isReactTemplate,
+  isJsFile,
 } = require('./utils');
 const { registerFilters } = require('./filtersRegistry');
 const { registerHooks } = require('./hooksRegistry');
@@ -637,7 +638,10 @@ class Generator {
         if (!valid) return log.debug(`${relativeSourceFile} was not generated because condition specified for this file in template configuration in conditionalFiles matched`);
       }
     }
-    if (this.isNonRenderableFile(relativeSourceFile)) return await copyFile(sourceFile, targetFile);
+
+    if (this.isNonRenderableFile(relativeSourceFile)) {
+      return await copyFile(sourceFile, targetFile);
+    }
 
     await this.renderAndWriteToFile(asyncapiDocument, sourceFile, targetFile);
   }
@@ -687,7 +691,9 @@ class Generator {
    */
   isNonRenderableFile(fileName) {
     if (!Array.isArray(this.templateConfig.nonRenderableFiles)) return false;
-    return this.templateConfig.nonRenderableFiles.some(globExp => minimatch(fileName, globExp));
+    if (this.templateConfig.nonRenderableFiles.some(globExp => minimatch(fileName, globExp))) return true;
+    if (isReactTemplate(this.templateConfig) && !isJsFile(fileName)) return true;
+    return false;
   }
 
   /**
--- a/lib/generator.js
+++ b/lib/generator.js
@@ -688,8 +688,9 @@ class Generator {
    * @return {boolean}
    */
   isNonRenderableFile(fileName) {
-    if (!Array.isArray(this.templateConfig.nonRenderableFiles)) return false;
-    if (this.templateConfig.nonRenderableFiles.some(globExp => minimatch(fileName, globExp))) return true;
+    const nonRenderableFiles = this.templateConfig.nonRenderableFiles || [];
+    if (!Array.isArray(nonRenderableFiles)) return false;
+    if (nonRenderableFiles.some(globExp => minimatch(fileName, globExp))) return true;
     if (isReactTemplate(this.templateConfig) && !isJsFile(fileName)) return true;
     return false;
   }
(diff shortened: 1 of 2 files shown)


lib/generator.js
```diff
@@ -637,7 +638,10 @@ class Generator {
         if (!valid) return log.debug(`${relativeSourceFile} was not generated because condition specified for this file in template configuration in conditionalFiles matched`);
       }
     }
-    if (this.isNonRenderableFile(relativeSourceFile)) return await copyFile(sourceFile, targetFile);
```
derberg, line 640: is it because of prettier?

lib/utils.js
```diff
@@ -121,7 +121,18 @@ utils.isFilePath = (str) => {
 };
 
 /**
- *  Get version of the generator
+ * Checks if given file path is JS file
+ * 
+ * @param {String} filename Information representing file path
```
derberg, line 126: `information`

lib/utils.js
```diff
@@ -121,7 +121,18 @@ utils.isFilePath = (str) => {
...
+ * @returns {Boolean}
+ */
+utils.isJsFile = (filepath) => {
+  const ext = filepath.split('.').pop() || '';
+  return ['js', 'jsx', 'cjs'].includes(ext);
```
derberg, line 131: what about TS?

derberg (changes requested):
awesome, just few questions

lib/generator.js
magicmatatjahu: I don't like the if-s without brackets 😄

derberg: but this is a one-liner? Do you have issues with all ifs? even one liner? 😮

lib/utils.js
magicmatatjahu: At the moment we don't support TS in sdk https://github.com/asyncapi/generator-react-sdk/issues/3 so I don't want to add this.

lib/generator.js
derberg: for the record, I love them 😄 I'm ok to accept whatever some tool enforces though

lib/utils.js
derberg: You got me lost here 😄 but I also have to honestly tell you I don't remember what the underlying problem is here 😄 I mean I remember we talked about it and I asked you to create an issue but do not remember how it works now, what with other files with other extensions 😄 I thought the problem is only with js/ts related files and the whole rest is copied over to output

lib/generator.js
magicmatatjahu: I can change it to inline 😄

lib/utils.js
magicmatatjahu: Ok, let's describe the problem:
- Nunjucks flow works in this way that it reads each file in template and render to output's string, so your file may not have any Nunjucks syntax and Generator should copy it (render it) without problems. However you can have binary files and this is not easy to render to string 😄 so for that we have `nonRenderableFiles` array in config - to skip binary files and other non-renderable stuff
- in React we can render only files with `.js, .jsx, .cjs`(in future also `tsx`) - we skip another in React-SDK ([here](https://github.com/asyncapi/generator-react-sdk/blob/master/src/renderer/template.ts#L13)), so in the React case, Generator's flow works like (we didn't change anything in Generator when we developed React engine, only add new renderer): put filepath and context (parsed asyncapi etc) to [this](https://github.com/asyncapi/generator-react-sdk/blob/master/src/renderer/template.ts#L12) function, if file has one of above extension render it, if not skip it (not copy!). This PR contains fix to copy non-js files for React templates. Also, if someone doesn't want render some `.js` file, with this PR can includes it inside `nonRenderableFiles` array to copy to output.

Much simpler, Generator for every renderer works like:
1. take file in template
2. check if `nonRenderableFiles` array includes given file, if yes skip rendering,
3. put path of file and context to renderer (Nunjucks or React)
4. save output to file
5. make 1-4 steps for next file in template

So in React we have a problem between 2-3 steps. We're skipping files without `.js` inside React-SDK because we decided in develop that it'll be better for DX.

lib/generator.js
magicmatatjahu: Changed hah 😄

lib/utils.js
derberg: thanks for reminding me 😄 

to summarize, after this implementation these files will be copied over without adding them to the list of non-renderable files:
- https://github.com/asyncapi/python-paho-template/blob/master/template/README.md
- https://github.com/asyncapi/python-paho-template/blob/master/template/config-template.ini
- https://github.com/asyncapi/nodejs-template/blob/master/template/.eslintrc

but what about https://github.com/asyncapi/nodejs-template/blob/master/template/src/lib/config.js. This one should still explicitly be added to non-renderable files right? which makes sense, just want to confirm we have the same understanding 😉

magicmatatjahu: If `paho` will be rewritten to React, yes :) As I wrote Nunjucks has another flow with files inside template and React another due to this line https://github.com/asyncapi/generator/pull/472/files#diff-ccc178369ef3ad89acc2592970d94838d2dc83c38731610f3675be2ee3e6f99aR693

derberg: of course, I assume python and nodejs templates will be rewritten to react 😛 so I ask upfront what is missing before we migrate

derberg (approved):

derberg (approved):

PR #481, opened 2021-01-14
feat: changed to node>12.16 and npm>6.13.7

### LINKED ISSUE #480: Support later versions of NodeJS
As suggested by @derberg in our conversation in Slack:

> I think it would make sense to actually just have "node": ">12.16" to enable others to use newer versions and report issues upfront so we know about them before we jump to the new version.

Personally I'd like to use this package in my `node v15.5.0` project and provide feedback.

orfins:
Resolves #480

### DIFF
--- a/package.json
+++ b/package.json
@@ -8,8 +8,8 @@
     "ag": "./cli.js"
   },
   "engines": {
-    "node": ">12.16 <15",
-    "npm": ">6.13.7 <7"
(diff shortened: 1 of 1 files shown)


orfins:
I accidentally auto-deleted the forked repository so I'm closing this PR and opening a new one 🤦🏻‍♂️

PR #503, opened 2021-02-09
fix: remove restriction to node<15 and npm<7 to enable testing on newer versions

### LINKED ISSUE #480: Support later versions of NodeJS

orfins:
Resolves #480

### DIFF
--- a/README.md
+++ b/README.md
@@ -66,10 +66,12 @@ You can find above templates and the ones provided by the community in **[this l
 
 ## Requirements
 
-* v12.16+ < Node.js < 15
-* v6.13.7+ < npm < 7
+* v12.16+ < Node.js
+* v6.13.7+ < npm
 
 Install both packages using [official installer](https://nodejs.org/en/download/). After installation make sure both packages have proper version by running `node -v` and `npm -v`. To upgrade invalid npm version run `npm install npm@latest -g`
+
+**Please note that non-LTS versions of Node and npm are not supported. **
 
 ## Using from the command-line interface (CLI)
 
--- a/README.md
+++ b/README.md
@@ -71,7 +71,9 @@ You can find above templates and the ones provided by the community in **[this l
 
 Install both packages using [official installer](https://nodejs.org/en/download/). After installation make sure both packages have proper version by running `node -v` and `npm -v`. To upgrade invalid npm version run `npm install npm@latest -g`
 
-**Please note that non-LTS versions of Node and npm are not supported. **
+**Generator is tested at the moment against Node 14 and NPM 6. Using newer versions is enabled but we do not guarantee they work well. Please provide feedback on the issues.**
+
+how about this?
 
 ## Using from the command-line interface (CLI)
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -66,11 +66,13 @@ You can find above templates and the ones provided by the community in **[this l
...
 ## Requirements
 
-* v12.16+ < Node.js < 15
-* v6.13.7+ < npm < 7
+* v12.16+ < Node.js
+* v6.13.7+ < npm
 
 Install both packages using [official installer](https://nodejs.org/en/download/). After installation make sure both packages have proper version by running `node -v` and `npm -v`. To upgrade invalid npm version run `npm install npm@latest -g`
 
+**Please note that non-LTS versions of Node and npm are not supported. **
```
derberg, line 74: I'm afraid [npm 7 already reached GA](https://github.blog/2021-02-02-npm-7-is-now-generally-available/) so we need to improve this message a bit 😅 

```suggestion
**Generator is tested at the moment against Node 14 and NPM 6. Using newer versions is enabled but we do not guarantee they work well. Please provide feedback on the issues.**

how about this?
```

derberg:
I change a bit the title to make it super clear what is the most important change and why

README.md
```diff
@@ -66,11 +66,15 @@ You can find above templates and the ones provided by the community in **[this l
...
-* v12.16+ < Node.js < 15
-* v6.13.7+ < npm < 7
+* v12.16+ < Node.js
+* v6.13.7+ < npm
 
 Install both packages using [official installer](https://nodejs.org/en/download/). After installation make sure both packages have proper version by running `node -v` and `npm -v`. To upgrade invalid npm version run `npm install npm@latest -g`
 
+**Generator is tested at the moment against Node 14 and NPM 6. Using newer versions is enabled but we do not guarantee they work well. Please provide feedback on the issues.**
+
+how about this?
```
derberg, line 76: sorry, I did a wrong suggestion 🤦🏼 

```suggestion
> Generator is tested at the moment against Node 14 and NPM 6. Using newer versions is enabled but we do not guarantee they work well. Please provide feedback on the issues.
```

orfins: Yeah i saw the last line just now lol

orfins:
@derberg should I squash the commits and change the message to be similar to the title?

derberg:
@orfins no need. We by default allow only the `Squash and merge` option, and it also allows to change the commit message, so I'll handle that 👍🏼

derberg:
I'm just thinking if it really should be `feat` (minor release) and not `fix` (patch release). In the end, it is not a bug fix, but not a feature too 😅

orfins:
> I'm just thinking if it really should be `feat` (minor release) and not `fix` (patch release). In the end, it is not a bug fix, but not a feature too 😅

What about `refactor`? Or maybe we could say that installing the package on node>14 and npm>6 is a feature? 🤣

derberg:
`refactor` will not trigger the release. I'll change to `fix` and we will get 1.1.9 out of it. It is for sure not a feature and we are kinda fixing a bug in the configuration

derberg (approved):

PR #513, opened 2021-02-16
feat: map schema references from baseUrl to folder

### LINKED ISSUE #512: map schema base url to local folder
#### Reason/Context

1. The resolution of schema references with relative pathes e.g. `../mypath` depend on the implementing library and may not work. Absolue URIs solve the problem.
2. During unit testing, you often need to use local JSON Schema files instead of downloading the schema from a Web-Server.
3. In cases where a schema URI is an [identifier](https://json-schema.org/draft/2019-09/json-schema-core.html#rfc.section.8.2.2) but not necessarily a network locator, it is required to do a mapping to local files. 

For example, I would love to map `https://example.com/` to a local base folder, so the schema file `common/shared.json` can be resolved locally:
```
{
  "$id": "https://example.com/foo/bar.json",
  "items": {
    "$ref": "https://example.com/common/shared.json"
  }
}
```

#### Description

Allow to map from a schema base URI to a local folder with an additional CLI parameter:

```
--map-base-url <url>:<localBaseFolder>
```

muenchhausen:
**Description**

introdce two new cli parameters

- `--mapFromBaseUrl <url>`: maps all schema references with this base url to folder specified by `--mapToBaseFolder`
- `--mapToBaseFolder <baseFolder>`: maps all schema references starting with url specified by `--mapFromBaseUrl` to this folder

**Related issue(s)**
Resolves #512

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -14,7 +14,7 @@ Contributions are more than welcome. If you want to contribute, please make sure
 Before creating a Pull Request you should validate your code with ESLint.
 
 ```
-npm run eslint
+npm run lint
 ```
 
 ## Developing with Docker
--- a/cli.js
+++ b/cli.js
@@ -19,6 +19,7 @@ let template;
 const params = {};
 const noOverwriteGlobs = [];
 const disabledHooks = {};
+const mapBaseUrlToFolder = { url: '', folder: process.cwd()};
 
 const parseOutput = dir => path.resolve(dir);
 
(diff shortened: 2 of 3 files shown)


derberg:
@muenchhausen do you plan to continue to work on this one? not sure if you noticed my comment in the issue

muenchhausen:
> @muenchhausen do you plan to continue to work on this one? not sure if you noticed my comment in the issue

Sorry I was quite busy the last days. Thanks - it is a valid point that cli flags should not depend on each other - if possible. I will prepare a patch this weekend.

muenchhausen:
This pull request got replaced by https://github.com/asyncapi/generator/pull/529 because I had to use a different branch.

PR #514, opened 2021-02-17
ci: add simple integration test

smoya:
**Description**
This PR adds a simple integration test useful for CI which will ensure the generator works as a whole.

- Adds a new script called `test:integration`, which basically generates a test spec (Streetlights tutorial located at https://bit.ly/asyncapi) by using the `cli.js`.
- It runs the new `test:integration` script in the `test` script so the [CI workflow](https://github.com/asyncapi/generator/blob/master/.github/workflows/if-nodejs-pr-testing.yml) runs it on each PR.

**Related issue(s)**
The new test will catch dependency issues like https://github.com/asyncapi/generator/pull/510

### DIFF
--- a/package.json
+++ b/package.json
@@ -12,8 +12,9 @@
     "npm": ">6.13.7"
   },
   "scripts": {
-    "test": "jest --coverage && npm run test:integration",
-    "test:dev": "jest --watchAll --coverage",
+    "test": "npm run test:unit && npm run test:integration",
+    "test:unit": "jest --coverage",
+    "test:dev": "npm run test:unit -- --watchAll",
     "test:integration": "node cli.js https://bit.ly/asyncapi @asyncapi/html-template -o test/output --force-write && test -e test/output/index.html",
     "docs": "jsdoc2md lib/generator.js > docs/api.md",
     "release": "semantic-release",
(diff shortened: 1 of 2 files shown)


derberg:
@smoya the error you're getting atm on windows is related to a bug for windows and nvm (GA uses nvm) and unfortunately can be fixed only with https://github.com/asyncapi/ts-nats-template/blob/master/package.json#L34 as `npmi` that we use in generator for templates installation is not actively maintained 😞

smoya:
Thanks @derberg for pointing that out. I'm gonna add that hack into the package.json scripts then so it can work properly on Windows.

fmvilas (approved):
Nooice 🚫 🧊

derberg:
@smoya for the conventional commits scopes are rather for packages, parts of the project, not for stuff like ci, so my recommendation is to do prefix `test:` or `ci:`

package.json
```diff
@@ -12,8 +12,9 @@
     "npm": ">6.13.7"
   },
   "scripts": {
-    "test": "jest --coverage",
+    "test": "jest --coverage && npm run test:integration",
```
derberg, line 15: what if we introduce also `test:unit` for `jest --coverage`?

while developing I often run `npm test` but now will stop as I do not want to run integration every time and best would be to have `npm run test:unit` too. We have a similar approach in the parser https://github.com/asyncapi/parser-js/blob/master/package.json

smoya: Completely agree 👍  Changing!

smoya:
> @smoya for the conventional commits scopes are rather for packages, parts of the project, not for stuff like ci, so my recommendation is to do prefix `test:` or `ci:`

@derberg Thanks for the clarification. I will made the change into the squashed commit before merging.

fmvilas (approved):

Issue #515, opened 2021-02-24
Support for asynchronous filters in Nunjucks


magicmatatjahu:
I noticed during the `model-sdk` integration in `java-spring-template` that we do not have support for asynchronous filters. Unfortunately, our `sdk-model` generates models asynchronously. I have an idea how to fix it and add asynchronous filters :)

Related to https://github.com/asyncapi/shape-up-process/issues/49

PR #516, opened 2021-02-24
feat: add support for asynchronous filters in Nunjucks

magicmatatjahu:
**Description**

Add support for asynchronous filters in Nunjucks:
- check before saving filter to registry that function is asynchronous, if yes, add as third argument to `addFilter` function `true` - it indicates that to filter should be always added callback with signature `callback(err, result)`.
- update docs with example.

In Nunjucks docs we can read:
image

**Related issue(s)**
Resolves https://github.com/asyncapi/generator/issues/515

### DIFF
--- a/docs/authoring.md
+++ b/docs/authoring.md
@@ -343,7 +343,7 @@ You can use the filter function in your template as in the following example:
 const {{ channelName | camelCase }} = '{{ channelName }}';
 ```
 
-The generator supports also asynchronous filters. Asynchronous filters receive as last argument a callback to resume rendering. Asynchronous filters must be annotated with the `async` keyword. See the example of how to use asynchronous filters:
+The generator supports also asynchronous filters. Asynchronous filters receive as last argument a callback to resume rendering. Asynchronous filters must be annotated with the `async` keyword. Make sure to call the callback with two arguments: `callback(err, res)`. `err` can be `null`. See the example of how to use asynchronous filters:
 
 ```js
 async function asyncCamelCase(str, callback) {
--- a/docs/authoring.md
+++ b/docs/authoring.md
@@ -343,4 +343,21 @@ You can use the filter function in your template as in the following example:
 const {{ channelName | camelCase }} = '{{ channelName }}';
 ```
 
+The generator supports also asynchronous filters. Asynchronous filters receive as last argument a callback to resume rendering. Asynchronous filters must be annotated with the `async` keyword. See the example of how to use asynchronous filters:
+
+```js
+async function asyncCamelCase(str, callback) {
+  try {
+    const result = // logic for camel casing str
+    callback(null, result);
+  } catch (error) {
+    callback(error);
+  }
+}
+filter.renderAsyncContent = renderAsyncContent;
+
+// using in template
+{{ channelName | asyncCamelCase }}
+```
+
 In case you have more than one template and want to reuse filters, you can put them in a single library. You can configure such a library in the template configuration under `filters` property. You can also use the official AsyncAPI [filters library](https://github.com/asyncapi/generator-filters). To learn how to add such filters to configuration [read more about the configuration file](#configuration-file).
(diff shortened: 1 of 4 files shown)


magicmatatjahu:
@derberg I have one code smell with line `const asyncFunctionCtor = (async () => {}).constructor;` Unfornatelly this is needed to check that function is async.

derberg:
@magicmatatjahu is there no other way? what about `const isAsync = myFunction.constructor.name === "AsyncFunction";`

magicmatatjahu:
@derberg Comparing constructor's name is usually not a good idea, but I'll see if the sonar has a problem with that.

magicmatatjahu:
@derberg I changed it and it helped to remove the smell code, but I still think that comparing name is not good idea, but I can accept it here.

docs/authoring.md
```diff
@@ -343,4 +343,21 @@ You can use the filter function in your template as in the following example:
 const {{ channelName | camelCase }} = '{{ channelName }}';
 ```
 
+The generator supports also asynchronous filters. Asynchronous filters receive as last argument a callback to resume rendering. Asynchronous filters must be annotated with the `async` keyword. Make sure to call the callback with two arguments: `callback(err, res)`. `err` can be `null`. See the example of how to use asynchronous filters:
```
jonaslagoni, line 346:
```suggestion
The generator also supports asynchronous filters. Asynchronous filters receive as last argument a callback to resume rendering. Asynchronous filters must be annotated with the `async` keyword. Make sure to call the callback with two arguments: `callback(err, res)`. `err` can be `null`. See the following example of how to use asynchronous filters:
```

What about filters that return a promise? 🤔

lib/utils.js
```diff
@@ -150,3 +150,13 @@ utils.getInvalidOptions = (generatorOptions, options) => {
   if (typeof options !== 'object') return [];
   return Object.keys(options).filter(param => !generatorOptions.includes(param));
 };
+
+/**
+ * Determine whether the given `fn` is an async function.
```
jonaslagoni, line 155:
```suggestion
 * Determine whether the given function is asynchronous.
```

lib/utils.js
```diff
@@ -150,3 +150,13 @@ utils.getInvalidOptions = (generatorOptions, options) => {
...
+ * @private
+ * @param {*} fn
+ * @returns {Boolean}
```
jonaslagoni, line 158:
```suggestion
 * @param {*} fn to check 
 * @returns {Boolean} is function asynchronous
```

jonaslagoni (changes requested):
Just got some small suggestions 😄

docs/authoring.md
magicmatatjahu: Nunjucks doesn't support it. Determine that function returns Promise is impossible without invoke it, so we must think about another way to resolve problem.

jonaslagoni: Yea, let's not focus on that at all, if people want it they will ask for it 😄

jonaslagoni (dismissed):
LGTM 👍

docs/authoring.md
magicmatatjahu: My first thought how to resolve that: we should return as filter not function but object with metadata that function is asynchronous, like

```js
const filters = module.export;

filters.someFilter = {
  isAsync: true,
  fn: function() { return new Promise() ... }
}
```

docs/authoring.md
```diff
@@ -343,4 +343,21 @@ You can use the filter function in your template as in the following example:
 const {{ channelName | camelCase }} = '{{ channelName }}';
 ```
 
+The generator also supports asynchronous filters. Asynchronous filters receive as last argument a callback to resume rendering. Asynchronous filters must be annotated with the `async` keyword. Make sure to call the callback with two arguments: `callback(err, res)`. `err` can be `null`. See the following example of how to use asynchronous filters:
+
```
jonaslagoni, line 347: Maybe we should add the fact that promise cannot be used 🤔

magicmatatjahu: Or make this workaround https://github.com/asyncapi/generator/pull/516#discussion_r585493590 for Promises 😄

jonaslagoni: I feel like it is more a bandage and complicates the filters more than we should.

magicmatatjahu:
@jonaslagoni I updated docs with Promise example :)

jonaslagoni (approved):

PR #517, opened 2021-02-24
feat: enable npm 7 support and remove npmi

### LINKED ISSUE #391: The generate does not install template modules correctly if local

### LINKED ISSUE #425: The generator does not use templates installed in node_modules

### LINKED ISSUE #449: docker run fetches npm dependencies each time
#### Describe the bug
whenever a want to generate html template based on provided docker command, it goes to fetch all node dependencies again.

#### How to Reproduce
as provided in [docs](https://github.com/asyncapi/generator#cli-usage-with-docker)

#### Expected behavior
the docker file should be organized in a way that it can use the installed npm modules in container volume

### LINKED ISSUE #482: How to use pre-install template?
#### Reason/Context

We have all API files managed in one git repo and have CI pipeline to generate document from these files. I'm going to support asyncapi and use @asyncapi/generator to generate HTML document. The CI pipeline runs in a Docker container so I need to pre-install both generator and html-template into the Docker image. 

There are several reasons that we don't want generator to install template on-the-fly: we can't secure the internet connection in the CI environment; we don't want to waste time to download in every CI job; and also our strategy is that everything in CI should be immutable. The way described in #449 to mount a volume doesn't fit for our CI neither.

I have tried to install both generator and template in Dockerfile but generator doesn't use the global installed template. It looks for template under its own node_module folder: `/usr/local/lib/node_modules/@asyncapi/generator/node_modules/`

```
RUN npm add -g @asyncapi/generator @asyncapi/html-template
```

I also tried to install template using generator CLI, but the command failed because it only download and install template during generation.

```
RUN npm add -g @asyncapi/generator
RUN ag -i @asyncapi/html-template
```
Finally I made it work by adding a dummy asyncapi file into the image then generate it during Docker build:

```
COPY docker-content/dummy-asyncapi.yaml .
RUN npm add -g @asyncapi/generator \
    && ag -o /tmp/output -i dummy-asyncapi.yaml @asyncapi/html-template \
    && rm dummy-asyncapi.yaml && rm -rf /tmp/output
```

However it's just a workaround and not looks elegant.

#### Description

I see there might be two alternatives to resolve the problem:

**Alt#1**

Generator uses the global installed template. It skips on-the-fly installation if template already exists.

**Alt#2**

Generator CLI provides template management command. For example `ag install-template <template>` to install the template into its node_modules folder.

### LINKED ISSUE #505: issue with with npm 7.x
#### Describe the bug

I am unable to run the generator with v15.8.0 (with npm 7.5.1).
It works fine with v14.15.0.

```
> async-api-docs@1.0.0 build
> ./node_modules/.bin/ag ifc-pre-processor/asyncapi.yaml @asyncapi/html-template -o dist

Something went wrong:
TypeError: must call as: npm.load(callback)
    at Object.load (/usr/local/lib/node_modules/npm/lib/npm.js:129:13)
    at npmi (/Users/adameri/Developer/bimspot/async-api-docs/node_modules/npmi/npmi.js:189:9)
    at /Users/adameri/Developer/bimspot/async-api-docs/node_modules/@asyncapi/generator/lib/generator.js:377:7
npm ERR! code 1
npm ERR! path /Users/adameri/Developer/bimspot/async-api-docs
npm ERR! command failed
npm ERR! command sh -c ./node_modules/.bin/ag ifc-pre-processor/asyncapi.yaml @asyncapi/html-template -o dist
```

#### How to Reproduce

Install node 15, delete existing output folder, run generator.

#### Expected behavior

Docs are generated in the output folder

derberg:
The ambition of this PR is to fix all the issues we have with the installation of templates. I think the main guilty dude here is `npmi`. I researched and all community tools for npm install are abandoned, or not actively maintained. Looks like recently people use `execa` or other tools to just call npm CLI programmatically but for our use case, it adds a lot of complexity (for the use case where we install a template from disk or GitHub URL (and we do not know the package name). Luckily looks like there is finally something better available https://github.com/npm/arborist that is built by the npm team ❤️ 

**Description of what so far is done**

- fix issues using generator with nvm. Workaround like manual bumping of one `npmi` dependency with a fix is not needed -> `"postinstall": "cd \"node_modules/npmi\" && npm install global-npm@0.4.1"`
- support latest node (15) and npm (7)
- basic library integration test using both, nunjucks and react templates using jest snapshots
- added special workflow and test project to test generator as dependency on node 12, 14, 15
- added new integration test that test generator and its templates, both as dependency in a project
- added a lot of debug logs so it is not super easy and clearer to see how the template installation went

**Related issue(s)**
Fixes #505 
Fixes #482 
Fixes #425
Fixes #449


stavalfi:
My humble opinion is - drop all `npm` related code from the generator. 

Use case:
1. 
My company is moving to yarn2 to have the zero-installation feature. After the migration, the installation time is 7s. 

There is no node_modules.

Also, There is a new package-manager in rust. node_modules in probably not going to be there as well (https://dev.to/cloudx/a-new-npm-is-coming-5heh)

2. generating website takes too much time because it doesn't use the already-downloaded deps when using yarn mono-repo.

----

I understand that this is a huge request, I'm only asking you to consider it.


Thanks for the great work!

lib/generator.js
```diff
@@ -364,32 +363,36 @@ class Generator {
...
+
+      try {
+        const installResult = await arb.reify({
+          add: [this.templateName],
+          saveType: 'prod',
+          save: false
+        });
+        //trying to figure out if this is the right way to access the name of added package https://github.com/npm/arborist/issues/243
+        const addResult = arb[Symbol.for('resolvedAdd')];
+        if (!addResult) reject('Unable to resolve the name of added package. It was most probably not added to node_modules successfully');
```
fmvilas, line 385:
```suggestion
        if (!addResult) return reject('Unable to resolve the name of the added package. It was most probably not added to node_modules successfully');
```

test/__mocks__/@npmcli/arborist.js
```diff
@@ -0,0 +1,3 @@
+const arb = jest.genMockFromModule('@npmcli/arborist');
```
fmvilas, line 1: I think this `__mocks__` directory should be placed in the root directory, next to `node_modules`: https://jestjs.io/docs/en/manual-mocks#mocking-node-modules. Maybe that's why you can't manage to mock the module?

fmvilas (commented):
I couldn't test it but left some hints. Hope they help 🤞

test/__mocks__/@npmcli/arborist.js
derberg: the thing is that mocking module works well, but the class that I mock, doesn't contain property that is later exposed in the class (this property is added after `reify()`. So I have to somehow extend mock with additional mocked property, and this is where I fail. I mean I am able to add props, and when I console log the module inside test, the function is there. Thing is that it is lost inside the code that is tested because `new` (in the code) overrides what I add in the test . I hope that makes sense

derberg:
@stavalfi thanks a lot for taking the time to share your view

> My company is moving to yarn2 to have the zero-installation feature

can you share something more, or some references, is it working like npmx?
npm 7 us super fast compared to 6

> Also, There is a new package-manager in rust. node_modules in probably not going to be there as well (https://dev.to/cloudx/a-new-npm-is-coming-5heh)

yay, another package manager 😄 beta in March (blog post from April 2020 talked about it, but there are no RCs there yet so I don't believe it is going to happen in March), I definitely do not plan to touch it before it is out 😅 

> generating website takes too much time because it doesn't use the already-downloaded deps when using yarn mono-repo.

yeah, this is a different story, doesn't matter what package manager we use here, this is another part of the code that I did not touch yet but will. One of the issues about not reusing preinstalled modules is listed in the description (I also want to fix it)

test/__mocks__/@npmcli/arborist.js
fmvilas: I see. There are two things that come to my mind here:

1. Why don't you create the `resolvedAdd` mock here in this file? This way, it should be available without having to explicitly do it on the test.
2. When you do `new`, what you get is the result of "copying" the `prototype` of the object. So probably, what you have to do is to mock `arb.prototype.resolvedAdd` instead of `arb.resolvedAdd`.

Does it make sense?

derberg: > When you do new, what you get is the result of "copying" the prototype of the object

omg, fundamental knowledge 🤦🏼 
I love you mate! ❤️  tests are green like fields of Ireland ☘️

derberg:
Status for now. The current implementation works with the arborist the way it worked before. I work with the generator this way 2 days now. So support for the latest node and npm will be a fact after merge.

Just need to figure out what check we're missing that preinstalled templates (the ones added to dependencies) are ignored by the generator

stavalfi:
@derberg thanks for responding.

Let me ask different question - why the generator can't expect that the packages it needs - will be already installed? and then, the generator can just call `require(<some-deps>)`? the package-manager will find it. why the generator needs to install it for the user? 

I understand the motivation, but in really, the generator need to manually support every package-manager in the eco-system. why is it worth it?

derberg:
@stavalfi yeah, I meant npx of course, sorry 😄 

> Let me ask different question - why the generator can't expect that the packages it needs - will be already installed? and then, the generator can just call require(<some-deps>)? the package-manager will find it. why the generator needs to install it for the user?

it doesn't have to always install. This is a bug I want to solve here too. If you look into code we already have some logic to check if package is already available locally or not (we use it for development, for example, when I want to generate something from the template I develop locally). So the plan is to enable users to add templates to dependencies or install globally. It is important not only for speed but also for locking generation on a specific version of the template

derberg:
basic integration tests added

derberg:
ok, I finally added a test project where generator and template are used as dependencies, and I have a test that generates output and expects that something is missing from the output because older template version is used - test is failing for not because in code there is still a bug related to this. Test ready, now I can work on a fix

derberg:
Ready for the review 🚀 
Good luck and let the force be with you 💪🏼

jonaslagoni (changes requested):
Just a quick comment before reviewing @derberg remember to change https://github.com/asyncapi/generator#requirements

test/test-project/package.json
```diff
@@ -0,0 +1,15 @@
...
+        "test:cleanup": "rimraf \"../temp\""
+    },
+    "dependencies": {
+        "@asyncapi/generator": "file:../..",
+        "@asyncapi/html-template": "0.16.0"
+    },
+    "devDependencies": {
+        "jest": "^25.5.0",
+        "fs-extra": "^9.1.0"
+    }
```
jonaslagoni, line 14: Shouldn't this include `rimraf` as a dev dependency?

derberg:
@jonaslagoni good point, done 💪🏼

.github/workflows/pr-testing-with-test-project.yml
```diff
@@ -0,0 +1,30 @@
+#This test runs test on the test projects to make sure generator works as a library where it is a Node dependency along with the template
```
jonaslagoni, line 1:
```suggestion
#This workflow runs the tests in the test projects to make sure the generator works as a library where it is a Node dependency along with the template.
```

lib/generator.js
```diff
@@ -340,56 +338,67 @@ class Generator {
...
+        path: ROOT_DIR
       });
+
+      try {
+        const installResult = await arb.reify({
+          add: [this.templateName],
+          saveType: 'prod',
+          save: false
+        });
+        //trying to figure out if this is the right way to access the name of added package https://github.com/npm/arborist/issues/243
```
jonaslagoni, line 384: Does this sound like a temporary comment? 🤔

jonaslagoni (changes requested):
Looks great, just a few comments 😄 🎉

lib/generator.js
```diff
@@ -340,56 +338,67 @@ class Generator {
...
-          loglevel: 'http',
-          save: false,
-          audit: false,
-          progress: false,
-          only: 'prod'
-        },
-      }, (err, result) => {
-        if (err) return reject(err);
-        resolve(beautifyNpmiResult(result));
+      const debugMessage = force ? 'as you passed --install flag' : 'because the template cannot be found on disk';
```
fmvilas, line 369: Not saying "as" is incorrect there, just reads strangely to me:

```suggestion
      const debugMessage = force ? 'because you passed --install flag' : 'because the template cannot be found on disk';
```

fmvilas (changes requested):

test/test-project/package.json
derberg: it should, done 🙇🏼

lib/generator.js
derberg: 👍🏼 pushed fix here and in tests

lib/generator.js
derberg: yeap, they never replied 🤷🏼‍♂️ 
I will keep my issue open and if they write something important I will work on a new PR

comment removed

fmvilas (dismissed):
LGTM 👍 Ship it and break the internet!

lib/filtersRegistry.js
```diff
@@ -69,9 +68,16 @@ async function registerConfigFilters(nunjucks, templateDir, templateConfig) {
...
+  const promises = confFilters.map(async filtersModule => {
+    let mod;
+    try {
+      //first we try to grab module with filters by the module name
+      //this is when generation is used on production using remote templates
+      mod = require(path.resolve(DEFAULT_MODULES_DIR, filtersModule));
+    } catch (error) {
+      //in case template is local but was not installed in node_modules of the generator then we need to explicitly provide modules location
+      mod = require(path.resolve(templateDir, DEFAULT_MODULES_DIR, filtersModule));
+    }
```
laat, line 80: I think we must do the same in the hooksRegistry

https://github.com/derberg/generator/blob/7e94a395fcc8883271a9c82f0cab4e1aed504050/lib/hooksRegistry.js#L76-L77

derberg: oh, that was a good call to get you on a PR. thanks so much, good 👁️ , done

laat (approved):
LGTM 🎉

derberg:
@fmvilas can you approve again? I had to do an additional refactor for hooks as @laat noticed I forgot something

@jonaslagoni anything else from your side?

jonaslagoni (approved):
LGTM 🎉

fmvilas (approved):
🚀

derberg:
Thanks a lot for the review. I'll wait and merge tomorrow morning, maybe someone else will jump until then and spot something. Nah, it is not that I'm afraid to merge, nooooo...... it is just better to merge in the morning to have more time to address bug fixes before midnight 😆

Issue #455, opened 2020-12-08
Support TS in hooks and filters


magicmatatjahu:
#### Reason/Context

At the moment `generator` doesn't support TS in hooks and filters. Problematic is that if someone wants to use TS in template he/she can't, because mentioned hooks and filters won't work due to handle them by generator itself, so generator should handle transpilation to JS before rendering process.

#### Description

I don't think it will be a breaking change. We should do something like that https://github.com/kyma-project/website/blob/master/gatsby/index.js#L3 so before saving filters to nunjucks we should transpile TS code to JS code. This same for hooks.

fmvilas:
> Problematic is that if someone wants to use TS in template he/she can't...

I'm curious how many times this has been a problem 😄

PR #519, opened 2021-03-01
feat: add TypeScript support for hooks and filters 

magicmatatjahu:
**Description**

Add TypeScript supports in hooks and Nunjucks's filters:
- this PR only enables TS support for hooks/filters. Support for React will be added here https://github.com/asyncapi/generator-react-sdk/issues/3 
- code for TS is transpiling on-the-fly so developer doesn't have to transpile it before publishing to npm, but by this we must include `ts-node`, `typescript` and `source-map` packages in generator.
- add docs with general assumptions for using TS in hooks/filters.

**Related issue(s)**
Resolve https://github.com/asyncapi/generator/issues/455

### DIFF
--- a/docs/authoring.md
+++ b/docs/authoring.md
@@ -399,4 +399,4 @@ The AsyncAPI generator has TypeScript support for [hooks](#hooks) and Nunjucks's
 - Installing the `typescript` package and creating the` tsconfig.json` file isn't necessary.
 - Source code of the hook/filter must have `.ts` extension.
 - Each package related to the typings for TypeScript like `@types/node` must be installed in the template under `dependencies` array. This is because the Generator transpiles the TypeScript code on-the-fly while rendering the template, and cannot use packages under `devDependencies`.
-- Each package should have installed `@types/node` package to support typings for Node.
+- Each template should have installed `@types/node` package to support typings for Node.
--- a/docs/authoring.md
+++ b/docs/authoring.md
@@ -391,3 +391,12 @@ async function asyncCamelCase(str, callback) {
 ```
 
 In case you have more than one template and want to reuse filters, you can put them in a single library. You can configure such a library in the template configuration under `filters` property. You can also use the official AsyncAPI [filters library](https://github.com/asyncapi/generator-filters). To learn how to add such filters to configuration [read more about the configuration file](#configuration-file).
+
+## TypeScript support 
+
+The AsyncAPI generator has support for the TypeScript for [hooks](#hooks) and Nunjucks's [filters](#filters). Assumptions:
+
+- Installing the `typescript` package and creating the` tsconfig.json` file isn't necessary.
+- Source code of the hook/filter must have `ts` or `tsx` extension.
+- Each package related with the typings for TypeScript like `@types/node` must be installed in the template under `dependencies` array. This is because the Generator transpiles the TypeScript code on-the-fly while rendering the template.
+- Each package should have installed `@types/node` package to support typings for Node.
--- a/docs/authoring.md
+++ b/docs/authoring.md
@@ -397,6 +397,6 @@ In case you have more than one template and want to reuse filters, you can put t
 The AsyncAPI generator has support for the TypeScript for [hooks](#hooks) and Nunjucks's [filters](#filters). Assumptions:
 
 - Installing the `typescript` package and creating the` tsconfig.json` file isn't necessary.
-- Source code of the hook/filter must have `ts` or `tsx` extension.
+- Source code of the hook/filter must have `ts` extension.
 - Each package related with the typings for TypeScript like `@types/node` must be installed in the template under `dependencies` array. This is because the Generator transpiles the TypeScript code on-the-fly while rendering the template.
 - Each package should have installed `@types/node` package to support typings for Node.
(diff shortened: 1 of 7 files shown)


fmvilas:
I'm afraid this is introducing much more complexity to the Generator, especially the transpilation process. What need exactly is driving this PR?

magicmatatjahu:
@fmvilas How complexity? It only transpiles TS code on the fly when we import hooks or filters in `hook/filter` registries - user doesn't need to transpile anything before rendering template  :) Maybe misunderstanding for issue and PR is that I missed to link to another issue in `generator-react-sdk` - https://github.com/asyncapi/generator-react-sdk/issues/3 If we will have a TS support in react renderer, then we will have inconsistencies because hooks will have to be written in JS and the template itself will be written entirely in TS. Maybe TS support for filters is not needed, but this PR is only a preparation for TS support for React templates.

fmvilas:
Oh, ok 👍 I was missing that context. Thanks!

docs/authoring.md
```diff
@@ -391,3 +391,12 @@ async function asyncCamelCase(str, callback) {
...
 
 In case you have more than one template and want to reuse filters, you can put them in a single library. You can configure such a library in the template configuration under `filters` property. You can also use the official AsyncAPI [filters library](https://github.com/asyncapi/generator-filters). To learn how to add such filters to configuration [read more about the configuration file](#configuration-file).
+
+## TypeScript support 
+
+The AsyncAPI generator has TypeScript support for [hooks](#hooks) and Nunjucks's [filters](#filters). Assumptions:
+
+- Installing the `typescript` package and creating the` tsconfig.json` file isn't necessary.
+- Source code of the hook/filter must have `.ts` extension.
+- Each package related with the typings for TypeScript like `@types/node` must be installed in the template under `dependencies` array. This is because the Generator transpiles the TypeScript code on-the-fly while rendering the template, so cannot uses packages under `devDependencies`.
```
derberg, line 401:
```suggestion
- Each package related to the typings for TypeScript like `@types/node` must be installed in the template under `dependencies` array. This is because the Generator transpiles the TypeScript code on-the-fly while rendering the template, and cannot use packages under `devDependencies`.
```

docs/authoring.md
```diff
@@ -391,3 +391,12 @@ async function asyncCamelCase(str, callback) {
...
+- Each package should have installed `@types/node` package to support typings for Node.
```
derberg, line 402: what package?

lib/utils.js
```diff
@@ -127,3 +127,21 @@ utils.getInvalidOptions = (generatorOptions, options) => {
 utils.isAsyncFunction = (fn) => {  
   return fn && fn.constructor && fn.constructor.name === 'AsyncFunction';
 };
+
+/**
+ * Register `source-map-support` package.
```
derberg, line 132: please also put here the purpose of the source map support

lib/hooksRegistry.js
```diff
@@ -98,6 +98,11 @@ async function registerConfigHooks(hooks, templateDir, templateConfig) {
  * @param {Array} config List of hooks configured in template configuration
  */
 function addHook(hooks, mod, config) {
+  // for ESM `export default {}`
```
derberg, line 101: what does it mean, what esm has to do with this part?

derberg (commented):
@magicmatatjahu great work, only few comments, all docs related

docs/authoring.md
magicmatatjahu: Sorry, here should be `template` -> `Each template...`

lib/hooksRegistry.js
magicmatatjahu: For example. If you write export for hooks in CommonJS, then you have:

```js
module.exports = {
  'generate:before': (generator) => {
  }
};
```

But if you are using ESM in TS, you can write:

```ts
export default {
  'generate:before': (generator) => {
  }
}
```

Unfortunately, then it is transpiled to:

```js
module.exports = {
  default: {
      'generate:before': (generator) => {
      }
  }
};
```

derberg: oh, ok, you just made a comment to explain the `if`. Can you make it more descriptive 🙏🏼 I totally didn't map it in my head, probably because I use ESM very rarely. So please write this comment for people like me too 😄

docs/authoring.md
```diff
@@ -391,3 +391,12 @@ async function asyncCamelCase(str, callback) {
...
 In case you have more than one template and want to reuse filters, you can put them in a single library. You can configure such a library in the template configuration under `filters` property. You can also use the official AsyncAPI [filters library](https://github.com/asyncapi/generator-filters). To learn how to add such filters to configuration [read more about the configuration file](#configuration-file).
+
+## TypeScript support 
+
+The AsyncAPI generator has TypeScript support for [hooks](#hooks) and Nunjucks's [filters](#filters). Assumptions:
+
+- Installing the `typescript` package and creating the` tsconfig.json` file isn't necessary.
+- Source code of the hook/filter must have `.ts` extension.
+- Each package related to the typings for TypeScript like `@types/node` must be installed in the template under `dependencies` array. This is because the Generator transpiles the TypeScript code on-the-fly while rendering the template, and cannot use packages under `devDependencies`.
+- Each template should have installed `@types/node` package to support typings for Node.
```
derberg, line 402:
```suggestion
- Each template should have `@types/node` package installed to enable support for typings for Node.
```

derberg (approved):

PR #529, opened 2021-03-07
feat: map schema references from baseUrl to folder

### LINKED ISSUE #512: map schema base url to local folder

muenchhausen:
**Description**

introduce a new cli parameter

`--map-base-url <url>:<baseFolder>`: maps all schema references with this base url to folder

This can be tested by running 

```
./cli.js test/docs/apiwithref.json @asyncapi/html-template -o ./build/ --force-write --map-base-url https://schema.example.com/crm/:./test/docs/
```

**Related issue(s)**
Resolves #512

**Open**

1. Run the test from above as unit or integration test. How do you want me to implement that? Should I test against `cli.js` or `generator.js`? 
2. There is a lint Warning in generator.js 92:3: How do you want me to refactor this?

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -21,6 +21,16 @@ npm run lint
 
 In case you want to quickly build a docker image locally to check if it works then run `npm run docker-build`
 
+In case you don't have a local node environment you can run 
+
+```
+docker build -t asyncapi/generator:latest .
+docker run -v `pwd`:`pwd` -w `pwd` -it --rm --entrypoint /bin/sh  asyncapi/generator:latest
+npm install
+npm run lint
+./cli.js --help
+```
+
 ## Conventional commits
 
 This project follows [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/#summary) specification. Releasing to GitHub and NPM is done with the support of [semantic-release](https://semantic-release.gitbook.io/semantic-release/).
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -27,7 +27,6 @@ In case you don't have a local node environment you can run
 docker build -t asyncapi/generator:latest .
 docker run -v `pwd`:`pwd` -w `pwd` -it --rm --entrypoint /bin/sh  asyncapi/generator:latest
 npm install
-npm run lint
 ./cli.js --help
 ```
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -14,7 +14,7 @@ Contributions are more than welcome. If you want to contribute, please make sure
 Before creating a Pull Request you should validate your code with ESLint.
 
 ```
-npm run eslint
+npm run lint
 ```
 
 ## Developing with Docker
--- a/cli.js
+++ b/cli.js
@@ -42,11 +42,16 @@ const disableHooksParser = v => {
   }
 };
 
-const mapFromBaseUrlParser = v => {
-  mapBaseUrlToFolder.url = v;
-};
-const mapToBaseFolderParser = v => {
-  mapBaseUrlToFolder.folder = v;
+const mapBaseUrlParser = v => {
+  const re = /(.*):(.*)/g;
+  let mapping = [];
+  if ((mapping = re.exec(v))===null || mapping.length!==3) {
+    throw new Error('Invalid --map-base-url flag. A mapping <url>:<baseFolder> with delimiter : expected.');
+  }
+  // folder is without trailing slash, so make sure that url has also no trailing slash:
+  mapBaseUrlToFolder.url = mapping[1].replace(/\/$/, '');
+  mapBaseUrlToFolder.folder = path.resolve(mapping[2]);
+  console.log(`mapping url ${mapBaseUrlToFolder.url} to folder ${mapBaseUrlToFolder.folder}`);
 };
 
 const showError = err => {
@@ -75,8 +80,7 @@ program
   .option('-p, --param <name=value>', 'additional param to pass to templates', paramParser)
   .option('--force-write', 'force writing of the generated files to given directory even if it is a git repo with unstaged files or not empty dir (defaults to false)')
   .option('--watch-template', 'watches the template directory and the AsyncAPI document, and re-generate the files when changes occur. Ignores the output directory. This flag should be used only for template development.')
-  .option('--mapFromBaseUrl <url>','maps all schema references with this base url to folder specified by --mapToBaseFolder',mapFromBaseUrlParser)
-  .option('--mapToBaseFolder <baseFolder>','maps all schema references starting with url specified by --mapFromBaseUrl to this folder',mapToBaseFolderParser)
+  .option('--map-base-url <url>:<baseFolder>','maps all schema references with this base url to folder',mapBaseUrlParser)
   .parse(process.argv);
 
 if (!asyncapiDocPath) {
--- a/cli.js
+++ b/cli.js
@@ -43,13 +43,14 @@ const disableHooksParser = v => {
 };
 
 const mapBaseUrlParser = v => {
-  // example value for regular expression: https://schema.example.com/crm/:./test/docs/
+  // Example value for regular expression: https://schema.example.com/crm/:./test/docs/
+  // it splits on last occurrence of : into the groups all, url and folder
   const re = /(.*):(.*)/g;
   let mapping = [];
   if ((mapping = re.exec(v))===null || mapping.length!==3) {
     throw new Error('Invalid --map-base-url flag. A mapping <url>:<baseFolder> with delimiter : expected.');
   }
-  // folder is without trailing slash, so make sure that url has also no trailing slash:
+  // Folder is without trailing slash, so make sure that url has also no trailing slash:
   mapBaseUrlToFolder.url = mapping[1].replace(/\/$/, '');
   mapBaseUrlToFolder.folder = path.resolve(mapping[2]);
 };
--- a/cli.js
+++ b/cli.js
@@ -81,7 +81,7 @@ program
   .option('-p, --param <name=value>', 'additional param to pass to templates', paramParser)
   .option('--force-write', 'force writing of the generated files to given directory even if it is a git repo with unstaged files or not empty dir (defaults to false)')
   .option('--watch-template', 'watches the template directory and the AsyncAPI document, and re-generate the files when changes occur. Ignores the output directory. This flag should be used only for template development.')
-  .option('--map-base-url <url>:<baseFolder>','maps all schema references with this base url to folder',mapBaseUrlParser)
+  .option('--map-base-url <url>:<baseFolder>','maps all schema references from base url to local folder',mapBaseUrlParser)
   .parse(process.argv);
 
 if (!asyncapiDocPath) {
--- a/cli.js
+++ b/cli.js
@@ -43,13 +43,15 @@ const disableHooksParser = v => {
 };
 
 const mapBaseUrlParser = v => {
-  const mapping = v.split(':');
-  if (mapping.length !== 2) {
+  // example value for regular expression: https://schema.example.com/crm/:./test/docs/
+  const re = /(.*):(.*)/g;
+  let mapping = [];
+  if ((mapping = re.exec(v))===null || mapping.length!==3) {
     throw new Error('Invalid --map-base-url flag. A mapping <url>:<baseFolder> with delimiter : expected.');
   }
   // folder is without trailing slash, so make sure that url has also no trailing slash:
-  mapBaseUrlToFolder.url = mapping[0].replace(/\/$/, '');
-  mapBaseUrlToFolder.folder = path.resolve(mapping[1]);
+  mapBaseUrlToFolder.url = mapping[1].replace(/\/$/, '');
+  mapBaseUrlToFolder.folder = path.resolve(mapping[2]);
 };
 
 const showError = err => {
--- a/cli.js
+++ b/cli.js
@@ -81,7 +81,7 @@ program
   .option('-p, --param <name=value>', 'additional param to pass to templates', paramParser)
   .option('--force-write', 'force writing of the generated files to given directory even if it is a git repo with unstaged files or not empty dir (defaults to false)')
   .option('--watch-template', 'watches the template directory and the AsyncAPI document, and re-generate the files when changes occur. Ignores the output directory. This flag should be used only for template development.')
-  .option('--map-base-url <url>:<baseFolder>','maps all schema references from base url to local folder',mapBaseUrlParser)
+  .option('--map-base-url <url:folder>','maps all schema references from base url to local folder',mapBaseUrlParser)
   .parse(process.argv);
 
 if (!asyncapiDocPath) {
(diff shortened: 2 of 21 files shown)


fmvilas (changes requested):
I love it! I'd only suggest removing all these `console.log` before it's merged :)

CONTRIBUTING.md
```diff
@@ -14,13 +14,23 @@ Contributions are more than welcome. If you want to contribute, please make sure
...
 
 In case you want to quickly build a docker image locally to check if it works then run `npm run docker-build`
 
+In case you don't have a local node environment you can run 
+
+```
+docker build -t asyncapi/generator:latest .
+docker run -v `pwd`:`pwd` -w `pwd` -it --rm --entrypoint /bin/sh  asyncapi/generator:latest
+npm install
+npm run lint
```
magicmatatjahu, line 30: I understand why you put `npm run lint` command, but maybe there we shouldn't write about it? `Developing with Docker` section should be point only for developing, not to action before creating PR. If you have a different opinion on this, feel free to discuss.

cli.js
```diff
@@ -41,6 +42,18 @@ const disableHooksParser = v => {
...
+const mapBaseUrlParser = v => {
+  const re = /(.*):(.*)/g;
+  let mapping = [];
+  if ((mapping = re.exec(v))===null || mapping.length!==3) {
+    throw new Error('Invalid --map-base-url flag. A mapping <url>:<baseFolder> with delimiter : expected.');
+  }
+  // folder is without trailing slash, so make sure that url has also no trailing slash:
+  mapBaseUrlToFolder.url = mapping[1].replace(/\/$/, '');
+  mapBaseUrlToFolder.folder = path.resolve(mapping[2]);
+  console.log(`mapping url ${mapBaseUrlToFolder.url} to folder ${mapBaseUrlToFolder.folder}`);
```
magicmatatjahu, line 54:
```suggestion
  const mapping = v.split(':');
  if (mapping.length !== 2) {
    throw new Error('Invalid --map-base-url flag. A mapping <url>:<baseFolder> with delimiter : expected.');
  }
  // folder is without trailing slash, so make sure that url has also no trailing slash:
  mapBaseUrlToFolder.url = mapping[0].replace(/\/$/, '');
  mapBaseUrlToFolder.folder = path.resolve(mapping[1]);
  console.log(`mapping url ${mapBaseUrlToFolder.url} to folder ${mapBaseUrlToFolder.folder}`);
```

is much simpler.

lib/generator.js
```diff
@@ -131,6 +134,42 @@ class Generator {
...
+                  resolve(data);
+                }
+              });
+            } catch (err) {
+              reject(`Error opening file "${path}"`);
+            }
+          }));
+        }
+      };
+    };
```
magicmatatjahu, line 172: Could you move this logic to separate util function inside `Generator` class? Currently with your changes, the constructor is very complex. Also util function should be called here https://github.com/asyncapi/generator/blob/5406966b73da20f6e9db111e61ed0b4c32203437/lib/generator.js#L293 like

```js
parserOptions.resolve = {resolver: this.mapBaseUrlToFolderResolver()};
```

and util function should return resolver (object) with `order` etc fields.

derberg:
Please add information about new CLI param to the readme

lib/generator.js
muenchhausen: Thanks for the omments so far 👍 
I am still working on some tests and hope to finish the next days.

derberg:
@muenchhausen please have a look at the failing tests

muenchhausen:
I did not manage to create an integration test that checks the renedered result for `test/docs/apiwithref.json`. I got some problems with some mocked libraries - even if I remove some `jest.mock`. I have the feeling, that integration testing is a sepearate issue. 

```javascript

const path = require('path');
const Generator = require('../lib/generator');

//jest.mock('../lib/filtersRegistry');
//jest.mock('../lib/hooksRegistry');

describe('Generator', () => {
  describe('#generateFromFile', () => {

    it('test', async () => {
      const gen = new Generator('@asyncapi/html-template', '/Users/z0024p1y/prototypes/generator/build', {
        mapBaseUrlToFolder: 'https://schema.example.com/crm/:./test/docs/',
        forceWrite: true
      });
      await gen.generateFromFile(path.resolve('./test/docs/apiwithref.json'));
    });
  });
});
```

magicmatatjahu:
@muenchhausen Hi! Thanks for your work! I see that you have very old revision in tests in your branch:
- master: https://github.com/asyncapi/generator/tree/master/test
- your branch: https://github.com/muenchhausen/generator/tree/map-schema/test

Please update branch and write tests like here https://github.com/asyncapi/generator/blob/master/test/integration.test.js :)

CONTRIBUTING.md
muenchhausen: I have the feeling, that a simple quick start for contribution is helpful. Not all contributers are node developers. Where would be a better location in this project?

cli.js
muenchhausen: Thanks - I took over your suggestion!

lib/generator.js
```diff
@@ -252,6 +255,10 @@ class Generator {
   async generateFromString(asyncapiString, parserOptions = {}) {
     if (!asyncapiString || typeof asyncapiString !== 'string') throw new Error('Parameter "asyncapiString" must be a non-empty string.');
 
+    if (Object.keys(this.mapBaseUrlToFolder).length > 0) {
```
magicmatatjahu, line 258: This `if` will be always true if someone use CLI, because in `cli.js` we have 

```js
const mapBaseUrlToFolder = { url: '', folder: process.cwd()};
```

Could you change it to:

```suggestion
    if (this.mapBaseUrlToFolder.url) {
```
? Then we have sure that someone pass url.

lib/generator.js
```diff
@@ -835,6 +842,42 @@ class Generator {
...
   }
+
+  /**
+   * Creates a custom resolver that maps mapBaseUrlToFolder.url to mapBaseUrlToFolder.folder
+   *
+   * @private
+   * @param  {string} fileName Name of the file to check against a list of glob patterns.
+   * @return {{read(*, *, *): Promise<unknown>, canRead(*): boolean, order: number}}
+   */
+  getMapBaseUrlToFolderResolver(mapBaseUrlToFolder) {
```
magicmatatjahu, line 853: This function is inside class, so you can use `this` for `mapBaseUrlToFolder` argument :)

```js
getMapBaseUrlToFolderResolver() {
...
const baseUrl = this.mapBaseUrlToFolder.url;
const baseDir = this.mapBaseUrlToFolder.folder;
```

magicmatatjahu (changes requested):
Awesome, only two improvements :) @derberg Do you want also make review?

derberg:
We are definitely missing an update to the Readme with example usage. This is my only concern.

lib/generator.js
```diff
@@ -252,6 +255,10 @@ class Generator {
   async generateFromString(asyncapiString, parserOptions = {}) {
     if (!asyncapiString || typeof asyncapiString !== 'string') throw new Error('Parameter "asyncapiString" must be a non-empty string.');
 
+    if (Object.keys(this.mapBaseUrlToFolder).length > 0) {
+      parserOptions.resolve = {resolver: this.getMapBaseUrlToFolderResolver(this.mapBaseUrlToFolder)};
```
fmvilas, line 259: I wonder how this would conflict if we use both `mapBaseUrlToFolder` and the `parserOptions` [here](https://github.com/asyncapi/generator/blob/master/docs/api.md#Generator+generateFromString).

cli.js
muenchhausen: The regular expression is required, because the utl might have also a ':'.

lib/generator.js
muenchhausen: good eyes :) Thanks!

lib/generator.js
muenchhausen: There is a conflict, if `generateFromString` is called with a valid `resolve` configuration. Then `resolve` is overwritten by `mapBaseUrlToFolder`. I will recheck. Maybe I can move the code to `generateFromFile` and `generateFromURL`.

cli.js
magicmatatjahu: but as I understand, regular expression also will split parameter to 3 parts when someone pass url with `:`? Am I right, or not? 😅

muenchhausen: I wrote a simple test. The greedy nature of `.*` will make sure that we split correctly. Here an example:
  
``` js
   it('mapBaseUrlParser works with url mapping', async () => {
      const v = 'https:::::://schema.example.com/crm/:./test/docs/';
      const re = /(.*):(.*)/;
      const mapping = re.exec(v);
      expect(mapping.length).toBe(3);
      expect(mapping[0]).toBe(v);
      expect(mapping[1]).toBe('https:::::://schema.example.com/crm/');
      expect(mapping[2]).toBe('./test/docs/');
    });
```

lib/generator.js
muenchhausen: Conflict is solved.

muenchhausen:
I think that am done for now! Let me know, if there is anything to do.

Please squash all my commits since I had to try some parts forward and backward.

cli.js
```diff
@@ -67,6 +81,7 @@ program
   .option('-p, --param <name=value>', 'additional param to pass to templates', paramParser)
   .option('--force-write', 'force writing of the generated files to given directory even if it is a git repo with unstaged files or not empty dir (defaults to false)')
   .option('--watch-template', 'watches the template directory and the AsyncAPI document, and re-generate the files when changes occur. Ignores the output directory. This flag should be used only for template development.')
+  .option('--map-base-url <url>:<baseFolder>','maps all schema references with this base url to folder',mapBaseUrlParser)
```
derberg, line 84:
```suggestion
  .option('--map-base-url <url>:<baseFolder>','maps all schema references from base url to local folder',mapBaseUrlParser)
```

lib/generator.js
```diff
@@ -835,6 +847,43 @@ class Generator {
       if (!isDirEmpty) throw new Error(`"${this.targetDir}" is not an empty directory. You might override your work. To skip this rule, please make your code a git repository or use the --force-write flag (not recommended).`);
     }
   }
+
+  /**
+   * Creates a custom resolver that maps mapBaseUrlToFolder.url to mapBaseUrlToFolder.folder
+   *
+   * @private
+   * @param  {string} fileName Name of the file to check against a list of glob patterns.
```
derberg, line 855: this jsdoc requires an update

lib/generator.js
```diff
@@ -89,8 +89,9 @@ class Generator {
    * @param {Boolean} [options.forceWrite=false] Force writing of the generated files to given directory even if it is a git repo with unstaged files or not empty dir. Default is set to false.
    * @param {Boolean} [options.install=false] Install the template and its dependencies, even when the template has already been installed.
    * @param {Boolean} [options.debug=false] Enable more specific errors in the console. At the moment it only shows specific errors about filters. Keep in mind that as a result errors about template are less descriptive.
+   * @param {String} [options.mapBaseUrlToFolder]       Optional parameter to map schema references from URL to a local base folder e.g. https://schema.example.com/crm/:./test/docs/.. The delimiter between URL and folder is a ':'.
```
derberg, line 92: this no longer is a string, requires update

derberg (commented):
We always squash all commits into one so no worries 👍🏼 

Please extend `options` list from the readme https://github.com/asyncapi/generator#cli-usage

lib/generator.js
```diff
@@ -835,6 +847,43 @@ class Generator {
...
   }
+
+  /**
+   * Creates a custom resolver that maps mapBaseUrlToFolder.url to mapBaseUrlToFolder.folder
+   *
+   * @private
+   * @param  {string} fileName Name of the file to check against a list of glob patterns.
+   * @return {{read(*, *, *): Promise<unknown>, canRead(*): boolean, order: number}}
+   */
+  static getMapBaseUrlToFolderResolver(urlToFolder) {
```
fmvilas, line 858: I don't think this method should be here. It's not particularly specific to the Generator class. It could actually be a separate package or a helper. Anyway, not a blocker for me, just my two cents.

fmvilas (approved):

lib/generator.js
magicmatatjahu: I don't remember previous code, but this function wasn't static 😅 I agree with Fran, but also for me this isn't a blocker.

magicmatatjahu (approved):

lib/generator.js
derberg: @fmvilas @magicmatatjahu you mean it should be moved to `utils.js` ? yup, makes sense

most important are updated to docs

magicmatatjahu: Yep, as I remember this function was integrate with instance of Generator, not definition of Generator and performed logic using `this.mapBaseUrlToFolder`. With current logic, we should move it to `utils.js`, but as I said, it's not a blocker for me, someone from community can handle it in followup as `good first issue` :)

muenchhausen: I did not want to have this as member function. Moving to util is a much better solution, thanks!

muenchhausen:
I implemented all suggestions and tested this branch successfully in a company CI. From my perspective this PR might be ready to be merged.

cli.js
```diff
@@ -41,6 +42,19 @@ const disableHooksParser = v => {
   }
 };
 
+const mapBaseUrlParser = v => {
+  // Example value for regular expression: https://schema.example.com/crm/:./test/docs/
+  // it splits on last occurrence of : into the groups all, url and folder
+  const re = /(.*):(.*)/g;
+  let mapping = [];
+  if ((mapping = re.exec(v))===null || mapping.length!==3) {
```
derberg, line 50: just a comment that this statement will not work with such situation:
```bash
ag test/docs/apiwithref.json @asyncapi/html-template -o ./build/ --force-write --map-base-url https://schema.example.com/crm/
```

because `https://schema.example.com/crm/` has `:` so `const re = /(.*):(.*)/g;` returns 3 elements and therefor it contunes to work and fails when trying to resolve refs

```
Something went wrong:
Error: Error downloading https://schema.example.com/crm/shared.json 
getaddrinfo ENOTFOUND schema.example.com
    at dereference (/Users/wookiee/sources/generator/node_modules/@asyncapi/parser/lib/parser.js:155:11)
    at processTicksAndRejections (internal/process/task_queues.js:97:5)
```

muenchhausen: Thanks - good point! I fixed it. Of course it would be better to put cli.js under test.

fmvilas (approved):
🚀

derberg:
@all-contributors please add @muenchhausen for code

PR #530, opened 2021-03-09
feat: support templates installed as global npm or yarn packages

### LINKED ISSUE #482: How to use pre-install template?

### LINKED ISSUE #523: EACCESS: permission denied
#### Describe the bug
Running a simple ag command fails with a permission error

#### How to Reproduce
```
$ sudo npm -g install @asyncapi/html-template --unsafe-perm=true
...

$ sudo npm -g install @asyncapi/generator
...

$ ag --debug api_description.yml @asyncapi/html-template
Errors while trying to resolve package location at undefined TypeError [ERR_INVALID_ARG_TYPE]: The "path" argument must be of type string. Received undefined
    at validateString (internal/validators.js:124:11)
    at Object.join (path.js:1039:7)
    at /usr/local/lib/node_modules/@asyncapi/generator/lib/generator.js:353:41
    at new Promise (<anonymous>)
    at Generator.installTemplate (/usr/local/lib/node_modules/@asyncapi/generator/lib/generator.js:339:12)
    at Generator.generate (/usr/local/lib/node_modules/@asyncapi/generator/lib/generator.js:168:73)
    at processTicksAndRejections (internal/process/task_queues.js:93:5)
    at async /usr/local/lib/node_modules/@asyncapi/generator/cli.js:132:9 {
  code: 'ERR_INVALID_ARG_TYPE'
}
Template installation started because the template cannot be found on disk
Something went wrong:
Error: EACCES: permission denied, mkdir '/usr/local/lib/node_modules/@asyncapi/generator/node_modules/@asyncapi/html-template'
```

#### Expected behavior
I expected ag to generate some nice pretty HTML, not try to install files in a system location. I expected ag to find the previously installed template. I also expected that if ag did need to download anything, it would not do so into a place that needed root privileges.

derberg:
**Description**

- if the template cannot be found in local dependencies, then we will try to look for it among globally installed templates 
- we look for in yarn and later in npm global locations
- I moved some logic to utils as generator.js would get more unnecessary complexity. In utils, I added a lot of comments so this beast is easier to maintain by others
- I do not use `resolve-global` and `resolveg` (not mature enough and not released for some time already). They both have the same issue - they fail on our templates as templates do not have entrypoints, this is why I use 2 different dependencies, to get global location (globalDirs) of packages and resolve packages using local paths (resolveFrom)
- TODO: integration testing is a huge challenge, I think that we just need to accept that test-project integration tests can run only in an isolated CI environment
- TODO: update readme

**Related issue(s)**
Resolves #482 
Resolves #523


lib/utils.js
```diff
@@ -146,3 +150,57 @@ utils.registerSourceMap = () => {
...
+        pkgPackageJsonPath = resolveFrom(globalDirs.yarn.packages, path.join(name, PACKAGE_JSON_FILENAME));
+      } catch (_) {
+        //trying to find in global npm location
+        pkgPackageJsonPath = resolveFrom(globalDirs.npm.packages, path.join(name, PACKAGE_JSON_FILENAME));
+      }
+
+      if (pkgPackageJsonPath) {
+        log.debug(`${name} not found in local dependencies but found it installed as a global package.`);
+        //removing package.json from path
+        pkgPath = pkgPackageJsonPath.substring(0, pkgPackageJsonPath.lastIndexOf('/'));
```
laat, line 193: for cross platform support it's better to use `path.dirname(pkgPackageJsonPath
)`.

laat (commented):
I like this solution. 👍

I'm also surprised that there is no a package `resolve-pkg-global` that does this resolution already. I looked, and found none.

lib/utils.js
derberg: love it mate, thanks! fixed

derberg:
@laat thanks for looking into this. I was also surprised that the assumption is that every package has an entrypoint 🤷🏼  one day maybe I extract it and release it as a standalone package 🤞🏼 maybe...

I'm still thinking about what to do to make test-project integration tests work local too. Testing global package usage locally is not possible without messing with developers setup. Pushing to CI to test integration is not a nice dev experience. Thinking about docker

test/test-project/README.md
```diff
@@ -1,4 +1,10 @@
...
\ No newline at end of file
+The version of the html-template must be hardcoded to version `0.16.0` because tests check if output is generated using this given version of the template, that newer version is not downloaded even though `0.16.0` was installed with `npm install` first.
+
+There are two custom test scripts:
+- `npm test:project` starts library integration tests that can run the same way both on local and CI
+- `npm test:global` always fail on local as it tests installation of template under global location. This would mean you have to install template globally before running the test. It means messing with your developer setup. Instead you have a docker compose configuration that you can use to run all tests from the test project in an isolated environment
+
+Instead of running tests with `npm test`, make sure you have Docker Compose and run `TAG=14 docker-compose up`. `TAG` is the version of Node image that you want to use in test.
```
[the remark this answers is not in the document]
magicmatatjahu: `IMAGE_TAG` is better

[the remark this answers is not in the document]
derberg: `NODE_IMAGE_TAG`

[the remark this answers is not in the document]
magicmatatjahu: image

.github/workflows/pr-testing-with-test-project.yml
```diff
@@ -9,22 +9,13 @@ jobs:
   test:
     if: github.event.pull_request.draft == false
     name: Test generator as dependency with Node ${{ matrix.node }}
-    runs-on: macos-latest
+    runs-on: ubuntu-latest
```
jonaslagoni, line 12: Is there a reason why we dont use `matrix`?
```
    name: ${{ matrix.os }}
    runs-on: ${{ matrix.os }}
    strategy:
      matrix:
        os: [ubuntu-latest, macos-latest, windows-latest]
```

test/test-project/test.sh
```diff
@@ -0,0 +1,15 @@
...
+npm install
+cd test/test-project
+npm install
+npm run test:project
+#installing html template globally before global tests
+PUPPETEER_SKIP_DOWNLOAD=true npm install -g @asyncapi/html-template@0.16.0
+#remove previously installed html template to make sure it is not picked up in the test
+rm -rf node_modules/@asyncapi/html-template
+rm -rf ../../node_modules/@asyncapi/html-template
+npm run test:global
```
jonaslagoni, line 15: Since we are using Docker-compose wouldn't it be fitting to wrap this in a dockerfile instead 🤔

jonaslagoni (changes requested):
good job, only had a few questions and suggestions 👍

lib/utils.js
```diff
@@ -146,3 +150,57 @@ utils.registerSourceMap = () => {
...
+
+    //no we need to look for the template in global location
+    //we use resolveFrom with package.json reference as normal resolve fails on template because they do not have entrypoint ("main" property) as packages normally do.
+    if (!pkgPath) {
+      try {
+        //trying to find in global yarn location
+        pkgPackageJsonPath = resolveFrom(globalDirs.yarn.packages, path.join(name, PACKAGE_JSON_FILENAME));
+      } catch (_) {
+        //trying to find in global npm location
+        pkgPackageJsonPath = resolveFrom(globalDirs.npm.packages, path.join(name, PACKAGE_JSON_FILENAME));
```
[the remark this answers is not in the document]
magicmatatjahu: Also someone can use https://pnpm.js.org/ and I don't know how it is similar to `npm` and does it use this same global dir like `npm` ... 😞

test/test-project/test.sh
magicmatatjahu: @jonaslagoni  You can also run it locally as I understand from new section in docs :)

test/test-project/test.sh
```diff
@@ -0,0 +1,15 @@
...
+sudo chown -R 1001:121 "/root/.npm"
+cd app
+npm install
+cd test/test-project
+npm install
+npm run test:project
+#installing html template globally before global tests
+PUPPETEER_SKIP_DOWNLOAD=true npm install -g @asyncapi/html-template@0.16.0
+#remove previously installed html template to make sure it is not picked up in the test
+rm -rf node_modules/@asyncapi/html-template
```
magicmatatjahu, line 13:
```suggestion
rm -rf ./node_modules/@asyncapi/html-template
```

As I remember some commands (I don't remember if `rm` also) treat path without `./` prefix as global path.

test/utils.test.js
```diff
@@ -0,0 +1,62 @@
...
+      });
+    });
+
+    it('works with an npm package', () => {
+      log.debug = jest.fn();
+      utils.isFileSystemPath = jest.fn(() => false);
+      const packagePath = path.join(Generator.DEFAULT_TEMPLATES_DIR, templateNpmName);
+      resolvePkg.__resolvePkgValue = packagePath;
+      const result = utils.getTemplateDetails(templateNpmName, 'package.json');
+      expect(log.debug).not.toHaveBeenCalledWith('Remember that your local template must have its own node_modules installed first, \"npm install\" is not triggered by the generator.');
```
magicmatatjahu, line 37: Not for this PR, but we should move all texts hardcoded in code to separate file. I mean, by changing the text in code you have to change the tests immediately, and this a problem for me.

.github/workflows/pr-testing-with-test-project.yml
derberg: you mean why we do not test on all Os? I just did not want to extend testing time to much, this is why the other tests test on all OS, while these integration tests test only on Linux but different node versions

and only ubuntu, for now, has docker-compose preinstalled (but this is minor stuff)

lib/utils.js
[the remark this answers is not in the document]
derberg: @jonaslagoni it is not needed to catch it again as this util is already used in try-catch where we do not want to throw an error but just go into the installation

@magicmatatjahu this is why this PR claims support for npm and yarn only. If `pnpm` users will create an issue that generator doesn't work well with their pnpm packages, we can try solving or asking them to solve 🤷🏼‍♂️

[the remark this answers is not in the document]
laat: pnpm is strict, and this is an unlikely scenario for those users. They should declare all dependencies in package.json and not depend on global packages. They do not support require of global packages.
https://github.com/pnpm/pnpm/issues/1210#issuecomment-683350153

test/test-project/test.sh
derberg: sure we can use Dockerfile and instead of `docker-compose up` we could do `docker build`

I choose docker compose:
- because I kinda prefer to read docker compose config file instead of Dockerfile
- in `docker-compose` stored under `test-project` I could map volumes like `../../:/app` and in Dockerfile it would not work because of context related issues and Dockerfile would have to be on the root of the project. I felt that this is just weird that all `test-project` related files are in the related directory, except of this one single Dockerfile

test/utils.test.js
derberg: Love the idea https://github.com/asyncapi/generator/issues/533

test/test-project/test.sh
derberg: @magicmatatjahu you want me to check it or you already did and this is why you made a code change suggestion 😃

lib/utils.js
```diff
@@ -146,3 +150,57 @@ utils.registerSourceMap = () => {
...
   require('ts-node').register();
 };
+
+/**
+ * Check if given template is installed and return details about it
+ * 
+ * @private
+ * @param {String} name name of the template
+ * @param {String} PACKAGE_JSON_FILENAME standard name of the package.json file
+ * @returns {Object|undefined} object with fields like pkgPath, name, version
```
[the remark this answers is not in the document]
derberg: I just clarified in JSDoc that it is package.json as this should clarify I guess
all good?

derberg:
@jonaslagoni @magicmatatjahu ready for another review round

lib/utils.js
```diff
@@ -146,3 +150,56 @@ utils.registerSourceMap = () => {
...
   require('ts-node').register();
 };
+
+/**
+ * Check if given template is installed and return details about it
+ * 
+ * @private
+ * @param {String} name name of the template
+ * @param {String} PACKAGE_JSON_FILENAME standard name of the package.json file
+ * @returns {Object|undefined} object representing package.json file with additional field -> pkgPath
```
magicmatatjahu, line 160:
```suggestion
 * @returns {Object|undefined} object representing content of the package.json file with additional field -> pkgPath
```

Please add info that object represents content of package.json to clarify :)

lib/utils.js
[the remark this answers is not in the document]
jonaslagoni: Yea it was more to change the type from `Object` to an actual definition of what is returned, but yes, I think it is somewhat "impossible" to do 🤔

lib/utils.js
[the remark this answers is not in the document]
jonaslagoni: > @jonaslagoni it is not needed to catch it again as this util is already used in try-catch where we do not want to throw an error but just go into the installation

But if line 187 throws an error, no try will catch the exception and the generator will terminate unexpectedly right?

.github/workflows/pr-testing-with-test-project.yml
jonaslagoni: Fair enough 😄

test/test-project/test.sh
jonaslagoni: Yea, I guess, the only reason why I don't like the script (even if it should be run locally) is that it does not force a clean environment but I see your point 👍

test/test-project/test.sh
derberg: so I guess I will not change it if it works this way? 😛

lib/utils.js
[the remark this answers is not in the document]
derberg: it is not catch here, but it is catch in generator.js code, where this until function is used. And there is this custom log `Unable to resolve template location at ...blabla`

I can refactor but will it change anything?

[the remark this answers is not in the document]
jonaslagoni: Ahh okay, then it's all good :+1:

jonaslagoni (dismissed):
Awesome 👍

jonaslagoni:
@derberg your PR testing check failed 😄

magicmatatjahu (dismissed):
🥳

lib/utils.js
```diff
@@ -146,3 +150,56 @@ utils.registerSourceMap = () => {
...
   require('ts-node').register();
 };
+
+/**
+ * Check if given template is installed and return details about it
+ * 
+ * @private
+ * @param {String} name name of the template
+ * @param {String} PACKAGE_JSON_FILENAME standard name of the package.json file
+ * @returns {Object|undefined} object representing content of the package.json file with additional field -> pkgPath (absolute path to the package)
```
derberg, line 160: @jonaslagoni @magicmatatjahu if you feel it is important for JSDoc to properly specify package.json object, I found https://github.com/sindresorhus/type-fest/blob/main/source/package-json.d.ts. Maybe there is a way to somehow link both worlds? I can research

magicmatatjahu: I think description tell what exactly is inside the object.

derberg:
@jonaslagoni damn timeouts. I think this is something we will have to live with cause, in the end, during testing sometimes we fetch the template from npm 🤷🏼 rerun is the only fix I have in my mind for now and for this PR

lib/utils.js
jonaslagoni: The problem is you would need to know what properties are part of the `package.json` file. From the maintainer's point of view, I would rather not have to go to some docs to find the correct properties 😄

jonaslagoni: By using those types you would be able to write something like this: 
```
/**
 * @typedef {Object} ResolveType
 * @prop {string} pkgPath absolute path to the package
 */

 * @returns {PackageJson & ResolveType | undefined} object representing content of the package.json file with additional field -> pkgPath (absolute path to the package)
 
```

derberg: @jonaslagoni is this what you're looking for:

```
/**
 * @typedef {Object} ResolveType
 * @prop {string} pkgPath absolute path to the package
 */

/**
 * Check if given template is installed and return details about it located in the package.json
 * 
 * @private 
 * @param {String} name name of the template
 * @param {String} PACKAGE_JSON_FILENAME standard name of the package.json file
 * @returns {ResolveType & PackageJson | undefined} object representing content of the package.json file with pkgPath property. Read {@link https://raw.githubusercontent.com/sindresorhus/type-fest/main/source/package-json.d.ts|PackageJson} for more detaisl on contents of the package.json object
 */
```

Screenshot 2021-03-12 at 12 42 53

`PackageJson` is still `any`. I would have to add type-fest to dependencies or copy-paste the type to the generator I think. Do we really want it?

jonaslagoni: 🤷 I recently had to add the parser types to ts-nats-template because I wanted to have the actual types accessible. Meaning I had to add the parser as a dependency. Not optimal but in my case, I found it necessary to improve maintainability. If you want to add it is up to you 😄 There are pro and cons for both sides.

derberg: well, if it would be up to me, I would just do what I posted above as I believe it is enough 😄 but I'm happy to do more if you feel it is important. I just think your use case with parser is different. In the template, you operate on parser API a lot and I see a lot of value in importing types definition, but it is much different here in the generator with package.json, we need it only for the name and version and that is it

jonaslagoni: Yep totally fair :) Up to you if you want `ResolveType` but if you don't use `PackageJson` I would remove it and add something like `Object` 😄

magicmatatjahu: the problem with `PackageJSON` is that it's a TS type, so most likely JSDoc won't treat it as `typedef`

derberg: @jonaslagoni @magicmatatjahu ok, I made just a small improvement then. Please approve again

lib/utils.js
```diff
@@ -152,12 +152,12 @@ utils.registerTypeScript = () => {
...
 
 /**
- * Check if given template is installed and return details about it
+ * Check if given template is installed and return details about it located in the package.json
  * 
  * @private
  * @param {String} name name of the template
  * @param {String} PACKAGE_JSON_FILENAME standard name of the package.json file
- * @returns {Object|undefined} object representing content of the package.json file with additional field -> pkgPath (absolute path to the package)
+ * @returns {Object|undefined} object representing content of the package.json file with additional field -> pkgPath (absolute path to the package). Read {@link https://raw.githubusercontent.com/sindresorhus/type-fest/main/source/package-json.d.ts|PackageJson} for more detaisl on contents of the package.json object
```
jonaslagoni, line 160:
```suggestion
 * @returns {Object|undefined} object representing content of the package.json file with additional field -> pkgPath (absolute path to the package). Read {@link https://raw.githubusercontent.com/sindresorhus/type-fest/main/source/package-json.d.ts|PackageJson} for more details on the content of the package.json object

```

jonaslagoni (changes requested):

lib/utils.js
```diff
@@ -146,3 +150,56 @@ utils.registerSourceMap = () => {
...
   require('ts-node').register();
 };
+
+/**
+ * Check if given template is installed and return details about it located in the package.json
+ * 
+ * @private
+ * @param {String} name name of the template
+ * @param {String} PACKAGE_JSON_FILENAME standard name of the package.json file
+ * @returns {Object|undefined} object representing content of the package.json file with additional field -> pkgPath (absolute path to the package). Read {@link https://raw.githubusercontent.com/sindresorhus/type-fest/main/source/package-json.d.ts|PackageJson} for more detaisl on contents of the package.json object
```
magicmatatjahu, line 160:
```suggestion
 * @returns {Object|undefined} object representing content of the package.json file with additional field -> pkgPath (absolute path to the package). Read {@link https://raw.githubusercontent.com/sindresorhus/type-fest/main/source/package-json.d.ts|PackageJson} for more details on contents of the package.json object
```

lib/utils.js
magicmatatjahu: 🤣

derberg: no comment 🤦🏼

jonaslagoni (approved):
It is only Friday once a week 😆 LGTM 👍

magicmatatjahu (approved):

PR #536, opened 2021-03-12
docs(dummy): add specification extensions

### LINKED ISSUE #535: Add specification extensions sample in the dummy.yml
#### Reason/Context
Please try answering few of those questions
- Why we need this improvement?
The generated html doc not showing the specification extension block
- How will this change help?
Show how it is rendering
- What is the motivation?
Trying to work on the bindinds display, i'm not able to easily see how the actual  specification extension block works

#### Description
Please try answering few of those questions
- What changes have to be introduced?
just few extensions in the dummy specification to illustration primitive, object and array type
- Will this be a breaking change?
- no
- How could it be implemented/designed?

M3lkior:
Add some specification extensions samples to illustrate the extension rendering

Close #535

Here is the result : 
image

### DIFF
--- a/test/__snapshots__/integration.test.js.snap
+++ b/test/__snapshots__/integration.test.js.snap
@@ -2182,7 +2182,7 @@ object
 >
   <div class=\\"\\">
     <div
-      class=\\" flex property\\"
+      class=\\"js-prop cursor-pointer py-2 flex property\\"
     >
       <div class=\\"pr-4\\" style=\\"margin-top: -2px; min-width: 25%\\">
         <span
@@ -2190,7 +2190,20 @@ object
           style=\\"word-break: break-word\\"
           >prop1</span
         >
-            
+          
+        <svg
+          class=\\"expand inline align-baseline\\"
+          version=\\"1.1\\"
+          viewBox=\\"0 0 24 24\\"
+          x=\\"0\\"
+          xmlns=\\"http://www.w3.org/2000/svg\\"
+          y=\\"0\\"
+        >
+          <polygon
+            points=\\"17.3 8.3 12 13.6 6.7 8.3 5.3 9.7 12 16.4 18.7 9.7 \\"
+          ></polygon>
+        </svg>
+          
       </div>
 
       <div>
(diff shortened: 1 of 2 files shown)


derberg (dismissed):

derberg:
@M3lkior you will need to update the test snapshot so it contains those new properties

M3lkior:
> @M3lkior you will need to update the test snapshot so it contains those new properties

Hum ; it seems hard to maintain and apparently other missing HTML block are missing in the snapshot test from previous dummy evolution

Example: 
image

derberg:
@M3lkior what changed in the UI of generated HTML after you added the extensions to dummy example. Did you try manually? afaik extensions are "expandable" which means the svg button for expand/collapse must be added, or?

M3lkior:
> @M3lkior what changed in the UI of generated HTML after you added the extensions to dummy example. Did you try manually? afaik extensions are "expandable" which means the svg button for expand/collapse must be added, or?

yes this is like my screenshot above.

I'm not familiar with Jest Snapshot but i will try to fix the tests

derberg:
snapshot is basically a copy of whatever you snapshot (js, HTML, css, whatever text file), so in case of this test, we generate HTML template and "copy" it and store initial HTML, and then in test we compare newly generated HTML against stored snapshot. When you add something to the dummy spec it basically means the the output of HTML template also changes, and test fails, so you just need to update the snapshot. IDE should give you an option to automatically do it, like update snapshot, if you have jest plugin I guess. Other solution is to do it manually, or just remove current snapshot and let tests generate it from scratch

M3lkior:
> snapshot is basically a copy of whatever you snapshot (js, HTML, css, whatever text file), so in case of this test, we generate HTML template and "copy" it and store initial HTML, and then in test we compare newly generated HTML against stored snapshot. When you add something to the dummy spec it basically means the the output of HTML template also changes, and test fails, so you just need to update the snapshot. IDE should give you an option to automatically do it, like update snapshot, if you have jest plugin I guess. Other solution is to do it manually, or just remove current snapshot and let tests generate it from scratch

done @derberg  🤟

derberg (approved):

PR #609, opened 2021-06-28
feat: use latest parser and support spec 2.1 validation

derberg:
so far it is just parser 1.6 because 1.7 is not yet released, changes in dummy.yml are for now commented out

### DIFF
--- a/test/docs/dummy.yml
+++ b/test/docs/dummy.yml
@@ -188,14 +188,14 @@ components:
       payload:
         $ref: "#/components/schemas/dummyInfo"
       examples:
-        - #name: option1example
-          #summary: this is dummy summary for option1example
+        - name: option1example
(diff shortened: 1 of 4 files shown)


magicmatatjahu (dismissed):
LGTM

test/docs/dummy.yml
```diff
@@ -188,12 +188,15 @@ components:
       payload:
         $ref: "#/components/schemas/dummyInfo"
       examples:
-        - headers:
+        - #name: option1example
```
magicmatatjahu, line 191: It should be uncommented also another comments.

derberg: omg!

magicmatatjahu (approved):

magicmatatjahu (approved):
LGTM 🚀

PR #613, opened 2021-07-09
test(dummy): extend dummy example with channel extension and complex schema

derberg:
- added example extension on channel level
- added more complex schema that was added recently to streetlight dummy example in react component


magicmatatjahu:
I will accept when new version of html-template will be released to keep the snapshot up to date :)

derberg:
but we are not using latest html template, we are fixed to old version, otherwise it would be a huge pain with snapshots

magicmatatjahu (approved):

PR #631, opened 2021-08-20
fix: move definition of asyncapi to code path of all generate* methods.

btimby:
**Description**

As things stand before my change:
- The `generate()` method does not set `this.asyncapi`
- `generateFromString()` sets `this.asyncapi` then calls `generate()`
- `generateFromUrl()` and `generateFromFile()` both call `generateFromString()`

Therefore, `generate()` cannot function when called directly, yet the other methods work by virtue of `generateFromString()`. I simply moved the definition of `this.asyncapi` to the `generate()` method as it is the root of the other methods.

**Related issue(s)**

https://github.com/asyncapi/html-template/issues/252

### DIFF
--- a/lib/generator.js
+++ b/lib/generator.js
@@ -166,6 +166,8 @@ class Generator {
   async generate(asyncapiDocument) {
     if (!(asyncapiDocument instanceof AsyncAPIDocument)) throw new Error('Parameter "asyncapiDocument" must be an AsyncAPIDocument object.');
 
+    this.asyncapi = asyncapiDocument;
+
     if (this.output === 'fs') {
       xfs.mkdirpSync(this.targetDir);
       if (!this.forceWrite) await this.verifyTargetDir(this.targetDir);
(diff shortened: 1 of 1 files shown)


magicmatatjahu:
@btimby Hello! Thanks for contribution! From code I see that it should works, but did you test generator witch changes locally for template?

btimby:
@magicmatatjahu Yes, in fact I am using my fork / branch in a project and it is working as intended.

btimby:
@magicmatatjahu I also tested the change using a test script before I finished with the PR. So my simplified test case (provided in the associated issue), and my larger project are working now.

btimby:
For reference, here is my issue (in html-template repository) and discussion about the problem.

https://github.com/asyncapi/html-template/issues/252

magicmatatjahu (approved):
LGTM! @btimby Thanks for PR! @derberg Do you want to add something?

derberg:
@magicmatatjahu my only question is, is it really a fix and not feat. It is not really a bug, it worked differently and now it is just improved 🤔

magicmatatjahu:
@derberg Changed to feat :)

btimby:
IMO, when calling a public method of an class, one would expect it do what the documention describes rather than throw an exception due to an undefined property.

This was clearly unintended behavior.

derberg:
@btimby makes sense, you're right. @magicmatatjahu 😉

PR #664, opened 2021-09-29
fix: bump latest sdk and fix tests

derberg:
Fix long hanging issue with these tests https://github.com/asyncapi/generator/pull/663

It started when we introduced autobumping of dependant repos for html-template. So it ran here https://github.com/asyncapi/html-template/runs/3721859351?check_suite_focus=true and as a result did this https://github.com/asyncapi/generator/pull/659

Problem is that tests in generator depend on HTML 0.16 and that should never be bumped

I'm not 100% sure how to prevent it in future, just checking it this was really the core of the problem and tests pass now


magicmatatjahu:
I cannot understand one thing, why we should be fixed with `0.16.X` version of `html-template`? Why not base tests on some newest version?

derberg:
> I cannot understand one thing, why we should be fixed with 0.16.X version of html-template? Why not base tests on some newest version?

opened for ideas, just newest will at some point become old too. I choose this way as in version 0.16.x we had no support for correlationID rendering so I can run integration tests that assure we pick up proper version of the template, with or without it. You know what I mean?

derberg:
@magicmatatjahu integration test worked. I see only one solution, push html-template to node_modules to this repo in test-project directory and remove it from package.json. Thoughts?

magicmatatjahu:
@derberg I understand and I know that it can be problem to bump always the html-template, but we can end with bumping only with new version of spec. What do you think?

>  I see only one solution, push html-template to node_modules to this repo in test-project directory and remove it from package.json. Thoughts?

And isn't that how it currently worked? It was bumping up versions in test-directory and not on root?

derberg:
> but we can end with bumping only when new version of spec will be released

sorry?

> And isn't that how it currently worked? It was bumping up versions in test-directory and not on root?

yes, but now we would remove in from package.json and just push dependency to node_modules to the repo, so it is always there, bump-proof

magicmatatjahu:
@derberg I meant to bump html-template versions only when a new version of AsyncAPI is released, e.g. recently 2.2.0 - bump to the latest html-template version (0.23.9) etc.

derberg:
@magicmatatjahu I'm afraid it  adds lots of complexity, how to modify bump workflow + update of tests on each release - but we do not really need it for these integration tests. They test "template installation" and are asyncapi-version-agnostic

magicmatatjahu:
Ok, if only it tests template installation, then we can go with `0.16.0` :)

magicmatatjahu (approved):

test/test-project/package.json
```diff
@@ -8,7 +8,7 @@
     },
     "dependencies": {
         "@asyncapi/generator": "file:../..",
-        "@asyncapi/html-template": "^0.23.8"
+        "@asyncapi/html-template": "^0.16.0"
```
magicmatatjahu, line 11: I think that it should be fixed without `^`

```suggestion
        "@asyncapi/html-template": "0.16.0"
```

magicmatatjahu (changes requested):

test/test-project/package.json
derberg: but bump workflow with bump it anyway

magicmatatjahu (approved):

PR #673, opened 2021-10-06
feat: :sparkles: added support for using current npm registry

### LINKED ISSUE #538: Generator doesn't pull template from private repo
#### Describe the bug
After upgrade to 1.6.0 version from 1.1.4, ag stopped to recognize npm configuration to use private repository

#### How to Reproduce
```javascript
install @asyncapi/generator@1.6.0

configure NPM to use private repository using:
npm config set registry <path to registry>

try to generate code
ag ./server.yaml @repo/template --force-write --output ./out

see error - Error: 404 Not Found - GET https://registry.npmjs.org/<template in private repo> - Not found

install @asyncapi/generator@1.1.4
ag ./server.yaml @repo/template --force-write --output ./out
Done! ✨

```
#### Expected behavior
Ag should pull template from private repository

rluvaton:
**Description**
Use the current NPM registry

**Related issue(s)**
Fixes #538

### DIFF
--- a/lib/generator.js
+++ b/lib/generator.js
@@ -16,6 +16,7 @@ const log = require('loglevel');
 const Arborist = require('@npmcli/arborist');
 const { validateTemplateConfig } = require('./templateConfigValidator');
 const {
+  getCurrentNpmRegistry,
   convertMapToObject,
   isFileSystemPath,
   readFile,
@@ -386,7 +387,8 @@ class Generator {
       if (isFileSystemPath(this.templateName)) log.debug(logMessage.NPM_INSTALL_TRIGGER);
 
       const arb = new Arborist({
-        path: ROOT_DIR
+        path: ROOT_DIR,
+        registry: getCurrentNpmRegistry()
       });
 
       try {
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -1,3 +1,4 @@
+const npmConfigReader = require('libnpmconfig');
 const fs = require('fs');
 const util = require('util');
 const path = require('path');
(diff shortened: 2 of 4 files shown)


rluvaton:
> > I would be happy if you could add the hacktoberfest-accepted label to this PR
> 
> 
> 
> is it needed if after all the whole repo has `hacktoberfest` topi?
> 
I haven't saw that, my bad, it's not needed than.


> There are 2 challenges I have with proposed solution:
> 
> - `libnpmconfig` is an archived project and one of its dependencies probably won't work with npm7, and even if it will, maintainers are pretty clear about [the library status](https://www.npmjs.com/package/figgy-pudding). This is not a critical package, but if its dependencies already have mixed status than long term, any security updates will cause lots of issues 😞 

Yeah I saw and I thought about it too but the library is not deprecated so instead of running child process with `npm config get registry` (which I'm trying to avoid) I thought it's better to do this instead, but I'm on the fence as there is also [npm/config](https://github.com/npm/config) package that NPM use it in their CLI code but it seems like it only getting you the config at specific place

> - we now use Arborist, work with npm programmatically without using npm installation, and this solution basically expects you have npm installed as your npm config is read. Why do you think it is better than enabling authorization on library API level?

I don't really understand what you  mean, Artborist doesn't use the currently used npm registry

derberg:
@rluvaton the reason we switched to Arborist as it is the most latest and modern approach to manage NPM modules installation. We had too many issues with reusing NPM clients. Thus my request is for us to use Arborist for private/custom registries. Using the config file could work as a fallback

rluvaton:
😀 so what we need to do is add those flags: `--registry`, `--username`, `--password`, `--auth-token`

jonaslagoni:
@zmt-Eason what do you need to continue here?

rluvaton:
I'm sorry, crazy times, if someone wants she/he can continue this PR...

PR #677, opened 2021-10-19
fix: audit dependencies

albertodiazdorado:
Fix: Audit dependency tree to remove well-known vulnerabilities

**Description:** Running `npm audit` for this project reveals a series of easily fixable vulnerabilities in the dependency tree. Many of them are automatically fixable running `npm audit fix`. Moreover, I have upgraded the devDependencies `jest` and `eslint` by some major versions, which also removes quite many vulnerabilities. I have also run the tests and the linter and nothing breaks. I have also opened the test results in my browser and everything looks fine, via `firefox test/output/index.html`

---

This package is great, but I get tons of `npm` warnings whenever I use them. I would like to receive just a few less, and fixing the issues in this PR was rather easy. Do I need to do something else so that this PR is compliant?

### DIFF
--- a/lib/__mocks__/utils.js
+++ b/lib/__mocks__/utils.js
@@ -1,5 +1,5 @@
 const utils = jest.genMockFromModule('../utils');
-const { getInvalidOptions } = require.requireActual('../utils');
+const { getInvalidOptions } = jest.requireActual('../utils');
 
 utils.getInvalidOptions = getInvalidOptions;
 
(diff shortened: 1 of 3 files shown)


derberg:
@albertodiazdorado hey, sorry but we do not manually bump version of the package, can you remove this part and resolve conflicts?

albertodiazdorado:
Hi @derberg , because fixing issues in `package-lock.json` is never an easy thing to do, I just created a new PR branching out from the newest `master`. Please have a look at it when you have time :)

https://github.com/asyncapi/generator/pull/708

PR #708, opened 2021-11-22
fix: audit and upgrade dependencies

albertodiazdorado:
Fix: Audit dependency tree to remove well-known vulnerabilities

**Description:** Running `npm audit` for this project reveals a series of easily fixable vulnerabilities in the dependency tree. Many of them are automatically fixable running `npm audit fix`. Moreover, I have upgraded the devDependencies `jest` and `eslint` by some major versions, which also removes quite many vulnerabilities. I have also run the tests and the linter and nothing breaks. I have also opened the test results in my browser and everything looks fine, via `firefox test/output/index.html`

---

This package is great, but I get tons of `npm` warnings whenever I use them. I would like to receive just a few less, and fixing the issues in this PR was rather easy. Do I need to do something else so that this PR is compliant?

### DIFF
--- a/lib/__mocks__/utils.js
+++ b/lib/__mocks__/utils.js
@@ -1,5 +1,5 @@
 const utils = jest.genMockFromModule('../utils');
-const { getInvalidOptions } = require.requireActual('../utils');
+const { getInvalidOptions } = jest.requireActual('../utils');
 
 utils.getInvalidOptions = getInvalidOptions;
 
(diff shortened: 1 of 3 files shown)


derberg:
@albertodiazdorado doesn't look like just update of deps is enough, looks like there were some breaking changes in the API

albertodiazdorado:
Thanks for the very fast response @derberg 

There is indeed a breaking change in one of the upgrades that requires node v14. We cannot do that upgrade if we want to keep backwards compatibility with node v12.

In the end, this was not as easy as I though. I will have a second look at it. Do you want me to close the PR and re-open it in the future, or may I leave it here?

derberg:
@albertodiazdorado up to you if you want to continue with this one or start new

derberg:
> Thanks for the very fast response @derberg

yeah, the AsyncAPI Conference has ended 😅 so have more time now 😄

jonaslagoni:
@albertodiazdorado what do you need to continue here?

PR #748, opened 2022-02-07
fix: mark ts dependencies as dev

deamont66:
**Description**

- When using asyncapi/generator as a dependency it installs typescript and ts-node under the production flag. This causes final application builds to be about 100MB bigger than necessary.
- In this PR I just moved these two dependencies under devDependecies where they belong.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -6,7 +6,7 @@
   "packages": {
     "": {
       "name": "@asyncapi/generator",
-      "version": "1.8.27",
+      "version": "1.9.0",
       "license": "Apache-2.0",
       "dependencies": {
         "@asyncapi/avro-schema-parser": "^1.0.0",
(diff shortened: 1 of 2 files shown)


magicmatatjahu:
@deamont66 Hello! Thanks for the PR, but please read that part of the docs https://github.com/asyncapi/generator/blob/master/docs/authoring.md#typescript-support :) `ts` and `ts-node` dependencies is needed in the production env to have such a feature described in the mentioned section.

deamont66:
Gotcha, I missed that part. I was just going after shrinking our node modules so if it's part of your requirements I think we can close this PR :) I will try to solve it somehow on our side. Thanks for the fast reply :)

magicmatatjahu:
@deamont66

> Gotcha, I missed that part. I was just going after shrinking our node modules so if it's part of your requirements I think we can close this PR :) I will try to solve it somehow on our side. Thanks for the fast reply :)

Do you have problem with big `node_modules` or what exactly? In the next major version for generator we will try to compile (transpile to JS) the every template before pushing to the npm, so TS and TS-Node won't be needed.

PR #764, opened 2022-03-29
fix: do not register a ts-node when an instance is already registered

### LINKED ISSUE #759: Cannot use async-api generator in projects that use ts-node
#### Describe the bug

When trying to run a project that uses ts-node AND async-api generator, the following error appears.

```
Users/x/async-api-nest/node_modules/@asyncapi/generator/node_modules/ts-node/src/index.ts:513
    return new TSError(diagnosticText, diagnosticCodes)
           ^
TSError: ⨯ Unable to compile TypeScript:
src/app.module.ts(4,55): error TS2339: Property 'decorate' does not exist on type 'typeof Reflect'.
src/app.module.ts(4,92): error TS2339: Property 'decorate' does not exist on type 'typeof Reflect'.
src/app.module.ts(16,6): error TS2695: Left side of comma operator is unused and has no side effects.

    at createTSError (/Users/x/async-api-nest/node_modules/@asyncapi/generator/node_modules/ts-node/src/index.ts:513:12)
```

`ts-node` can only be registered once, otherwise it will try to compile already compiled code, potentially leading to complication errors (depending on the compiler settings).
When using `ts-node` in a project that uses async-api generator, both register `ts-node`. The project does this when starting (`node -r ts-node/register app.ts`), async-api does it here: https://github.com/asyncapi/generator/blob/808783b607850d72ed88766e581c64456400e8b5/lib/utils.js#L152

#### How to Reproduce
Steps to reproduce the issue. Attach all resources that can help us understand the issue:

- Install async-api generator
- Install ts-node
- Import async-api generator in any file
- Run the project with `node -r ts-node/register app.ts`   

https://github.com/aimed/async-api-generator-ts-node-reproduction 

#### Expected behavior

I do not expect async-api generator to modify source code at run-time AT ALL. Instead using ts-node or any other type of compilation/transpilation process should be done by the consumer application.

magicmatatjahu:
**Description**

As in title: if the ts-node has already been registered before, do not register it again.

**Related issue(s)**
Fixes #759


jonaslagoni:
The tests seem to fail 🙂

derberg (approved):
tests seem to pass after update to latest master

magicmatatjahu:
@derberg Thanks!

Issue #801, opened 2022-06-23
node and npm version issues


derberg:
Where do I even start...

We started having a problem with tests in this project. We run them at the moment on nodejs 12,14,15 and 15 is causing issues. This is easy, we can just switch to 16 and 15 is anyway not LTS.

When we switch to 16, which we should do anyway, we get an issue that `semantic-release` must be bumped because the version doesn't match the Node release. Easy, just bump the package

And then the problem starts, we get a bunch of issues related to `peerDependency` in release-related plugins as they are referring wrong versions, all that stuff. 

Now wait, why these issues did not pop up before? it is because according to my knowledge early versions of node 16 had npm 7 while the latest versions include npm 8.

I tried to solve peer dependency issues - it was good. Worked like a charm...until I thought, I'll check with Node 12...and it failed as latest `semantic-release` do not support Node 12

---

There are 2 threads in this issue, 2 different challenges:
- Node 12 is no longer maintained by Node, and **IMHO we should just drop it**....but this is bigger, we have same release and pr testing workflows everywhere, we need to agree on dropping Node 12 everywhere, and the ones that do not agree, will need to maintain their custom workflows 🤷🏼 
- I think I made a fundamental mistake in 2020 when I designed the release pipeline by managing the dependencies through the `devDependencies` in `package.json`. In my opinion to make it easier in the future, and less error-prone (I mean why do we even care about release-related dependency during testing when these packages are not relevant) **we should actually remove release-related dependencies from `package.json` and install them as a separate step in the release workflow**, fixed to a specific version. We already do it like this in Go projects -> https://github.com/asyncapi/parser-go/blob/master/.github/workflows/release.yml#L50-L51 

So yeah, when I started writing this issue, I did not think I'll end up describing 2 fundamentally different solutions for one problem....especially since in my opinion we should actually work on these 2 solutions anyway.

Thoughts?

@jonaslagoni might be also related to https://github.com/asyncapi/generator/pull/797

derberg:
@jonaslagoni @fmvilas @magicmatatjahu need your view on this one

magicmatatjahu:
> Node 12 is no longer maintained by Node, and IMHO we should just drop it....but this is bigger, we have same release and pr testing workflows everywhere, we need to agree on dropping Node 12 everywhere, and the ones that do not agree, will need to maintain their custom workflows 🤷🏼

Let's not exaggerate the fact that someone is using v12. If there are problems it will be fixed on the repository side.

[...]

PR #815, opened 2022-08-18
fix: remove deps related to release and switch to npx and non-node config

derberg:
**Description**

This PR fixes an issue with generator not able to support node 16 because of issues with release dependencies.
It also is a test of a solution on how we can do release without `pacjage.json` config and dependencies. 

If this works, changes from this repo will go to `.github` repo and then distributed to all org

- stop using `package.json` for release config and switch to `.releaserc`. This way we will configure releases the same across all repos, no matter the technology behind
- remove all release-related dependencies from `package.json` and use `semantic-release` through `npx` and install additional missing dependency through release pipeline directly
- in theory it will make future release configuration for new repos even easier as `.releaserc` could potentially be stored in `.github` repo and distributed to selected repos.

**Related issue(s)**
Fixes https://github.com/asyncapi/generator/issues/801

### DIFF
--- a/.github/workflows/if-nodejs-release.yml
+++ b/.github/workflows/if-nodejs-release.yml
@@ -81,7 +81,12 @@ jobs:
         name: Install dependencies
         run: npm install
       - if: steps.packagejson.outputs.exists == 'true'
+        name: Add plugin for conventional commits
+        run: npm install conventional-changelog-conventionalcommits
+        working-directory: ./.github
+      - if: steps.packagejson.outputs.exists == 'true'
         name: Publish to any of NPM, Github, and Docker Hub
+        working-directory: ./.github
         id: release
         env:
           GITHUB_TOKEN: ${{ secrets.GH_TOKEN }}
@@ -92,7 +97,7 @@ jobs:
           GIT_AUTHOR_EMAIL: info@asyncapi.io
           GIT_COMMITTER_NAME: asyncapi-bot
           GIT_COMMITTER_EMAIL: info@asyncapi.io
-        run: npm run release
+        run: npx semantic-release
       - if: failure() # Only, on failure, send a message on the 94_bot-failing-ci slack channel
         name: Report workflow run status to Slack
         uses: 8398a7/action-slack@v3
--- a/.github/.releaserc
+++ b/.github/.releaserc
@@ -1,0 +1,10 @@
+---
+branches:
+- master
+plugins:
+- - "@semantic-release/commit-analyzer"
+  - preset: conventionalcommits
+- - "@semantic-release/release-notes-generator"
+  - preset: conventionalcommits
+- "@semantic-release/npm"
+- "@semantic-release/github"
--- a/.github/workflows/pr-testing-with-test-project.yml
+++ b/.github/workflows/pr-testing-with-test-project.yml
@@ -12,7 +12,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        node: [ '12', '14', '15' ]
+        node: [ '12', '14', '16' ]
     steps:
       - name: Checkout repository
         uses: actions/checkout@v2
--- a/package-lock.json
+++ b/package-lock.json
@@ -42,12 +42,6 @@
         "asyncapi-generator": "cli.js"
       },
       "devDependencies": {
-        "@semantic-release/commit-analyzer": "^8.0.1",
-        "@semantic-release/github": "^7.0.4",
-        "@semantic-release/npm": "^7.0.6",
-        "@semantic-release/release-notes-generator": "^9.0.1",
-        "all-contributors-cli": "^6.14.2",
-        "conventional-changelog-conventionalcommits": "^4.4.0",
         "eslint": "^6.8.0",
         "eslint-plugin-jest": "^23.8.2",
         "eslint-plugin-sonarjs": "^0.5.0",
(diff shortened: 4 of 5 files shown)


derberg:
semi proof all is good is that CI passes, and previous PRs fail https://github.com/asyncapi/generator/pull/799

magicmatatjahu:
@derberg I cannot understand some things, please clarify them:

> stop using package.json for release config and switch to .releaserc. This way we will configure releases the same across all repos, no matter the technology behind

So that file will be distributed to all repos as I guess. I see that we use `npm`, `github` etc plugins. In `studio` and `server-api` we use also plugin for `docker`'s image publishing. How we will handle that? What if the repo needs a new plugin without changes to `.github` - currently this problem does not exist, and it seems to me that it will appear after this PR.

> remove all release-related dependencies from package.json and use semantic-release through npx and install additional missing dependency through release pipeline directly

I see in this PR that you remove all deps for plugins from `package.json` but in workflow you only install `semantic-release` without any additional plugins. Is that correct or you miss something, or I don't understand? I think workflows passed in your PR because you didn't publish - after merge we will have a problems.

Thanks!

magicmatatjahu:
@derberg I see partially answer for my first problem:

> since few releases of semantic-release, dependencies that we now explicitly add as a dependence (like @semantic-release/github ) are actually included in semantic-release. So we only need to explicitly install only conventional-changelog-conventionalcommits

https://github.com/asyncapi/.github/issues/172

but still. what about additional plugins?

derberg:
@magicmatatjahu these are very good questions!

> So that file will be distributed to all repos as I guess

yes, but only to those that are not having any special modifications. Global workflow that distributes workflows across entire organization supports "selective" approach to workflows distributions. So for example, we won't be able to put generic `.releaserc` there because it has custom release configuration that also includes upload or artifacts to the github release.

> I see that we use npm, github etc plugins. In studio and server-api we use also plugin for docker's image publishing. How we will handle that? What if the repo needs a new plugin without changes to .github - currently this problem does not exist, and it seems to me that it will appear after this PR.

Semantic release package, since few months (wasn't like that at the time the first release pipeline was designed), includes some plugins by default, like:
```
"@semantic-release/commit-analyzer"
"@semantic-release/release-notes-generator"
"@semantic-release/npm"
"@semantic-release/github"
```

So ☝🏼 do not have to be installed, as they are there in `semantic-release` by default when we run it with `npx`. We only need to explicitly install one plugin, which is `conventional-changelog-conventionalcommits`. We do it inside workflow now. It works well as we have this approach in go packages already.

For docker:
So for example in this repo, we release to docker using docker directly -> https://github.com/asyncapi/generator/blob/master/.github/workflows/release-docker.yml. It works this way as we wanted to build docker basing on artifact from NPM. Same with github action for generator.

For studio and server-api we have 2 options:
- main release workflow can always by default install `@semantic-release-plus/docker` as well. `npm run generate:docker` would have to run as part of different script, not `release`. And these projects would have to maintain their own `.releaserc` with additional config for docker plugin
- these just switch to generator and generator action approach and release to docker as a separate workflow, as a reaction to `release` event, or even `push` but still as a separate workflow. We can turn https://github.com/asyncapi/github-action-for-generator/blob/master/.github/workflows/release-docker.yml into a global workflow too

> I see in this PR that you remove all deps for plugins from package.json but in workflow you only install semantic-release without any additional plugins. Is that correct or you miss something, or I don't understand? I think workflows passed in your PR because you didn't publish - after merge we will have a problems.

so new workflow only installs `conventional-changelog-conventionalcommits` inside workflow, and in case of `semantic-release` we call it through `npx`. I explained that other plugins are there already ☝🏼 

what I just modified now was to make sure we are installing and calling specific versions of the packages

> I think workflows passed in your PR because you didn't publish - after merge we will have a problems.

yes, more important for me in 🟢  CI is that finally tests are passing normally without issues with `npm install`

magicmatatjahu:
@derberg 

> So ☝🏼 do not have to be installed, as they are there in semantic-release by default when we run it with npx. We only need to explicitly install one plugin, which is conventional-changelog-conventionalcommits. We do it inside workflow now. It works well as we have this approach in go packages already.

Thanks! I realised that when I looked in the https://github.com/asyncapi/.github/issues/172 issue. Thanks!

> For studio and server-api we have 2 options:
>
> - main release workflow can always by default install @semantic-release-plus/docker as well. npm run generate:docker would have to run as part of different script, not release. And these projects would have to maintain their own .releaserc with additional config for docker plugin
> - these just switch to generator and generator action approach and release to docker as a separate workflow, as a reaction to release event, or even push but still as a separate workflow. We can turn https://github.com/asyncapi/github-action-for-generator/blob/master/.github/workflows/release-docker.yml into a global workflow too

Probably second approach will be better. In an ideal world it would be possible to somehow connect to the release process and add plugins, but it would be too much work.

.github/workflows/if-nodejs-release.yml
```diff
@@ -80,8 +80,13 @@ jobs:
       - if: steps.packagejson.outputs.exists == 'true'
         name: Install dependencies
         run: npm install
+      - if: steps.packagejson.outputs.exists == 'true'
+        name: Add plugin for conventional commits
+        run: npm install conventional-changelog-conventionalcommits@5.0.0
+        working-directory: ./.github
       - if: steps.packagejson.outputs.exists == 'true'
         name: Publish to any of NPM, Github, and Docker Hub
+        working-directory: ./.github
```
magicmatatjahu, line 89: My concern: if we run this script in the `./.github` directory, will it then see project in the root?

derberg: yes, working dir is basically set to point where release config file is. It works like a charm with go projects -> https://github.com/asyncapi/parser-go/blob/master/.github/workflows/release.yml#L54 (I just changed location to move it out of `workflows` dir)

.github/workflows/if-nodejs-release.yml
```diff
@@ -80,8 +80,13 @@ jobs:
       - if: steps.packagejson.outputs.exists == 'true'
         name: Install dependencies
         run: npm install
+      - if: steps.packagejson.outputs.exists == 'true'
+        name: Add plugin for conventional commits
+        run: npm install conventional-changelog-conventionalcommits@5.0.0
+        working-directory: ./.github
       - if: steps.packagejson.outputs.exists == 'true'
         name: Publish to any of NPM, Github, and Docker Hub
```
magicmatatjahu, line 88:
```suggestion
        name: Publish to any of NPM, and Github
```

.github/workflows/if-nodejs-release.yml
derberg: @magicmatatjahu you are right! I assumed all is good but in go projects there is no package json.

I did local dry run and got `ENOPKG Missing `package.json` file.`

I made proper changes. There is no way to point to releaserc file with a flag, so we just have to move it to the root

magicmatatjahu (approved):
LGTM!

PR #838, opened 2022-09-21
fix: use fs.stats instead of deprecated fs.exists which uses invalid …

### LINKED ISSUE #836: Replace fs.exists with fs.stats
#### Reason/Context
fs.exists function is deprecated (see: https://nodejs.org/docs/latest-v14.x/api/fs.html#fs_fs_exists_path_callback) and it causes to unstable behaviour if used with promisify with async hooks. (see: https://github.com/open-telemetry/opentelemetry-js-contrib/issues/1185)

#### Description
We can use [fs.stat()](https://nodejs.org/docs/latest-v14.x/api/fs.html#fs_fs_stat_path_options_callback) as it is suggested in node.js official documentation.

This change won't break any functionality, we can release a patch version. I can prepare a PR if it is OK for you :)

nodify-at:
…callback args.

**Description**
Replaces fs.exists with fs.stats, wrapped utils.exists to keep the code base stable.

**Related issue(s)**

Resolves #836

### DIFF
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -18,8 +18,18 @@ utils.readlink = util.promisify(fs.readlink);
 utils.readFile = util.promisify(fs.readFile);
 utils.writeFile = util.promisify(fs.writeFile);
 utils.copyFile = util.promisify(fs.copyFile);
-utils.exists = util.promisify(fs.exists);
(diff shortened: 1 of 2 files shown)


magicmatatjahu (approved):
LGTM!

PR #867, opened 2022-11-17
docs: add patch to docs so to fix styling and code issues breaking the docs

Florence-Njeri:
**Description**
Fix styling for the generator tool docs and also breaking links on the website

**Related issue(s)**

PR #872, opened 2022-11-28
ci: add script to push latests docs from generator repo to the website

Florence-Njeri:
**Description**
Create a script to push the latest generator documentation from the generator repository to update the website/tools/generator folder.

**Related issue(s)**


.github/workflows/update-website.yml
```diff
@@ -0,0 +1,50 @@
...
+      - name: Create branch
+        working-directory: ./website
+        run: |
+          git checkout -b update-generator-docs-${{ github.sha }}
+      - name: Copy generator folder from Current Repo to Another
+        working-directory: ./website
+        run: |
+          mkdir -p ./pages/docs/tools/generator
+          cp -R ../generator/docs ./pages/docs/tools/generator
+          printf "\`\`\`\ntitle: Generator\nweight: 3\n\`\`\`" > _section.md
```
derberg, line 40:
```suggestion
          mkdir -p ./pages/docs/tools/generator
          rm ../generator/docs/README.md
          cp -R ../generator/docs ./pages/docs/tools/generator
          printf "\`\`\`\ntitle: Generator\nweight: 3\n\`\`\`" > _section.md
```

README will be removed only in a clone done in workflow, not from the upstream

.github/workflows/update-website.yml
```diff
@@ -0,0 +1,50 @@
+name: Update latest generator documentation in the website
+
+on:
+  push:
+    branches:
+      - 'master'
+    paths:
+      - 'docs/*.md'
```
derberg, line 8:
```suggestion
      - 'docs/*.md'
      - '!docs/README.md'
```

taken from https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#example-including-and-excluding-paths

derberg (changes requested):
WELL DONE 👏🏼 

I only suggest:
- renaming file from `update-website` to `update-docs-in-website` so it is 💯 clear what workflow is inside the file
- even thought `README.md` that is in docs will be ignored, for clarity, we should remove it before we copy docs to website. I left suggestion there
- I also suggest to ignore/not run workflow when push contains changes to README only

derberg:
cc @alequetzalli 

pinging you so you see that automation is just around the corner. Once @Florence-Njeri makes some small adjustments, this approach can be copied over to other repos, like CLI

magicmatatjahu:
@derberg @Florence-Njeri Did you test new docs on `website` repo? I mean, checking if we will move docs to the `docs/tools/generator` then website will work with new docs pages? if not, please do that before merging that PR.

derberg:
@magicmatatjahu good question. The answer is yes of course 😄 @Florence-Njeri tested it, build locally, played with it and as a result improved -> https://github.com/asyncapi/generator/pull/867

we will anyway not merge this PR until all the other docs related PRs are merged (there are just 2 pending)

magicmatatjahu:
@derberg Great, thanks :)

.github/workflows/update-docs-in-website.yml
```diff
@@ -37,6 +37,8 @@ jobs:
         run: |
           mkdir -p ./pages/docs/tools/generator
           cp -R ../generator/docs ./pages/docs/tools/generator
+          rm ../generator/docs/README.md
+          rm ../generator/docs/jsdoc2md-handlebars
```
derberg, line 41: @Florence-Njeri did you test it on local, you sure it is not `rm -r` or something else?

`rm ../generator/docs/README.md` is fine as you remove one file. But removal of entire folder probably requires some flag

derberg: @Florence-Njeri hey, please have a look

Florence-Njeri: Hey @derberg, sorry I missed your comment. Let me take a look

Florence-Njeri: I've tested the `rmdir` command on localhost and it does remove the entire folfer

.github/workflows/update-docs-in-website.yml
```diff
@@ -0,0 +1,52 @@
...
+        working-directory: ./website
+        run: |
+          git checkout -b update-generator-docs-${{ github.sha }}
+      - name: Copy generator folder from Current Repo to Another
+        working-directory: ./website
+        run: |
+          mkdir -p ./pages/docs/tools/generator
+          cp -R ../generator/docs ./pages/docs/tools/generator
+          rm ../generator/docs/README.md
+          rmdir ../generator/docs/jsdoc2md-handlebars
```
derberg, line 41: @Florence-Njeri I don't think it will work, according to https://linuxize.com/post/remove-directory-linux/#removing-directories-with-rmdir `rmdir` works only on empty directories, but the one that you need to delete is not empty

Florence-Njeri: I see

derberg (approved):

PR #880, opened 2022-12-07
docs: adjust api.md for the website

derberg:
@Florence-Njeri @pratik2315 like I said during our meeting yesterday, I could not let it go and had to make sure `api.md` will work in the website 😄 it was more complex that I expected 😅 but I could not imagine going public with generator docs on website without it. I wanted to make sure you can celebrate having all current docs there.

**What is this PR doing**:
- customizing handlebar templates used by jsdoc2md library:
  - `custom-sig-name.hbs` is new handlebar based on [sig-name](https://github.com/jsdoc2md/dmd/blob/00fa4fedef0ba60bfbfe87575abe3845f69c9e1a/partials/shared/signature/sig-name.hbs)
  - override version of [defaultvalue.hbs](https://github.com/jsdoc2md/dmd/blob/00fa4fedef0ba60bfbfe87575abe3845f69c9e1a/partials/shared/value/defaultvalue.hbs)
  - override version of [header.hbs](https://github.com/jsdoc2md/dmd/blob/00fa4fedef0ba60bfbfe87575abe3845f69c9e1a/partials/all-docs/docs/header.hbs)
  - override version of [link.hbs](https://github.com/jsdoc2md/dmd/blob/00fa4fedef0ba60bfbfe87575abe3845f69c9e1a/partials/shared/value/link.hbs)
  - override version of [main.hbs](https://github.com/jsdoc2md/dmd/blob/master/partials/main.hbs)
  - override version of [params-table.hbs](https://github.com/jsdoc2md/dmd/blob/00fa4fedef0ba60bfbfe87575abe3845f69c9e1a/partials/all-docs/docs/body/params/params-table.hbs)
- customizing script call for docs generation to use above diagrams as override

It was basically needed as current generated doc was failing when added to website:
- Mainly because escaped values like `<`. Now no escaping is done
- Also way too many details were done initially as headings and made generated TOC on the right ugly and not really working. Now only functions are headers
- Issue was with types that had `or` (`|`) as in website it was interpreted as table delimiter and breaking. Now parameters are not show as table but simple list
- frontmatter with short intro is added

@Florence-Njeri you will need to update your PR with script. Now not only you need to remove `README.md` but also entire folder `docs/jsdoc2md-handlebars`

This is how it looks like in website 👇🏼 

Screenshot 2022-12-07 at 15 07 24


docs/api.md
```diff
@@ -1,3 +1,10 @@
+---
```
derberg, line 1: I know it would be updated automatically after merge, but pushed it to show how it looks like now

magicmatatjahu (approved):
LGTM! :) I'm very proud of you Łukasz. A few more years and you'll be called a mid software developer 👏🏼

derberg:
> mid software developer 

will it be a drop from `senior` or a promotion from `regular` 
your message is to short to figure sarcasm 😅 😆

PR #884, opened 2022-12-13
fix: move ts to devDependencies

laat:
**Description**
Moving ts-node and typescript to devDependencies


laat:
I have no idea why the integration tests fails

derberg:
@laat looks like random timeout, I rerun it

@magicmatatjahu can you have a look at this PR please

magicmatatjahu:
TS is needed to run hooks and nunjucks's filters written in Typescript, so it is a `dependency` and not a `devDependency`. Identically with the `ts-node`.

PR #901, opened 2023-01-17
removed the releaserc

### LINKED ISSUE #900: We need to remove obsolete release config file
For release config we have 2 files:
1. https://github.com/asyncapi/generator/blob/master/.releaserc
2. https://github.com/asyncapi/generator/blob/master/.github/workflows/.releaserc

The 1st one is no longer needed. The only requirement for this task is just to remove that file from the repo

sambhavgupta0705:
**Description**
Removed the releaserc file which was not in use now

- ...
- ...
- ...

**Related issue(s)**

Resolves #900


sambhavgupta0705:
cc @derberg

PR #925, opened 2023-03-23
feat!: use latest arborist 6 and drop support for node 12 and 14

derberg:
- we are 4 versions behind arborist
- as workaround we used hidden info from arborits, but  time to change that -> https://github.com/npm/arborist/issues/243#issuecomment-855152574

### BREAKING CHANGE:

tl;dr the only breaking change is that you will no longer be able to install Generator when using Node 12 or 14. This will break pipelines for people.

We need to explicitly require:
```
  "engines": {
    "node": "^14.17.0 || ^16.13.0 || >=18.0.0"
  },
```

Because this is what latest Arborist requires.

Node 12 is already not maintained by node folks, and Node 14 support finishes on April 30, this is why I suggest to also not support Node 14.

---

TODO: this PR cannot be merged before we update global pipelines to run on Node 16. Different story though. Review and approval of this PR is not blocked by this. You only need to know that this is why tests fail

### DIFF
--- a/.github/workflows/pr-testing-with-test-project.yml
+++ b/.github/workflows/pr-testing-with-test-project.yml
@@ -12,7 +12,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        node: [ '12', '14', '16' ]
+        node: [ '16.13', '18' ]
     steps:
       - name: Checkout repository
         uses: actions/checkout@v2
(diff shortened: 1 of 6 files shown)


fmvilas (approved):
Do it!

D1X7R4:
Hi!

Is there any progress on this?

Thanks in advance.

D1X7R4:
Snyk vulnerabilities findings
Screenshot 2023-05-24 at 14 45 24

derberg:
@D1X7R4 fyi I'm dropping this PR. https://github.com/asyncapi/generator/pull/1061 should solve what you need. I found that one major version less of Arborist will not drop node 12 support - so linked PR will use it for now. Jumping from version `2.2.4` to `5.6.3` should solve many security issues

PR #1004, opened 2023-07-03
feat: build and publish docker image for arm and amd platforms

Ferror:
1. Use official docker github actions
2. Support arm and amd architectures
3. Generator version as a Dockerfile argument


Dockerfile
```diff
@@ -1,3 +1,5 @@
+ARG ASYNCAPI_GENERATOR_VERSION=0.50.0
```
derberg, line 1: why not `latest` ?

derberg (commented):
can you elaborate why you need `ASYNCAPI_GENERATOR_VERSION` if previous images for old versions are there anyway?

also, are you sure you want to invest in it? we will be deprecating `ag` as all the CLI-related dev power goes to https://github.com/asyncapi/cli. All missing features are already there in the CLI, so pretty soon we will open an issue that address these plans

Dockerfile
Ferror: It's possible that the latest version will cache and you will end up with a docker image with an invalid generator version. With an argument, you can test past versions locally and ensure that you install the right generator version. Seems like a declarative approach 😄 

PS. I also created a similar PR for CLI

derberg:
I also made deprecation plans official as we were so far just mentioning it is come issues/prs -> https://github.com/asyncapi/generator/issues/1008

.github/workflows/release-docker.yml
```diff
@@ -10,24 +10,35 @@ jobs:
...
       - name: Get version without v character
         id: version
         run: |
           VERSION=${{github.event.release.tag_name}}
           VERSION_WITHOUT_V=${VERSION:1}
-          echo "::set-output name=value::$(echo $VERSION_WITHOUT_V)"
-      - name: Release to Docker
-        run: | 
-          echo ${{secrets.DOCKER_PASSWORD}} | docker login -u ${{secrets.DOCKER_USERNAME}} --password-stdin
-          sleep 1m #docker image installs generator from npm, this sleep protects this step from any delays on npm side
```
derberg, line 26: we need this delay still in the workflow. This workflow reacts on new GitHub Release, but npm publishing is still in the process and might be that this workflow starts then package is not published yet, so `npm install` installs not the latest version really.

Ferror: So maybe instead of delay we should couple them and prevent concurrency. Just waiting is a foul's game 😄

derberg: what do you mean by `couple them`? the only option to not wait is to check with npm in loop if package is there - is it worth it?

derberg (commented):
@Ferror I saw you were force pushing something recently, but do not see any new things about delays

derberg:
@Ferror yo, should we keep it open?

Ferror:
@derberg I added the sleep. Personally, i dont think it will be required, but lets do little step. I have an idea how to handle more complex workflows. When there will be some space to refactor them then we can enter :)

.github/workflows/release-docker.yml
```diff
@@ -10,24 +10,39 @@ jobs:
...
+          push: true
+          load: false
+          build-args: |
+            ASYNCAPI_GENERATOR_VERSION=${{ steps.version.outputs.value }}
+          tags: |
+            asyncapi/generator:${{ steps.version.outputs.value }}
+            asyncapi/generator:latest
+          platforms: linux/amd64,linux/arm64
+          cache-from: type=gha
+          cache-to: type=gha
```
derberg, line 48:
```suggestion
      - name: Build Image
        uses: docker/build-push-action@v4
        with:
          push: true
          load: false
          build-args: |
            ASYNCAPI_GENERATOR_VERSION=${{ steps.version.outputs.value }}
          tags: |
            asyncapi/generator:${{ steps.version.outputs.value }}
            asyncapi/generator:latest
          platforms: linux/amd64,linux/arm64
          cache-from: type=gha
          cache-to: type=gha
      - uses: meeDamian/sync-readme@v1.0.6
         with:
           user: ${{secrets.DOCKER_USERNAME}}
           user: ${{ secrets.DOCKER_USERNAME }}
           pass: ${{ secrets.DOCKER_PASSWORD }}
           slug: asyncapi/generator
           description: Use your AsyncAPI definition to generate literally anything. Markdown documentation, Node.js code, HTML documentation, anything!
```

if you don't mind 🙏🏼

derberg (approved):
Thanks man

PR #1011, opened 2023-07-19
fix: bump dependencies and upgrade workflows

derberg:
Fixes: https://github.com/asyncapi/generator/issues/969
Also related: https://github.com/asyncapi/cli/issues/574

PS. removing the whole workflow for validation of templates list as nobody was really using it, it ran for almost 10m not but not really working. It is not a good way, we basically need all the templates to show up under asyncapi.com/tools


fmvilas (approved):

derberg:
@Florence-Njeri I need your approval as I'm removing a readme for the obsolete action

Issue #979, opened 2023-06-05
AsyncAPI spec v3 support in generator


jonaslagoni:
### Reason/Context
This Issue is used to track changes needed to support AsyncAPI v3. As a code owner, please edit this list of TODO tasks in order to properly track the progress 🙂 Once this issue is closed it means that v3 is now fully supported in this library.  

### Scope

- [ ] https://github.com/asyncapi/parser-js/issues/780
- [ ] Migrate to new JS parser API

jonaslagoni:
We need to add https://github.com/smoya/multi-parser-js as a dependency, and support a new `apiVersion` as `2` along side the rest.

derberg:
I'm all for `apiVersion` flag, main reason is to not harm adoption of future 3.x releases

How removing flag will harm future release:

- forget about v2, lets see we release v3 in September
- imagine `apiVersion` is not there
- generator supports only `Parser API v2`
- imagine there is AsciiDoc template (there actually is one, on GitLab, but imagine different)
- AsciiDoc maintainers took the effort to update the template to support spec v3.0
- Months pass, `Parser API v3` is released for some reason which means:
   - new JS Parser is released (probably major)
   - Generator new major must be released 
- Month after spec 3.1 is released
- AsciiDoc maintainers do not have time to migrate to new Generator and new Parser API
- Community do not adopt 3.1 as they prefer to workaround missing features and just use 3.0

No go for me

Unless we can assure that new Parser API in JS Parser will never mean a breaking change in library API

derberg:
Btw recording of online discussion about the generator -> https://www.youtube.com/watch?v=b4KcouHm958

jonaslagoni:
Agree 👍 

> regarding multi-parser-js - I strongly suggest it is part of the parser, not lib but parser repo, and versioned and released together with JS Parser, to simplify workflows.

It's something we can always figure out how to do at any time, so I am gonna integrate it with the generator to move this issue along.

PR #1017, opened 2023-08-01
feat: add `Parser-API` version 2 support

jonaslagoni:
**Description**
This PR adds support for the `parserApi` v2, which enables templates to work with it and effectively enables AsyncAPI v3 in templates. 

Tested with https://github.com/asyncapi/markdown-template/pull/360
Tested with https://github.com/asyncapi/html-template/pull/448

Fixes https://github.com/asyncapi/generator/issues/979


lib/parser.js
```diff
@@ -1,33 +1,47 @@
...
+    return apiVersion.substring('1');
+  }
+  return apiVersion;
+};
 
 parser.parse = (asyncapi, oldOptions, generator) => {
+  let apiVersion = this.sanitizeTemplateApiVersion(generator.templateConfig.apiVersion);
+  // Defaulting to v1 parser to convert it to the old parserAPI afterwards.
+  if (!this.usesNewAPI(generator.templateConfig)) {
+    apiVersion = '1';
```
derberg, line 21: why we default to `1`? kinda do not get `// Defaulting to v1 parser to convert it to the old parserAPI afterwards.`

why not `undefined`?

docs/configuration-file.md
```diff
@@ -28,7 +28,7 @@ The `generator` property from `package.json` file must contain a JSON object tha
 "generator":
 {
   "renderer": "react",
-  "apiVersion": "v1",
+  "apiVersion": "v2",
```
derberg, line 31: please also update test in line 11

from 
```
|`apiVersion`| String | Determines which **major** version of the [Parser-API](https://github.com/asyncapi/parser-api) the template uses. For example, `v1` for `v1.x.x`. If not specified, the Generator assumes the template is not compatible with the Parser-API so it will use the [Parser-JS v1 API](https://github.com/asyncapi/parser-js/tree/v1.18.1#api-documentation). If the template uses a version of the Parser-API that is not supported by the Generator, the Generator will throw an error.
```

to
```
|`apiVersion`| String | Determines which **major** version of the [Parser-API](https://github.com/asyncapi/parser-api) the template uses. For example, `v2` for `v2.x.x`. If not specified, the Generator assumes the template is not compatible with the Parser-API so it will use the [Parser-JS v1 API](https://github.com/asyncapi/parser-js/tree/v1.18.1#api-documentation). For templates that need to support AsyncAPI specification v3 make sure to use `v2` [Parser-API](https://github.com/asyncapi/parser-api). If the template uses a version of the Parser-API that is not supported by the Generator, the Generator will throw an error.
```

docs/generator-template.md
```diff
@@ -85,7 +85,7 @@ The **package.json** file is used to define the dependencies for your template.
   "description": "A template that generates a Python MQTT client using MQTT.",
   "generator": {
     "renderer": "react",
-    "apiVersion": "v1",
+    "apiVersion": "v2",
```
derberg, line 88: did you check the code from this tutorial? parser functions that are used are also in v2?

derberg (approved):
some minor comments

lib/parser.js
jonaslagoni: If we want the old parser model, we still need to parse the document using one of the parser versions, for then to convert it afterward if necessary.

How can we clarify this better? 🤔

docs/generator-template.md
jonaslagoni: Ahh, no 😆 Reverting!

docs/configuration-file.md
jonaslagoni: Done 👍

lib/parser.js
derberg: oh, ok, during review I did not look into `@smoya/multi-parser` code.

> // Defaulting to v1 parser to convert it to the old parserAPI afterwards.

by `old parserAPI` you do not mean JS Parser v1 but JS Parser v2 (Parser-API v1) ? right?

so `@smoya/multi-parser` is just for stuff related to Parser-API

---
I got confused, as for me `old` is the old API before invention of Parser-API

later in code you write

> // Convert to old version

and then it is related to JS Parser v1

---
> How can we clarify this better?

so just be precise with naming and use phrases like `JS Parser v1`, `Parser-API v1`, `Parser-API v2`

in sentence `Defaulting to v1 parser to convert it to the old parserAPI afterwards.` - like I literally have no idea what `v1 parser` is for you here as `old parserAPI` for me is `JS Parser v2` not `v1`

derberg: I can suggests edits, but first you need to confirm if I understood stuff

jonaslagoni: Yea that makes sense 👍 

Please do make a suggestion 🙇

lib/parser.js
```diff
@@ -1,33 +1,47 @@
...
+  }
   const options = convertOldOptionsToNew(oldOptions, generator);
-  return defaultParser.parse(asyncapi, options);
+  const parser = NewParser(apiVersion, {parserOptions: options, includeSchemaParsers: true});
+  return parser.parse(asyncapi, options);
 };
 
+/**
+ * If the template expect one of the new parser API versions, it must be above 0
```
derberg, line 29:
```suggestion
 * If the template expect one of the Parser-API versions, it must be above 0
```

lib/parser.js
```diff
@@ -1,33 +1,47 @@
...
 
-parser.getProperApiDocument = (asyncapiDocument, templateConfig) => {
-  return parser.usesNewAPI(templateConfig) ? asyncapiDocument : convertToOldAPI(asyncapiDocument);
+/**
+ * Based on the current parsed AsyncAPI document, convert it to expected API version from the template.
+ */
+parser.getProperApiDocument = (asyncapiDocument, templateConfig = {}) => {
+  const apiVersion = this.sanitizeTemplateApiVersion(templateConfig.apiVersion);
+  if (apiVersion === undefined) {
+    // Convert to old version
```
derberg, line 41:
```suggestion
    // Convert to old API from JS Parser v1
```

lib/parser.js
derberg: done, wdyt?

jonaslagoni: Looks good to me 👍

derberg (approved):
LGTM 🚀

jonaslagoni:
Thanks for the review @derberg 🙇

PR #1031, opened 2023-08-24
chore: remove me as maintainer

fmvilas:
👋  My job here is done. I'm actually not investing any time in maintaining Generator and I won't be able to do it so I'm stepping down. I'd be happy to join some conversations though so ping me if you feel I'm needed.

Fun fact: I had to sync my fork to create this PR because the CODEOWNERS file didn't exist in my fork 😅


jonaslagoni (approved):
👍

Issue #1038, opened 2023-09-19
apiVersion 2 only works with AsyncAPI v2 documents


jonaslagoni:
#### Describe the bug
When I created the feature i did not notice that all CLIs used the generateFromString function as an entry point and not generate.

This means that the template configuration is not taken into account when parsing the document, i.e. it defaults to parsing with apiVersion 1 which does not support AsyncAPI v3.

jonaslagoni:
I am working on a fix

PR #1039, opened 2023-09-27
feat: enable string input for `generate`

jonaslagoni:
**Description**
This feature was driven by the following problems that needed to be fixed:

- In https://github.com/asyncapi/generator/pull/1017 I did not know that CLI used `generateFromString` instead of generate, and that was not changed at all.
- `generateFromString` parses the document with the old parser, and that means it cannot understand v3 documents. 

Changes:
- `generateFromString` is being deprecated and `generate` should be used instead.
- `generate` now handles string inputs and parses the document according to the template configuration `apiVersion`
- `validateTemplateConfig` now accesses the server object according to the parser API

**Related issue(s)**
Fixes https://github.com/asyncapi/generator/issues/1038


jonaslagoni:
I cant for the life of me disable or ignore the sonar cloud issues. If anyone knows how do let me know..

lib/generator.js
```diff
@@ -155,13 +155,19 @@ class Generator {
...
+   * @param  {Object} [parseOptions={}] AsyncAPI Parser parse options. Check out {@link https://www.github.com/asyncapi/parser-js|@asyncapi/parser} for more information. Remember to use the right options to the right parser depending on the template you are using.
    * @return {Promise}
    */
-  async generate(asyncapiDocument) {
-    if (!isAsyncAPIDocument(asyncapiDocument)) throw new Error('Parameter "asyncapiDocument" must be an AsyncAPIDocument object.');
-
+  // NOSONAR
+  async generate(asyncapiDocument, parseOptions = {}) {
+    const isAlreadyParsedDocument = isAsyncAPIDocument(asyncapiDocument);
+    const isParsableCompatible = asyncapiDocument || typeof asyncapiString === 'string';
```
derberg, line 165: where is `asyncapiString` coming from 🤔

derberg (changes requested):
in case of sonar, just remove these `NOSONAR` comments and modify https://github.com/asyncapi/generator/blob/master/.sonarcloud.properties

just exclude all tests like `test/**/*`

for cognitive complexity, put this before the function `/* eslint-disable sonarjs/cognitive-complexity */`

---

also please adjust title, it is a feat imho because of deprecation. So write what you change and not what you fix

lib/generator.js
jonaslagoni: Nice catch, was actually a bug 😆 

Added a test to catch it.

jonaslagoni:
> for cognitive complexity, put this before the function /* eslint-disable sonarjs/cognitive-complexity */

This is not respected by sonar.

> just exclude all tests like test/**/*

Added 👍 

> also please adjust title, it is a feat imho because of deprecation. So write what you change and not what you fix

Alright 👍

jonaslagoni:
@derberg yea but that will just increase global threshold, not really ignoring single function 😅

derberg:
@jonaslagoni which is ok right? better just increase threshold, and then we should worry when we go beyond, and see what we can do, rather than just silent forever?

jonaslagoni:
> @jonaslagoni which is ok right? better just increase threshold, and then we should worry when we go beyond, and see what we can do, rather than just silent forever?

I guess 😄 Updated it.

derberg:
looks like chatGPT lied 😅 yeah, remove the fake setting (including `// eslint-disable-next-line sonarjs/cognitive-complexity`) and open a followup good first issue to solve complexity issue - and let us merge

jonaslagoni:
> looks like chatGPT lied 😅 yeah, remove the fake setting (including `// eslint-disable-next-line sonarjs/cognitive-complexity`) and open a followup good first issue to solve complexity issue - and let us merge

Had to leave the `eslint-disable-next-line` as otherwise our linting tool would complain.

Did a quick test with the HTML template re-write just to make sure everything still worked as expected with v3 documents. And found a problem with tests and implementation. 

@derberg should be all ready for the final review now ✌️

derberg (changes requested):
Did you check with other templates, that are not yet migrated, like for example `nodejs-template`

I'm getting:
```bash
# running from local from fix_v3_support branch
$ ./cli.js https://bit.ly/asyncapi @asyncapi/nodejs-template -p server=production -o lolo --force-write
Something went wrong:
TypeError [ERR_INVALID_ARG_TYPE]: The "data" argument must be of type string or an instance of Buffer, TypedArray, or DataView. Received undefined
    at Object.writeFileSync (node:fs:2146:5)
    at createAsyncapiFile (/Users/xxx/generator/node_modules/@asyncapi/generator-hooks/src/index.js:28:6)
    at /Users/xxx/generator/lib/generator.js:827:20
    at Array.map (<anonymous>)
    at Generator.launchHook (/Users/xxx/generator/lib/generator.js:824:28)
    at Generator.generate (/Users/xxx/generator/lib/generator.js:208:18)
    at async /Users/xxx/generator/cli.js:159:9
```

derberg:
error comes from `at createAsyncapiFile (/Users/xxx/generator/node_modules/@asyncapi/generator-hooks/src/index.js:28:6)`

console log in the location
```
  console.log('dupa', asyncapi)
  fs.writeFileSync(asyncapiOutputLocation, asyncapi);
```

returns `undefined`

`asyncapi` variable comes from `const asyncapi = generator.originalAsyncAPI;`

and yeah, I checked, `generator` no longer provides `originalAsyncAPI`

lib/generator.js
```diff
@@ -250,22 +277,15 @@ class Generator {
...
    * @param  {String} asyncapiString AsyncAPI string to use as source.
    * @param  {Object} [parseOptions={}] AsyncAPI Parser parse options. Check out {@link https://www.github.com/asyncapi/parser-js|@asyncapi/parser} for more information.
+   * @deprecated Use the `generate` function instead. Just change the function name and it works out of the box.
    * @return {Promise}
    */
   async generateFromString(asyncapiString, parseOptions = {}) {
-    if (!asyncapiString || typeof asyncapiString !== 'string') throw new Error('Parameter "asyncapiString" must be a non-empty string.');
-
-    /** @type {String} AsyncAPI string to use as a source. */
-    this.originalAsyncAPI = asyncapiString;
```
derberg, line 259: 👀

derberg: related to my general comment

jonaslagoni:
@derberg not sure why I removed it 🤔

jonaslagoni:
Fixed it and added it to the tests

derberg (approved):
lgtm! 🚀 

will you create the followup for cognitive stuff or want me to?

jonaslagoni:
> will you create the followup for cognitive stuff or want me to?

If you got time please do 👍 Heading over to the templates right now to fix the remaining stuff so they are ready to be reviewed as well.

jonaslagoni:
Thanks @derberg 👍

derberg:
done https://github.com/asyncapi/generator/issues/1040

Issue #1002, opened 2023-06-30
Generator cannot handle referenced types (in separate files)


oskarcarlstedt:
#### Describe the bug
It looks like the generator cannot handle  types that are referenced in other files. Declaring the type locally in the schemas section works perfectly, but not when declaring the type in a separate file and referencing it.

I like to use the same model definitions in OpenAPI and AsyncAPI implementations. Therefore I want to share some types between these two types of API definitions.

#### How to Reproduce
Just run the generator  using the java-spring template and we gen an error saying this is not an AsynAPI file.

asyncapi: "2.6.0"
info:
  title: An asynchronous API
  version: 1.0.0
  description: |
    An API dealing with asynchronous actions on some objects

defaultContentType: application/json
channels:
  objects.new:
    subscribe:
      operationId: pushNewObjectToSubscribers
      message:
        $ref: "#/components/messages/ObjectMessage"

components:
  messages:
    ObjectMessage:
      payload:
        $ref: "types.yaml#/components/schemas/SomeType".  <<<--- This is the type that cannot be found

#### Expected behavior
Find and read the type in the reference file.

derberg:
@oskarcarlstedt please format your example asyncapi file so it is easy to copy and paste so we can try replicating the issue. And share `types.yaml` too 🙏🏼

michael-ball-ctct:
Hi @derberg 
Obviously, not quite the same example, but I'm fairly sure I'm running into the same issue with the following files in the same folder:
```yml
# asyncapi.yml
asyncapi: '2.6.0'
info:
  title: Account Service 2
  version: 1.0.0
  description: This service is in charge of processing user signups
channels:
  user/signedup:
    subscribe:
      message:
        $ref: '#/components/messages/UserSignedUp'
components:
  messages:
    UserSignedUp:
      payload:
        type: object
        properties:
          displayName:
            type: string
            description: Name of the user
          email:
            type: string
            format: email
            description: Email of the user
    SignUpFailed:
      payload:
        $ref: ./asyncapi-ref.yml#/components/messages/SignUpFailure
```

```yml
# asyncapi-ref.yml
components:
  messages:
    SignUpFailure:
      payload:
        type: object
        properties:
          displayName:
            type: string
            description: Name of the user
          email:
            type: string
            format: email
            description: Email of the user
          failure_reason:
            type: string
            description: The reason this signup was unsuccessful
```

Which results in:
[...]

PR #1058, opened 2023-10-30
feat: report parse failure reasons

### LINKED ISSUE #1051: Report parse failure reasons
#### Reason/Context

This is a general Quality of Life improvement for anyone using the generator. By implementing this, the speed at which someone trying to process AsyncAPI docs can diagnose and understand the issues with their document formatting will significantly increase. 

#### Description

The changes for this are fairly simple. In `lib/generator.js - Generator.parseInput()` block where the document is reported as "not correct" (currently at https://github.com/asyncapi/generator/blob/master/lib/generator.js#L353 ), before the error the `diagnostics` from the error response can be looped over to `console.error` out each of the issues found before the error is thrown. Something like:
``` js
      if (!document) {
        const err = new Error('Input is not a correct AsyncAPI document so it cannot be processed.');
        err.diagnostics = diagnostics;

        for (const diag of diagnostics) {
          console.error(`Diagnostic err: ${diag['message']} in path ${JSON.stringify(diag['path'])} starting at L${diag['range']['start']['line']} C${diag['range']['start']['character']}, and ending at L${diag['range']['end']['line']} C${diag['range']['end']['character']}`);
        }

        throw err;
      } else {
        this.asyncapi = document;
      }
```

This would result in an error log being emitted before the thrown error ("Error: Input is not a correct AsyncAPI document so it cannot be processed.") such as:
```
Diagnostic err: Property "subscribe" is not expected to be here in path ["channels","user/signedup","subscribe"] starting at  L0 L185, and ending at L0 L240
```

Obviously, the exact phrasing of the error is less important than the information. `console.error("Diagnostic err:", diag);` would still be a huge improvement here (and I guess also be less brittle, as it doesn't rely so heavily on the structure of  the diagnostic objects that get returns being consistent...)

Thanks in advance for any consideration of this. I'm happy to make a PR for this if you want - just let me know which logging strategy is preferred.

michael-ball-ctct:
**Description**

Adds simple logging for parsing failures, to improve debugging speed and experience of incorrectly formatted asyncapi.yml files for users.

I opted for the more basic logging style of the two I mentioned in the issue - I don't know enough about the diagnostics objects to be sure that they'll always contain the same fields, so erred on the side of caution rather than trying to access fields within fields that aren't guaranteed to exist or adding overcomplicated presence-checking logic (as well as including all available information, regardless of what is provided).

**Related issue(s)**

Resolves #1051


lib/generator.js
```diff
@@ -353,6 +353,9 @@ class Generator {
       if (!document) {
         const err = new Error('Input is not a correct AsyncAPI document so it cannot be processed.');
         err.diagnostics = diagnostics;
+        for (const diag of diagnostics) {
+          console.error('Diagnotic err:', diag);
```
derberg, line 357: here you have a pretty good overview of diagnostics structure -> https://github.com/asyncapi/cli/blob/master/src/parser.ts#L73

I was wondering, how do you use generator? over CLI?

michael-ball-ctct: I'm generally importing the generator and evoking it via a script, similar to the example I provide in https://github.com/asyncapi/generator/issues/1002#issuecomment-1724923172 (though I've updated to use `generate`, rather than `generateFromString`, since that was recently deprecated).

Well, if I'm tracing it properly, that Diagnostics structure is a fun wee jaunt from your link to AsyncAPI/cli, to [AsyncAPI Parser](https://github.com/asyncapi/parser-js/blob/master/src/types.ts), to a small extension in [Stoplight's Spectral/core](https://github.com/stoplightio/spectral/blob/1cae2245f0e89b980a8dcabb3eeeb9ae294508cd/packages/core/src/types/spectral.ts#L13), and finally ending up in [Stoplight's dianostic type](https://github.com/stoplightio/types/blob/master/src/diagnostics.ts) (with an interesting circular dependency to their [parser type](https://github.com/stoplightio/types/blob/master/src/parsers.ts) file for some of the Diagnostic Properties, as well). But I guess if we can rely on that being fairly stable, then we can use the more human-friendly summary.

I also tried tracing back exactly where the `diagnostics` object is created, to see if there was any variation in there I should be concerned about, but got a bit lost trying to figure out exactly what `NewParser` would be creating.

Anyway, I'll switch out just dumping the `diagnostics` object for the more human-friendly logging.

michael-ball-ctct: @derberg Apologies for the delay, stuff came up in paid work. Change made.

It's interesting testing it against a `v2.6.0` AsyncAPI.yml definition, vs. a `v3.0.0` definition with a simple indentation error. `v2.6.0` was noisy to the point of me questioning if this was even going to be an effective change, whereas `v3.0.0` was comparatively terse and useful. (Payloads and repsonses listed below.) Obviously, in generally I assume it's recommended to try and use as current a version as possible, but backwards compatibility is always nice.

Listing line number can be slightly misleading, such as in my use case I listed above, where the initial format is first converted (in my case for the purpose of resolving external `$ref`s) and so line numbers can be lost/changed before parsing - but I'm willing to chalk that up to user error (especially in this case).

Overall, I think it's still better than the current lack of reporting, but not necessarily by much. Unsure how/if to proceed with this. Any thoughts?

Thanks, as always.

-----

3.0.0 payload.
```
asyncapi: '3.0.0'
info:
  title: Account Service 2
  version: 1.0.0
  description: This service is in charge of processing user signups
channels:
  user/signedup:
    subscribe:
      message:
        $ref: '#/components/messages/UserSignedUp'
components:
  messages:
    UserSignedUp:
        payload: #incorrect indentation error
        type: object
        properties:
          displayName:
            type: string
            description: Name of the user
          email:
            type: string
            format: email
            description: Email of the user

```
3.0.0 error logs:
```
Diagnostic err: Property "subscribe" is not expected to be here in path ["channels","user/signedup","subscribe"] starting L10 C18, ending L12 C54
Diagnostic err: "payload" property type must be object,boolean in path ["components","messages","UserSignedUp","payload"] starting L20 C19, ending L20 C23
Diagnostic err: Property "type" is not expected to be here in path ["components","messages","UserSignedUp","type"] starting L21 C16, ending L21 C24
Diagnostic err: Property "properties" is not expected to be here in path ["components","messages","UserSignedUp","properties"] starting L22 C21, ending L30 C46

/home/mball/dev/api-portal-alpha/node_modules/@asyncapi/generator/lib/generator.js:225
        const err = new Error('Input is not a correct AsyncAPI document so it cannot be processed.');
                    ^
Error: Input is not a correct AsyncAPI document so it cannot be processed.
```

2.6.0 payload:
```
asyncapi: '2.6.0'
info:
  title: Account Service 2
  version: 1.0.0
  description: This service is in charge of processing user signups
channels:
  user/signedup:
    subscribe:
      message:
        $ref: '#/components/messages/UserSignedUp'
components:
  messages:
    UserSignedUp:
        payload: #incorrect indentation error
        type: object
        properties:
          displayName:
            type: string
            description: Name of the user
          email:
            type: string
            format: email
            description: Email of the user

```
2.6.0 error logs:
```
Diagnostic err: AsyncAPI document should have "defaultContentType" field. in path [] starting L1 C0, ending L30 C46
Diagnostic err: AsyncAPI document should have "id" field. in path [] starting L1 C0, ending L30 C46
Diagnostic err: AsyncAPI document should have non-empty "servers" object. in path [] starting L1 C0, ending L30 C46
Diagnostic err: AsyncAPI object should have non-empty "tags" array. in path [] starting L1 C0, ending L30 C46
Diagnostic err: The latest version of AsyncAPi is not used. It is recommended update to the "3.0.0" version. in path ["asyncapi"] starting L2 C14, ending L2 C21
Diagnostic err: Info object should have "contact" object. in path ["info"] starting L3 C9, ending L6 C73
Diagnostic err: Info object should have "license" object. in path ["info"] starting L3 C9, ending L6 C73
Diagnostic err: Operation should have an "operationId" field defined. in path ["channels","user/signedup","subscribe"] starting L10 C18, ending L12 C54
Diagnostic err: "message" property must have required property "oneOf" in path ["channels","user/signedup","subscribe","message"] starting L11 C18, ending L12 C54
Diagnostic err: "message" property must match "then" schema in path ["channels","user/signedup","subscribe","message"] starting L11 C18, ending L12 C54
Diagnostic err: Property "payload" is not expected to be here in path ["channels","user/signedup","subscribe","message","payload"] starting L11 C18, ending L12 C54
Diagnostic err: "payload" property type must be object,boolean in path ["channels","user/signedup","subscribe","message","payload"] starting L11 C18, ending L12 C54
Diagnostic err: Property "properties" is not expected to be here in path ["channels","user/signedup","subscribe","message","properties"] starting L11 C18, ending L12 C54
Diagnostic err: Property "type" is not expected to be here in path ["channels","user/signedup","subscribe","message","type"] starting L11 C18, ending L12 C54
Diagnostic err: Message should have a "messageId" field defined. in path ["channels","user/signedup","subscribe","message"] starting L11 C18, ending L12 C54
Diagnostic err: "UserSignedUp" property must have required property "oneOf" in path ["components","messages","UserSignedUp"] starting L19 C21, ending L30 C46
Diagnostic err: "UserSignedUp" property must match "then" schema in path ["components","messages","UserSignedUp"] starting L19 C21, ending L30 C46
Diagnostic err: Message should have a "messageId" field defined. in path ["components","messages","UserSignedUp"] starting L19 C21, ending L30 C46
Diagnostic err: Property "payload" is not expected to be here in path ["components","messages","UserSignedUp","payload"] starting L20 C19, ending L20 C23
Diagnostic err: "payload" property type must be object,boolean in path ["components","messages","UserSignedUp","payload"] starting L20 C19, ending L20 C23
Diagnostic err: Property "type" is not expected to be here in path ["components","messages","UserSignedUp","type"] starting L21 C16, ending L21 C24
Diagnostic err: Property "properties" is not expected to be here in path ["components","messages","UserSignedUp","properties"] starting L22 C21, ending L30 C46

/home/mball/dev/api-portal-alpha/node_modules/@asyncapi/generator/lib/generator.js:225
        const err = new Error('Input is not a correct AsyncAPI document so it cannot be processed.');
                    ^
Error: Input is not a correct AsyncAPI document so it cannot be processed.

```

derberg: lgtm

derberg (approved):
I think it is ok to merge it this way. I quickly checked on local what effect error logs will have on AsyncAPI CLI and looks good

Sorry it took so long but nov/dec was AsyncAPI v3 release and then usual holidays

derberg:
@allcontributors please add @michael-ball-ctct for code

PR #1061, opened 2023-11-07
feat: allow generator to pull from private npm repo

AayushSaini101:
Resolves #538
Resolves #1100

**Description:**
* Allow the Generator to pull the template from the private repo. [ Check docs/installing-privateTemplate.md]
* Adding new testing to verify the Generator pulls template from the private Repository.
* Integrate project tests now run in isolation, each separately
* Now tests do not make any changes to the file system of the host (entrypoint script) - they run longer but yeah, there is no other way to do it
* The code of html-template no longer needs to be stored inside the test project
* Some await/async bugs and removal of redundant promises
* update arborist to as latest as possible without breaking node 12 support

PR #1069, opened 2023-11-17
feat!: template pulling with npm config

pierrick-boule:
**Description**
Use npm config values in arborist in order to inherit form the global config when pulling a template.
This allow to pass config like private registry and proxy-auth.

**Related issue(s)**
https://github.com/asyncapi/generator/issues/538

# BREAKING CHANGE

This PR drops support for Node versions prior v18.12.0
We know this version of Generator will not work work with Node v12. We also discourage using Node v14 and v16 that are no longer under long time support from Node community. From now on, generator tests run under Node v18 and v20


pierrick-boule:
Hello,
The indent was not to make a dupplicate of https://github.com/asyncapi/generator/pull/1061 , I was scripting the generator at work this week and been trying to work around the limits of the generator (our ci has no direct acces to the web and to ways : private repository, http proxy with auth). Both are ok with npm because we can config it but not with the generator.
Maybe this PR can help to find a convenient way of using the generator as easily that npm by reusing the same source of configuration.

derberg (commented):
@pierrick-boule sorry for keeping you waiting with review, but we were all focused on v3 release

Thanks a lot for that PR. I also do not think PR somehow contradicts the other one about private registry. Tbh the more flexible we are with generator the better for people, especially in ci/cd setups.

My proposal is that first we go first with the other PR and then adjust this one to match it, cause we will have to somehow condition which auth should be used: the one passed as options directly to library, or by reusing the existing conf

If you have a private registry you could test against the other PR, that would help out a lot

aeworxet:
@asyncapi/bounty_team

derberg:
@pierrick-boule as I wrote before. This https://github.com/asyncapi/generator/pull/1061 is already merged, so please update your PR with latest private npm support

pierrick-boule:
Hello @derberg, i updated this PR, was able to test localy : 
I put "127.0.0.1 registry.npmjs.org" in /etc/hosts to make it unavailable (to remove later)
set a private registry : 
npm config set registry https://private.registry.org/ 
Run few test OK
used a docker proxy https://hub.docker.com/r/thelebster/docker-squid-simple-proxy/
npm config set proxy http://username:password@localhost:3128
Run few test OK

pierrick-boule:
Hello, i fixed the code style.
The backwards compatibility with nodeJS 12 is not that easy to get, i will need more time to check the old npm code :-( .

lib/generator.js
```diff
@@ -573,38 +577,47 @@ class Generator {
...
+      flatten,
+      shorthands,
+      npmPath
+    });
+
+    await config.load();
+
+    const arbOptions = {...{path: ROOT_DIR}, ...config.flat};
+
+    if (Object.keys(this.registry).length !== 0) {
```
derberg, line 599: why we need to change it like this?

also, am I correct assuming `registry` setting passes to a library will be ignored if config is found?

derberg (commented):
@pierrick-boule  there is no strong requirement to be compatible with node 12

we can definitely make this PR trigger major release because of stop of support for node12 - that is completely fine, this is a very old node version:
- just update docs
- `package.json` requirements
- workflows

I'm more concerned about your custom unix-based solution with replacing paths. I'm pretty sure it will not work with windows. How about using something like https://www.npmjs.com/package/npm-run-path ?

Also, we should find a way how to test it 🤔 should not be hard as tests are already in place https://github.com/asyncapi/generator/blob/master/test/test-project/test-registry.test.js where we simulate npm environment

lib/generator.js
pierrick-boule: During my local test the `if (this.registry)` was always true because of the default value of the constructor of Generator ( `registry = {}` ).
In this code the config from npm are loaded first in the arbOptions and if something is present in registry arbOptions will be loaded with it (like a manual overloading)

pierrick-boule:
Hello, 
Yes, I tried few implementation of npm config of npm 12 and it's very different of this PR, it will be verbose to implement a retrocompatible version and not very clear/maintainable, i was also thinking about depreciating it, and maybe introduce the new LTS of node (20) if possible.

I will now focus on this way (taking your proposals).

lib/generator.js
derberg: oh, totally makes sense, thanks

derberg:
@pierrick-boule we should support what Node community supports, so minimum is Node 18 that is in maintenance until April next year -> https://nodejs.org/en/about/previous-releases

since we agreed we will run a major release with this PR:
- please update title to `feat!`
- lets refresh node support so make sure package.json and docs talk about Node 18 as minimum
- make sure testing is also done Node 18 and 20
- make sure package-lock is using `lockfileVersion` set to `3`

derberg:
@pierrick-boule whenever you're ready for another review round please rerequest review as I do not get notifications that you pushed all required changes.

also please solve conflicts with `package-lock.json`

pierrick-boule:
Hello, @derberg it should be ok now.

derberg (commented):
looking awesome with this new additional integration test - and CI is passing.

you just need to update https://github.com/asyncapi/generator/blob/master/docs/using-private-template.md

there is also issue with unit tests: https://github.com/asyncapi/generator/actions/runs/8701813512/job/23883476282?pr=1069

I think it is because `@npm/config` is not mocked.

add below file, config.js
Screenshot 2024-04-16 at 17 28 21

with content like 

```js
const npmConfig = jest.genMockFromModule('@npmcli/config');

module.exports = npmConfig;
```

derberg (commented):
@Florence-Njeri please have a look at docs updates. With this PR we're releasing a new major as we will drop support for older node versions

docs/using-private-template.md
```diff
@@ -2,14 +2,26 @@
...
+* **npm config set _auth** : Provide the base64 encoded value that represents the username and password for basic auth.
+* **npm config set _authToken** : Provide the access token generated by the registry.
+
+## Private registry overriding arguments:
 
 * **registry.url**: The URL of the registry where the private template is located. Defaults to `registry.npmjs.org`.
 * **registry.auth**: An optional parameter to pass the npm registry username and password encoded with base64, formatted as `username:password`. For example, if the username and password are `admin` and `nimda`, you need to encode them with the base64 value like `admin:nimda` which results in `YWRtaW46bmltZGE=`.
-**registry.token**: An optional parameter to pass to the npm registry authentication token. To get the token, you can first authenticate with the registry using `npm login` and then grab the generated token from the `.npmrc` file.
+* **registry.token**: An optional parameter to pass to the npm registry authentication token. To get the token, you can first authenticate with the registry using `npm login` and then grab the generated token from the `.npmrc` file.
```
Florence-Njeri, line 24:
```suggestion
**registry.token**: An optional parameter to pass to the npm registry authentication token. To get the token, you can first authenticate with the registry using `npm login` and then grab the generated token from the `.npmrc` file.
```

Florence-Njeri (approved):

derberg:
@all-contributors please add @pierrick-boule for code,test,doc

Issue #1042, opened 2023-10-03
We need to explain in documentation the recommended way for types generation


derberg:
- in main overview we need clear info that if people are interested only in types/model generation, they should use modelina directly
- we also need a separate document with some example that explains that better to use modelina for models generation, if template requires it. We have examples like https://github.com/asyncapi/ts-nats-template/blob/master/template/src/models/models.js for example

@jonaslagoni wdyt?

black-snow:
Yes please. I just opened an issue for I want models only and I only got here by accident.

jonaslagoni:
Makes sense to me yea 👍

derberg:
you mean it could be done better?

jonaslagoni:
Always 😄 But yea, the integration feels a bit rough right now. Never the less we can of course document it as is and it can always be improved later.

derberg:
ok so the scope (writing it as I put it up for bounty) is to:
- clarify in main doc, probably intro that for only-model generation better people look int modelina and use cli to generate models
- have a dedicated doc that explains that when developing template, client or server code, where models generation is also important, the modelina library should be used - and how (some code snippets)

@jonaslagoni anything specific from your side? I think https://github.com/asyncapi/ts-nats-template/blob/master/template/src/models/models.js is the best example (with additional code comments), unless you think it makes sense to have another one with custom preset? I don't think so, better link directory to modelina examples right? as it is not much about generator-functionality. Also, you would probably like to provide 2.0 version of above mentioned example?

aeworxet:
Bounty Issue's End Of Life (EOL): [2024-02-29 23:59:59 UTC-12:00](https://time.lol/#2024-02-29T23:59:59-12:00)

mhmohona:
Would like to work on this.

Florence-Njeri:
Hey @derberg I'd love to work on this issue

derberg:
Since @Florence-Njeri is a maintainer in this repo, she has priority.
For transparency: I also contacted Florence before putting this issue under bounty to make sure she has time to work on the issue as she knows the generator docs the most

aeworxet:
### Bounty Issue's Timeline
|Complexity Level|Assignment date (by GitHub)|Start date (by BP rules)|End date (by BP rules)|Draft PR submission|Final PR submission|Final PR merge|
|----------------|---------------------------|------------------------|----------------------|-------------------|-------------------|--------------|
[...]

PR #1096, opened 2023-12-15
docs: create the model generation guide

Florence-Njeri:
**Description**
- New guide explaining how to go about generating types using Modelina
- In the introduction and the template development docs, I've added a note stating that if someone is only interested in types/model generation, they should use Modelina directly.
- Updated CLI naming

**Related issue(s)**
Resolves https://github.com/asyncapi/generator/issues/1042


docs/model-generation.md
```diff
@@ -0,0 +1,128 @@
+---
+title: "Generating models and classes using Modelina"
+weight: 200
+---
+
+Suppose you want to generate models for your application from the data types you have using the asyncapi generator. Then you should use Modelina instead of the asyncapi generator with a template.
```
derberg, line 6:
```suggestion
Use [AsyncAPI Modelina](/tools/modelina) to generate models for an application. Since models generation is something that you usually perform as an automated step through CI/CD, best way to use Modelina is through [AsyncAPI CLI](/tools/cli).

In case you work on a template for generating an application and part of template responsibilities is to provide models, you should integrate Modelina library instead of templating model structures using features from AsyncAPI Generator.
```

derberg (changes requested):
You will need to adjust narrative in the main document as the scope of https://github.com/asyncapi/generator/issues/1042 was not a tutorial, but more of a guide that tells you what is needed and shows an example. 

- we do not need to mention https://github.com/derberg/python-mqtt-client-template. You can just add a note that the example you provide can be integrated in a template that one can create by following https://www.asyncapi.com/docs/tools/generator/generator-template
- I think it is enough that first step of the guide just says that `@asyncapi/modelina` needs to be added to template dependencies as `npm install --save @asyncapi/modelina` because then it is automatically added to `package.json`
- `Model generation using the AsyncAPI CLI` section can be removed as it is not adding anything new, template with or without modelina - generation is always the same. You can just link to https://www.asyncapi.com/docs/tools/generator/usage#generator-cli (btw we have error there as section should be called AsyncAPI CLI, and we should also have a link to https://www.asyncapi.com/docs/tools/cli there)
- please remember about proper linking inside asyncapi.com with relative links and also consistent usage of upper/lower cases with names of tools




also, in your PR can you add between line 37 and 38 in https://github.com/asyncapi/generator/blob/master/.github/workflows/update-docs-in-website.yml#L37-L38 a line `rm -r ./pages/docs/tools/generator` to always have a clean and empty state of generator docs folder in website? Problem is that we now do not remove, and just add, so when we rename files like this https://github.com/asyncapi/generator/blob/master/docs/generator-template.md then we now get a duplicated tutorial in website
Screenshot 2024-01-04 at 14 24 49

Florence-Njeri:
> You will need to adjust narrative in the main document as the scope of #1042 was not a tutorial, but more of a guide that tells you what is needed and shows an example.
> 
> * we do not need to mention https://github.com/derberg/python-mqtt-client-template. You can just add a note that the example you provide can be integrated in a template that one can create by following https://www.asyncapi.com/docs/tools/generator/generator-template
> * I think it is enough that first step of the guide just says that `@asyncapi/modelina` needs to be added to template dependencies as `npm install --save @asyncapi/modelina` because then it is automatically added to `package.json`
> * `Model generation using the AsyncAPI CLI` section can be removed as it is not adding anything new, template with or without modelina - generation is always the same. You can just link to https://www.asyncapi.com/docs/tools/generator/usage#generator-cli (btw we have error there as section should be called AsyncAPI CLI, and we should also have a link to https://www.asyncapi.com/docs/tools/cli there)
> * please remember about proper linking inside asyncapi.com with relative links and also consistent usage of upper/lower cases with names of tools
> 
> also, in your PR can you add between line 37 and 38 in https://github.com/asyncapi/generator/blob/master/.github/workflows/update-docs-in-website.yml#L37-L38 a line `rm -r ./pages/docs/tools/generator` to always have a clean and empty state of generator docs folder in website? Problem is that we now do not remove, and just add, so when we rename files like this https://github.com/asyncapi/generator/blob/master/docs/generator-template.md then we now get a duplicated tutorial in website Screenshot 2024-01-04 at 14 24 49

Hey @derberg, I have addressed all the feedback you left on this PR

docs/usage.md
```diff
@@ -43,7 +43,7 @@ npm install <folder>
 
 ### Global templates installed with `yarn` or `npm`
 
-You can preinstall templates globally before installing the generator CLI. The generator first tries to locate the template in local dependencies; if absent it checks where the global generator packages are installed.
+You can preinstall templates globally before installing the [AsyncAPI CLI](https://www.asyncapi.com/docs/tools/cli). The generator first tries to locate the template in local dependencies; if absent it checks where the global generator packages are installed.
```
Florence-Njeri, line 46: @derberg I'm not sure if this is the right place to link to the CLI tool docs. Let me know, what you think

Florence-Njeri:
@derberg this is ready for a second review

docs/usage.md
derberg: lgtm

docs/model-generation.md
```diff
@@ -0,0 +1,108 @@
...
+>Note:
+>You can integrate this example into your own template by following the [generator template tutorial](https://www.asyncapi.com/docs/tools/generator/generator-template). Ensure that you add `@asyncapi/modelina` to template dependencies using `npm install --save @asyncapi/modelina`.
+
+## Generate models
+
+To learn how to generate models using the AsyncAPI CLI, refer to the section on Generator CLI usage. Execute the command provided there, and upon successful generation, the output will guide you to the location of your newly created model files in the specified directory **test/project/src/models**.
+
+## Conclusion
+
+Modelina provides a flexible and powerful way to generate data models from AsyncAPI, OpenAPI, or JSON Schema documents. With the AsyncAPI CLI, you can easily generate models and integrate them into your development workflow. With the ability to customize the generated models, Modelina proves to be a valuable tool in the development of event-driven architectures.
```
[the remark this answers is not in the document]
Florence-Njeri: Taking a look

Florence-Njeri (commented):
Review from Lukasz

derberg:
@Florence-Njeri please have a look at this comment with big modification - there was suppose to be much more content, not just finish with adding dependency 🤔 

also do not forget about 👇🏼 
> please also update the title and description of PR - we do not merge tutorial with this PR

derberg:
sorry @Florence-Njeri my comment was weirdly interpreted by GitHub commit suggestion

pasting the same comment below:




---
title: "Adding models generation in template"
weight: 200
---

This guide will walk you through the process of enabling models/types generation in a template by using [Modelina](https://www.asyncapi.com/tools/modelina).

Modelina is an AsyncAPI library designed for generating data models by utilizing inputs like [AsyncAPI](generator/asyncapi-document), OpenAPI, or JSON schema inputs. Its functionality revolves around creating data models from the provided AsyncAPI document and the model template, which defines message payloads. It is better to use Modelina in your template to handle model generation rather than providing custom templates.

You can integrate work shown in this guide in a template that you can create by following [tutorial about creating a template](https://www.asyncapi.com/docs/tools/generator/generator-template).

In this guide, you'll learn how to use Modelina in a template code to enable support for Python data models generation. 

## Add Modelina dependency

Install Modelina in your project using npm: `npm install --save @asyncapi/modelina`.

Your template's `package.json` file should now contain Modelina pointing to its latest version:

  ```json
  "dependencies": {
     ...
     "@asyncapi/modelina": "^2.0.5"
     ...
   }
  ```

## Create a models.js file

Create a new directory in the **template** directory named **src/models** and create a **models.js** file within it. In the **models.js** file, add the following code:

```js
// 1
import { File } from '@asyncapi/generator-react-sdk';
// 2
import { PythonGenerator, FormatHelpers } from '@asyncapi/modelina';

/**
 * @typedef RenderArgument
 * @type {object}
 * @property {AsyncAPIDocument} asyncapi document object received from the generator.
 */

/**
 * Render all schema models
 * @param {RenderArgument} param0 
 * @returns 
 */
// 3
export default async function schemaRender({ asyncapi }) {
  // 4 
  const pythonGenerator = new PythonGenerator();
  // 5
  const models = await pythonGenerator.generate(asyncapi);
  // 6
  const files = [];
  // 7
  for (const model of models) {
    // 8
    const modelFileName = `${FormatHelpers.toPascalCase(model.modelName)}.py`;
    // 9
    files.push(<File name={modelFileName}>{model.result}</File>);
  }
  return files;
}
```

Let's break it down. The code snippet above does the following:

1. The **File** component from the [generator react SDK](https://github.com/asyncapi/generator-react-sdk) is needed to handle further rendering of generated models into files.
2. The **PythonGenerator** generator is a core that you need for models generation. Additionally you can import [**FormatHelpers**](https://github.com/asyncapi/modelina/blob/master/src/helpers/FormatHelpers.ts) that give you a set of helpers that make it easier to modify model names to match your required case.
3. You can change the name **schemaRender** to anything else like **modelRenderer**. More important is that this must be an `async` function and be a default export. This function is invoked during generation process and it should contain the logic behind models generation.
4. First you create an instance of the **PythonGenerator** model generator. If you decide to use present functionality from Modelina, you need to pass your presets here during instance creation.
5. Actual models generation is one line of code, and as a result you get an array of models that later you need to turn into files.
6. You need to define an array that must be returned from **schemaRender** function. The array must contain React components, and in this case `<File>` component.
7. Iterate over generated models and use their content to create proper definitions of `<File>` component.
8. Notice how using Modelina helpers, in this case **toPascalCase** function, you can make sure that the filename of your model follows specific case pattern.
9. Each component must be added to **files** array that you later return from the default function. Notice the definition of `<File>` component that enables you to provide the name of resulting file and the content of the model. Notice also `model.result` that shows that initially generated array with models did not contain raw models content but a set of output objects that contain not only `result` but also other info, like for example `modelName`.

With such model template that uses Modelina, as a result of generation process you would receive a set of model files in `$OUTPUT_DIR/src/models` directory.

## Conclusion

Modelina provides a flexible and powerful way to generate data models from AsyncAPI, OpenAPI, or JSON Schema documents. By integrating Modelina you can much faster enable models generation in your template. 
```

suggested restructure and rewrite:
- give more details
- follow guide not tutorial
- remove CLI as in this topic CLI is not relevant, especially that Generator can also be used as library and not only as tool integrated in AsyncAPI CLI

@Florence-Njeri I suck at grammar, so please after accepting the change, do a review and probably improvement
```

derberg (approved):
@jonaslagoni @kennethaasan @magicmatatjahu since this is Modelina related, anything from your side? is it a good way to promote Modelina as component that enables models generation in app template?

maybe long term would be better to have a dedicated React component that makes it even easier to embed model generation - but for now I think it should be enough, especially that there is nothing else at the moment

jonaslagoni (approved):
LGTM 👍

Issue #1116, opened 2024-02-08
Allow $$operation$$ file template


lukas-mertens:
For the v3 spec operations are much more important for templates. Please also support `$$operation$$`. 

https://github.com/asyncapi/generator/blob/7629c168532dbc1c6504d7417242084fa7c2d2ef/docs/file-templates.md?plain=1#L6C1-L16C49

derberg:
closing this issue

we should not add more features to Nunjucks.

and for React - functionality is already in place - we just need to properly document it -> https://github.com/asyncapi/generator/issues/1199

PR #1117, opened 2024-02-08
feat: add `$$operation$$` file template

lukas-mertens:
**Description**

I added `$$operation$$` as a file template.

**Related issue(s)**

#1116


lukas-mertens:
@jonaslagoni I see that you introduced the v3 support. Does it make any sense what I am suggesting or am I missing something?

jonaslagoni (changes requested):
Seems like the tests are failing :v:

lib/generator.js
```diff
@@ -743,6 +744,7 @@ class Generator {
         objectSchema,
         parameter: convertMapToObject(this.getAllParameters(asyncapiDocument)),
         everySchema: convertMapToObject(asyncapiDocument.allSchemas()),
+        operation: convertCollectionToObject(asyncapiDocument.allOperations()),
```
derberg, line 747: shouldn't you use `convertMapToObject` like other parts?

derberg:
@jonaslagoni yeah, to be honest I did not recall `file templates` are actually useful for nunjucks only 😄 

we have at least comparizon from the perspective of "template for templates". So we see file templates in use in https://github.com/asyncapi/template-for-generator-templates/tree/nunjucks/template/schemas in times of nunjucks and then after conversion to react https://github.com/asyncapi/template-for-generator-templates/blob/master/template/schemas/schema.js

just to be sure, @lukas-mertens what templating engine are you using?

lukas-mertens:
@derberg Yeah, I am using the old one. I didn't see it in any documentation that a new one exists and I was building my template based on https://github.com/asyncapi/dotnet-nats-template/tree/master

I didn't know it is that outdated back then. It would be really nice to have a few reference-implementations based on the new api version and new renderer

derberg:
actually https://github.com/asyncapi/dotnet-nats-template/tree/master runs on new engine, the `react` one

also in docs that talk about template development we reference this template example implementation -> https://github.com/asyncapi/template-for-generator-templates

is the source code of your template somewhere available?

my assumption is you use `react` so you do not need file templates. Have a look at examples I shared in previous comment and lemme know

derberg:
@lukas-mertens hey, did you have a chance to have a look at it?

derberg:
closing this PR

we should not add more features to Nunjucks.

and for React - functionality is already in place - we just need to properly document it -> https://github.com/asyncapi/generator/issues/1199

Issue #1044, opened 2023-10-04
How about a monorepo 😄 


derberg:
## Current state

- https://github.com/asyncapi/generator/ that is a core
- https://github.com/asyncapi/generator-react-sdk for react render engine
- https://github.com/asyncapi/generator-filters/ with shared filters
- https://github.com/asyncapi/generator-hooks/ with shared hooks
- Multiple repositories with templates

This granularity is technically possible so templated development is flexible. 
**But the fact that we make it possible for others, to easily create private, independent and use case specific templates doesn't mean we should have this level of granularity in asyncapi GH org**

## Proposal - discussion

How about we turn `generator` into a monorepo?

- `generator-react-sdk` - not a separate npm package, just natively part of generator
- `generator-filters` and `generator-hooks` also natively a part of generator (we still want to enable people to provide theirs independently)
- yes, templates as part of generator repo, but in case of templates, I guess we would need to publish separate packages 🤔 although I'm not 100% sure we need to

## Why

- having to use `generator-react-sdk` as separate package is kinda confusing and in general my impression is that motivation for having it as a separate package was that then it can be written in TS
- over last 2 year we had zero improvements in hooks and filters, they are hidden small repos and template developers basically work on their own filters and hooks in their template repos. 
- it would not be obligatory to move `@asyncapi/*-template` to the monorepo. But I think many would, as having it in our repo, with multiple maintainers/champions (yeah, we would have champion like setup like in modelina) would make it easier and quicker to reuse solutions from other templates, filters, components and other stuff. Under one repo it would be easier to focus or reusability and good standards (like always using modelina for models generation)

We would still be flexible, independent templates would be possible, hooks and filters too.

Having all under one umbrella would make generator a more vibrant project, more actively developed and visible.

wdyt?

any disadvantages that I do not see?

AyushNautiyalDeveloper:
@derberg I saw we need to implement monorepo for this. I would like to work on this. please let me know how should i go about this.

ayushnau:
@derberg i saw we need to implement the monorepo for this i would like to work on this. Please let know how should i go about this.

derberg:
@ayushnau for now it is proposal and a discussion

ayushnau:
ok @derberg

derberg:
@jonaslagoni @magicmatatjahu wdyt

Aryann15:
Hey! would like to work on this issue, can i get it assigned? @derberg 
Thanks!

[...]

PR #1155, opened 2024-03-28
feat: introduce monorepo with filters package included

ayushnau:
Changes Added. 
1. Turborepo Added. 
2. Converted the generator repo to app,  inside the turborepo. 
3. Added the required scripts of the turborepo to the root package.json file. 

**Related issue(s)**

Resolves https://github.com/asyncapi/generator/issues/1044


derberg (commented):
this PR is huge

can we introduce the change at least in 2 PRs? like first add `generator-filters` and then move `generator` to apps? cause now you just did moved all other files inside `apps/generator` but without checking if it makes sense, like for example all GitHub workflows

ayushnau:
> this PR is huge
> 
> can we introduce the change at least in 2 PRs? like first add `generator-filters` and then move `generator` to apps? cause now you just did moved all other files inside `apps/generator` but without checking if it makes sense, like for example all GitHub workflows

@derberg should i convert the generator into turborepo. on its own. first . which means adding generator to the apps directory before as that would make sense.

derberg:
@ayushnau yeah, I know they will be big, we just need to figure some approach maybe that can show that new files that are added are not modified or something like that, so review is smooth

but yeah, in this particular PR I know that if you split in 2 it will be big anyway, but seeing you moving GH actions workflows inside apps made me think we need to do one app at a time

ayushnau:
> @ayushnau yeah, I know they will be big, we just need to figure some approach maybe that can show that new files that are added are not modified or something like that, so review is smooth
> 
> but yeah, in this particular PR I know that if you split in 2 it will be big anyway, but seeing you moving GH actions workflows inside apps made me think we need to do one app at a time

1. I think we can check by default in git which are only moved files
image

2. Actually I did removed the .github file but somehow the .github file changes didnt get added in the git changes.

derberg (commented):
- `apps/generator/.all-contributorsrc` must stay in the root
- same sonar clound properties config - should be one for all apps
- same eslint and editor config files right?
- same files like codeowners or code of conduct, license, contri guide
- same readme banner
- also readme seems to be duplicated

ayushnau:
> * `apps/generator/.all-contributorsrc` must stay in the root
> 
>     * same sonar clound properties config - should be one for all apps
> 
>     * same eslint and editor config files right?
> 
>     * same files like codeowners or code of conduct, license, contri guide
> 
>     * same readme banner
> 
>     * also readme seems to be duplicated

1. Ok and what about .editorConfig file. that also needs to be in root.

AyushNautiyalDeveloper:
@derberg for running docker command will i need any credentials. as I am unable to run the command.

ayushnau:
Acutually i was not able to get the changes in my branch. so need to refork it.

PR #1165, opened 2024-04-01
fix: `ts-node` is registered only when it's actually needed

### LINKED ISSUE #1030: `ts-node` is registered if `generator` is imported in a .js file
#### Describe the bug
`ts-node` is registered if `@asyncapi/generator` is simply imported in a file, even though the constructor or any other of `generator`'s methods is not called at all.
This means that after building a TypeScript project, when it is transpiled to JavaScript, `ts-node` is still registered, even though TypeScript is not needed anymore.
This also means that `ts-node` is registered in a JavaScript project that does not use TypeScript at all.
We noticed it because after adding `nestjs-asyncapi` (which uses `@asyncapi/generator`) to one of our bigger projects, the startup time of `npm start` increased by ~30 seconds.

#### How to Reproduce
Create a new file `test.js` which only imports `@asyncapi/generator` and does nothing else with it:
```
const { REGISTER_INSTANCE } = require('ts-node');
console.log(process[REGISTER_INSTANCE] == null); // prints "true"
const generator = require('@asyncapi/generator');
console.log(process[REGISTER_INSTANCE] == null); // prints "false"
```
run `node test.js` and look at the output

#### Expected behavior
`ts-node` is registered only when it's actually needed.

Gmin2:
**Description**

- Remove the unconditional registration 
- Created `checkForTypescriptFiles` and `hasTypescriptFilesDir` functions to check for the presence of  `ts`  files in the `filters` and `hooks` directories
- Conditionally register the `ts` transpiles in the `configureTemplateWorkflow` functions using `checkForTypescriptFiles`.

**Related issue(s)**
Fixes #1030


utnim2:
can you take a look @derberg

lib/filtersRegistry.js
```diff
@@ -11,10 +11,52 @@ const { isAsyncFunction } = require('./utils');
  * @param {String} filtersDir Directory where local filters are located.
  */
 module.exports.registerFilters = async (nunjucks, templateConfig, templateDir, filtersDir) => {
+  const hasTypescriptFiles = await checkForTypescriptFiles(templateDir, filtersDir);
+  if(hasTypescriptFiles) {
+    utils.registerTypescript(true)
```
lmgyuan, line 16: In my IDE, the utils seems to be undefined. You probably want to add `const { utils } = require('./utils');`

derberg: yup, @utnim2  please fix line 4 that for now requires only `isAsyncFunction`

derberg (commented):
can you explain why you think your solution must be implemented instead of what was suggested in https://github.com/asyncapi/generator/issues/1030#issuecomment-1705506844

wouldn't it just be easier to plug into the existing code that anyway "walks" over filters and hooks files and you can use it to figure if ts file is there or not?

utnim2:
> can you explain why you think your solution must be implemented instead of what was suggested in [#1030 (comment)](https://github.com/asyncapi/generator/issues/1030#issuecomment-1705506844)
> 
> wouldn't it just be easier to plug into the existing code that anyway "walks" over filters and hooks files and you can use it to figure if ts file is there or not?

The suggested solution iterates through all files even there is even there are no Typescript files present,my solution skips the registration process altogether if no Typescript file is present
I might be wrong here 🤔

lmgyuan:
> > can you explain why you think your solution must be implemented instead of what was suggested in [#1030 (comment)](https://github.com/asyncapi/generator/issues/1030#issuecomment-1705506844)
> > wouldn't it just be easier to plug into the existing code that anyway "walks" over filters and hooks files and you can use it to figure if ts file is there or not?
> 
> The suggested solution iterates through all files even there are no Typescript files present,my solution skips the registration process altogether if no Typescript file is present I might be wrong here 🤔

If I understand what @derberg said correctly, I think regardless of whether `hasTypeScript` is true or not in:

`
    const hasTypescriptFiles = await this.checkForTypescriptFiles();
    if(hasTypescriptFiles) {
      utils.registerTypeScript(true);
    }
`

This part of the code in `registerLocalFilters` will also be executed:
```
walker.on('file', async (root, stats, next) => {
      try {
        const filePath = path.resolve(templateDir, path.resolve(root, stats.name));
        // If it's a module constructor, inject dependencies to ensure consistent usage in remote templates in other projects or plain directories.
        delete require.cache[require.resolve(filePath)];
        const mod = require(filePath);

        addFilters(nunjucks, mod);
        
        next();
      } catch (e) {
        reject(e);
      }
    });
```
which goes through the directories where filters are located. It is the same case for hooks. Therefore, instead of checking it in a different place, it might be preferable to just look for TS files when looking for filters and hooks, to make the codebase clearer and more concise.

They are my understandings. Please correct me if any of it is incorrect : ) @derberg

derberg:
@lmgyuan yes, this is exactly it, thanks. We anyway iterate there, thus better to inject TS validation there rather than have another iteration only for TS. Better performance

utnim2:
> @lmgyuan yes, this is exactly it, thanks. We anyway iterate there, thus better to inject TS validation there rather than have another iteration only for TS. Better performance

Ok making the changes

derberg:
do you need some help @utnim2 ?

Gmin2:
This is ready for review and merge

cc @derberg

derberg (commented):
there is still a linter issue we discussed during the call, and as you can see GitHub action fails because of it

derberg:
@Gmin2 you need to solve conflicts

derberg (approved):

PR #1169, opened 2024-04-03
ci: docker-compose command in PR testing with test project workflow

JeelRajodiya:
**Description**

- Changed command `docker-compose up` to `docker compose build`

**Related**

[Discussion #116610](https://github.com/orgs/community/discussions/116610)

### DIFF
--- a/.github/workflows/pr-testing-with-test-project.yml
+++ b/.github/workflows/pr-testing-with-test-project.yml
@@ -17,5 +17,5 @@ jobs:
       - name: Checkout repository
         uses: actions/checkout@v3
       - name: Run test
-        run: NODE_IMAGE_TAG=${{ matrix.node }} docker compose build --abort-on-container-exit --remove-orphans --force-recreate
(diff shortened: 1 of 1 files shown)


derberg (approved):
Thanks, just changed prefix, as `fix` would trigger release and we do not need to release anything

.github/workflows/pr-testing-with-test-project.yml
```diff
@@ -17,5 +17,5 @@ jobs:
       - name: Checkout repository
         uses: actions/checkout@v3
       - name: Run test
-        run: NODE_IMAGE_TAG=${{ matrix.node }} docker-compose up --abort-on-container-exit --remove-orphans --force-recreate
+        run: NODE_IMAGE_TAG=${{ matrix.node }} docker compose build --abort-on-container-exit --remove-orphans --force-recreate
```
derberg, line 20:
```suggestion
        run: NODE_IMAGE_TAG=${{ matrix.node }} docker compose up --abort-on-container-exit --remove-orphans --force-recreate
```

let's try this according to https://docs.docker.com/compose/migrate/#unsupported-in-v2

Issue #521, opened 2021-03-01
React templates make generation process longer


derberg:
#### Describe the bug

```
    ✓ generated using Nunjucks template (492ms) #html-template
    ✓ generate using React template (4875ms) #markdown-template
```

Markdown template is the simples one we have, yet it takes 5s to generate `asyncapi.md` with it. Which is really a lot and I bet it will be worse with more complicated templates.

#### How to Reproduce

Just try to generate with HTML template and then markdown one and you'll see that there is a time difference each time.

The problem is in transpilation process with babel. @magicmatatjahu checked and:

```
5 times in a row with transpilation in each render:
3.768s
2.391s
2.236s
2.260s
2.414s
Average: ~2.6s
And here is a time without transpilation in each render:
191.515ms
194.49ms
185.167ms
192.192ms
199.48ms
Average: ~0.193s
```

#### Expected behaviour

IMHO by default, when you run generation, an already transpiled template is used to speed up (yes, it means we publish transpiled files to npm too). The generator is smart and spots transpiled files are missing and it needs to run transpilation. Users (template developers) also have a flag that allows them to explicitly run transpilation during development

derberg:
@magicmatatjahu good point, done

magicmatatjahu:
@derberg To compile the React template and use that "compiled" templated from each run. At the moment we transpile/compile React component by babel on each generation and it is a waste of time.

derberg:
@magicmatatjahu thanks, now I remember. I just think now that it should work like this by default. When you run generation, generator first look for sources of compiled template, if provided, and if not, performs old way. And yeah still in template for templates we need to show how to get it into npm

magicmatatjahu:
> I just think now that it should work like this by default. When you run generation, generator first look for sources of compiled template, if provided, and if not, performs old way

Exactly in this way, look for transpiled files, if don't exist then run transpilation/compilation and generate template.

jonaslagoni:
@magicmatatjahu would you be able to formulate this into a good first issue? 🙂

derberg:
@magicmatatjahu I second what @jonaslagoni wrote because of this thread https://asyncapi.slack.com/archives/CQVJXFNQL/p1648304288930959

we are getting more and more React-based templates, means more unhappy users. 


btw, I was wondering if we could not make it

derberg:
Reopening. Now with budget in bounty program maybe we can find someone who could work on improving things

derberg:
What needs to be done:

[...]

PR #1177, opened 2024-04-04
feat: added a compile option to enable skipping default transpilation in react

utnim2:
**Description**

The PR introduces a new option `compile` in the  `Generator` class contructor which is by default `false` ,by default the `React` template files are not transpiled. When setting the `compile` options to `true` the transpilation process takes place.

I have also added/updated test for it. Below are the result

image

**Related issue(s)**
Related to #521 
image
which added this functionality

### DIFF
--- a/lib/generator.js
+++ b/lib/generator.js
@@ -43,7 +43,7 @@ const DEFAULT_TEMPLATES_DIR = path.resolve(ROOT_DIR, 'node_modules');
 
 const TRANSPILED_TEMPLATE_LOCATION = '__transpiled';
 const TEMPLATE_CONTENT_DIRNAME = 'template';
-const GENERATOR_OPTIONS = ['debug', 'disabledHooks', 'entrypoint', 'forceWrite', 'install', 'noOverwriteGlobs', 'output', 'templateParams', 'mapBaseUrlToFolder', 'url', 'auth', 'token', 'registry'];
+const GENERATOR_OPTIONS = ['debug', 'disabledHooks', 'entrypoint', 'forceWrite', 'install', 'noOverwriteGlobs', 'output', 'templateParams', 'mapBaseUrlToFolder', 'url', 'auth', 'token', 'registry', 'compile'];
 const logMessage = require('./logMessages');
 
 const shouldIgnoreFile = filePath =>
@@ -90,13 +90,15 @@ class Generator {
    * @param {String} [options.registry.token]     Optional parameter to pass npm registry auth token that you can grab from .npmrc file
    */
 
-  constructor(templateName, targetDir, { templateParams = {}, entrypoint, noOverwriteGlobs, disabledHooks, output = 'fs', forceWrite = false, install = false, debug = false, mapBaseUrlToFolder = {}, registry = {}} = {}) {
+  constructor(templateName, targetDir, { templateParams = {}, entrypoint, noOverwriteGlobs, disabledHooks, output = 'fs', forceWrite = false, install = false, debug = false, mapBaseUrlToFolder = {}, registry = {}, compile = false } = {}) {
     const options = arguments[arguments.length - 1];
     this.verifyoptions(options);
     if (!templateName) throw new Error('No template name has been specified.');
     if (!entrypoint && !targetDir) throw new Error('No target directory has been specified.');
     if (!['fs', 'string'].includes(output)) throw new Error(`Invalid output type ${output}. Valid values are 'fs' and 'string'.`);
 
+    /** @type {Boolean} Whether to compile the template or use the cached transpiled version. */
+    this.compile = compile;
     /** @type {Object} Npm registry information. */
     this.registry = registry;
     /** @type {String} Name of the template to generate. */
@@ -391,7 +393,7 @@ class Generator {
    */
   async configureTemplate() {
     if (isReactTemplate(this.templateConfig)) {
-      await configureReact(this.templateDir, this.templateContentDir, TRANSPILED_TEMPLATE_LOCATION);
+      await configureReact(this.templateDir, this.templateContentDir, TRANSPILED_TEMPLATE_LOCATION, this.compile);
     } else {
       this.nunjucks = configureNunjucks(this.debug, this.templateDir);
     }
--- a/lib/renderer/react.js
+++ b/lib/renderer/react.js
@@ -14,11 +14,13 @@ const reactExport = module.exports;
  * @param {string} templateContentDir where the template content are located
  * @param {string} transpiledTemplateLocation folder for the transpiled code
  */
-reactExport.configureReact = async (templateLocation, templateContentDir, transpiledTemplateLocation) => {
+reactExport.configureReact = async (templateLocation, templateContentDir, transpiledTemplateLocation, compile) => {
   const outputDir = path.resolve(templateLocation, `./${transpiledTemplateLocation}`);
-  await AsyncReactSDK.transpileFiles(templateContentDir, outputDir, {
-    recursive: true
-  });
+  if (compile) {
+    await AsyncReactSDK.transpileFiles(templateContentDir, outputDir, {
+      recursive: true
+    });
+  }
 };
 
 /**
--- a/test/generator.test.js
+++ b/test/generator.test.js
@@ -26,6 +26,7 @@ describe('Generator', () => {
       expect(gen.forceWrite).toStrictEqual(false);
       expect(gen.install).toStrictEqual(false);
       expect(gen.templateParams).toStrictEqual({});
+      expect(gen.compile).toStrictEqual(false);
     });
 
     it('works with all the params', () => {
@@ -39,6 +40,7 @@ describe('Generator', () => {
         templateParams: {
           test: true,
         },
+        compile: true,
       });
       expect(gen.templateName).toStrictEqual('testTemplate');
       expect(gen.targetDir).toStrictEqual(__dirname);
@@ -48,6 +50,7 @@ describe('Generator', () => {
       expect(gen.output).toStrictEqual('string');
       expect(gen.forceWrite).toStrictEqual(true);
       expect(gen.install).toStrictEqual(true);
+      expect(gen.compile).toStrictEqual(true);
       expect(() => gen.templateParams.test).toThrow('Template parameter "test" has not been defined in the package.json file under generator property. Please make sure it\'s listed there before you use it in your template.');
 
       // Mock params on templateConfig so it doesn't fail.


lib/generator.js
```diff
@@ -90,13 +90,15 @@ class Generator {
...
 
-  constructor(templateName, targetDir, { templateParams = {}, entrypoint, noOverwriteGlobs, disabledHooks, output = 'fs', forceWrite = false, install = false, debug = false, mapBaseUrlToFolder = {}, registry = {}} = {}) {
+  constructor(templateName, targetDir, { templateParams = {}, entrypoint, noOverwriteGlobs, disabledHooks, output = 'fs', forceWrite = false, install = false, debug = false, mapBaseUrlToFolder = {}, registry = {}, compile = false } = {}) {
     const options = arguments[arguments.length - 1];
     this.verifyoptions(options);
     if (!templateName) throw new Error('No template name has been specified.');
     if (!entrypoint && !targetDir) throw new Error('No target directory has been specified.');
     if (!['fs', 'string'].includes(output)) throw new Error(`Invalid output type ${output}. Valid values are 'fs' and 'string'.`);
 
+    /** @type {Boolean} Whether to compile the template or use the cached transpiled version. */
```
derberg, line 100:
```suggestion
    /** @type {Boolean} Whether to compile the template or use the cached transpiled version provided by template in `__transpiled` folder. */
```

lib/renderer/react.js
```diff
@@ -14,11 +14,13 @@ const reactExport = module.exports;
  * @param {string} templateContentDir where the template content are located
  * @param {string} transpiledTemplateLocation folder for the transpiled code
```
derberg, line 15: update jsdoc please

lib/generator.js
```diff
@@ -391,7 +393,7 @@ class Generator {
    */
   async configureTemplate() {
     if (isReactTemplate(this.templateConfig)) {
-      await configureReact(this.templateDir, this.templateContentDir, TRANSPILED_TEMPLATE_LOCATION);
+      await configureReact(this.templateDir, this.templateContentDir, TRANSPILED_TEMPLATE_LOCATION, this.compile);
```
derberg, line 396: I'm wondering if it makes sense to modify `configureReact` or maybe better just do conditional here, like

```
if (isReactTemplate(this.templateConfig) && this.compile) {
      await configureReact(this.templateDir, this.templateContentDir, TRANSPILED_TEMPLATE_LOCATION);
```

utnim2: I have added conditional transpilation

lib/renderer/react.js
```diff
@@ -13,12 +13,15 @@ const reactExport = module.exports;
  * @param {string} templateLocation located for thetemplate
  * @param {string} templateContentDir where the template content are located
  * @param {string} transpiledTemplateLocation folder for the transpiled code
+ * @param {Boolean} compile Whether to compile the template files or used the cached transpiled version provided by the template in the '__transpiled' folder
  */
-reactExport.configureReact = async (templateLocation, templateContentDir, transpiledTemplateLocation) => {
+reactExport.configureReact = async (templateLocation, templateContentDir, transpiledTemplateLocation, compile) => {
```
lmgyuan, line 18: You probably want to revoke the extra argument `compile` added here. Currently the 

`        await configureReact(this.templateDir, this.templateContentDir, TRANSPILED_TEMPLATE_LOCATION);
`

in generator.js has syntax error, which expects 4 arguments but your updated codes with a conditional statement only give it three arguments. 

```suggestion
reactExport.configureReact = async (templateLocation, templateContentDir, transpiledTemplateLocation) => {
```

lib/renderer/react.js
```diff
@@ -13,12 +13,15 @@ const reactExport = module.exports;
...
   const outputDir = path.resolve(templateLocation, `./${transpiledTemplateLocation}`);
-  await AsyncReactSDK.transpileFiles(templateContentDir, outputDir, {
-    recursive: true
-  });
+  if (compile) {
+    await AsyncReactSDK.transpileFiles(templateContentDir, outputDir, {
+      recursive: true
+    });
+  }
```
lmgyuan, line 24: Similar to my previous suggested change, since @derberg suggested to not modify the `configureReact` method, you should revoke your changes here as well. 

```suggestion
 
```

lib/generator.js
```diff
@@ -390,8 +392,13 @@ class Generator {
...
   async configureTemplate() {
-    if (isReactTemplate(this.templateConfig)) {
-      await configureReact(this.templateDir, this.templateContentDir, TRANSPILED_TEMPLATE_LOCATION);
+    if(isReactTemplate(this.templateConfig)) {
+      if(this.compile) {
+        await configureReact(this.templateDir, this.templateContentDir, TRANSPILED_TEMPLATE_LOCATION);
+      } else {
+        // use the cached transpiled version provided by the template
+        this.templateContentDir = path.resolve(this.templateDir, TRANSPILED_TEMPLATE_LOCATION);
+      }
```
lmgyuan, line 401: @derberg is this part correct? I thought when the `this.compile` is set to false, we are supposed to not do the transpilation. But I also don't know exactly what is supposed to be done when `this.compile` is false so I will let you decide : )

lmgyuan, line 401: Is the `else` section correct? I thought we are only going to do the transpilation when `this.compile` is set to be True. If I understand it correctly, the suggested change should be:

```suggestion
if (isReactTemplate(this.templateConfig) && this.compile) {
      await configureReact(this.templateDir, this.templateContentDir, TRANSPILED_TEMPLATE_LOCATION);
```

without a `else` clause, because we are not configuring it again if `this.compile` is set to be false.

utnim2: by default `compile` is set to false

lib/renderer/react.js
utnim2: will do it once i get back home

utnim2:
> ok, we need to add a proper new integration test here, without proper test, feature development will be based on just our intuition
> 
> there are some integration tests already, that use https://github.com/asyncapi/html-template template
> 
> we just need another test, also using https://github.com/asyncapi/html-template but fixed to use version `2.3.0` where `__transpiled` folder was added to package.
> 
> there are 2 test scenarios:
> 
> * with `compile` flag `false`
>   you run html template generation with debug flag
>   console.log message should contain something about `Babel`
> * with `compile` flag `true`
>   you run html template generation with debug flag
>   console.log message should not contain something about `Babel`
> 
> Why? We know html-template as some big files and babel warns about it in logs, when transpilation runs. Now, we also know that we do not need to run transpilation as `__transpiled` dir is there. This is why we can, with such integrations test, confirm if transpilation is triggered only if `compile` is `true`
> 
> makes sense?

looks alright

derberg:
@utnim2 ok then, ping me when ready for another review

derberg:
@utnim2 do you continue with this one?

utnim2:
> @utnim2 do you continue with this one?

currently i am having my semester exam, will continue working on it after my exams are over 
 cc @derberg

Gmin2:
Hey @derberg, i guess compile flag is not working as expected

image

the integration test fails here and  logs does not contain something about `Babel`, and generating `react-template` test also failing currently investgating it

reference of test taken from [this](https://dev.to/zirkelc/how-to-test-console-log-5fhd)

lib/renderer/react.js
```diff
@@ -13,6 +13,7 @@ const reactExport = module.exports;
  * @param {string} templateLocation located for thetemplate
  * @param {string} templateContentDir where the template content are located
  * @param {string} transpiledTemplateLocation folder for the transpiled code
+ * @param {Boolean} compile Whether to compile the template files or used the cached transpiled version provided by the template in the '__transpiled' folder
```
derberg, line 16: did you do some changes in below code? cause now this jsdoc is not reflecting reality 🤔

test/integration.test.js
```diff
@@ -55,4 +55,38 @@ describe('Integration testing generateFromFile() to make sure the result of the
...
+
+  it('generates without transpilation when compile flag is false', async () => {
+    const outputDir = generateFolderName();
+    const generator = new Generator('@asyncapi/html-template@2.3.0', outputDir, {
+      forceWrite: true,
+      compile: false,
+      debug: true,
+    });
+  
+    const logSpy = jest.spyOn(console, 'log').mockImplementation(() => {});
```
derberg, line 67: to be consistent in project, for spying on console, please follow the same approach we have in `generator.test.js`

log.debug = jest.fn();
expect(log.debug).toHaveBeenCalledWith("redacted for brevity));

test/integration.test.js
```diff
@@ -55,4 +55,38 @@ describe('Integration testing generateFromFile() to make sure the result of the
     const file = await readFile(path.join(outputDir, testOutputFile), 'utf8');
     expect(file).toMatchSnapshot();
   });
+
+  it('generates without transpilation when compile flag is false', async () => {
+    const outputDir = generateFolderName();
+    const generator = new Generator('@asyncapi/html-template@2.3.0', outputDir, {
+      forceWrite: true,
+      compile: false,
```
derberg, line 63: it is false by default, so in test you should also not pass it

derberg (commented):
since last time, there were changes in integration tests and we should not rely on html-template anymore, you need to use https://github.com/asyncapi/generator/tree/master/test/test-templates/react-template

and yeah, in that template there are no babel logs we can rely on.

so maybe the best will  be to simply extend `configureReact` with a log message (only if debug flag) that basically drops log like "Transpilation of files DIR into OUTPUT_DIR started"

then probably enough is to add just unit test in generator.test.js, check but probably integration test is not needed


---
one more problem, `compile` as you put in title, should rerun transpilation. So if someone makes some modification in template and needs to refresh `__transpiled` dir, they need to pass this flag. 

**so compilation always runs by default if `__transpiled` dir is not present**

we need to have this also documented then in template-development.md (maybe other docs too):
- that when people publish templates based on react, they should include `__transpiled` dir in the package
- that during development they need to pass the flag to run `compile` explicitly to refresh `__transpiled` dir when needed

Gmin2:
> so maybe the best will be to simply extend `configureReact` with a log message (only if debug flag) that basically drops log like "Transpilation of files DIR into OUTPUT_DIR started"
> 
> then probably enough is to add just unit test in generator.test.js, check but probably integration test is not needed

i am confused here, so do we need to add integration test or do we just need to add test in `generator.test.js`

cc @derberg

derberg:
> i am confused here, so do we need to add integration test or do we just need to add test in generator.test.js

just try with suggestion to add log and test for it - if it will work, no integration test is needed

Gmin2:
@derberg any idea why workflow is failing? (idk why test seems to fail)

Gmin2:
Hey @derberg when i run locally, with `compile` true then it works perfectly
image

but dont know why test is failing 
image

apps/generator/lib/renderer/react.js
```diff
@@ -13,9 +15,12 @@ const reactExport = module.exports;
  * @param {string} templateLocation located for thetemplate
  * @param {string} templateContentDir where the template content are located
  * @param {string} transpiledTemplateLocation folder for the transpiled code
+ * @param {Boolean} compile Whether to compile the template files or used the cached transpiled version provided by the template in the '__transpiled' folder
  */
 reactExport.configureReact = async (templateLocation, templateContentDir, transpiledTemplateLocation) => {
   const outputDir = path.resolve(templateLocation, `./${transpiledTemplateLocation}`);
+  // console.log(`Transpilation of files ${templateContentDir} into ${outputDir} started`)
```
derberg, line 22: do we need this comment?

apps/generator/test/generator.test.js
```diff
@@ -5,13 +5,17 @@ const Generator = require('../lib/generator');
 const log = require('loglevel');
 const unixify = require('unixify');
 const dummyYAML = fs.readFileSync(path.resolve(__dirname, './docs/dummy.yml'), 'utf8');
+const { configureReact } = require('../lib/renderer/react.js');
```
derberg, line 8: we talked during meeting it is no longer needed and can be removed

apps/generator/test/generator.test.js
```diff
@@ -5,13 +5,17 @@ const Generator = require('../lib/generator');
...
 
 const logMessage = require('./../lib/logMessages.js');
 
 jest.mock('../lib/utils');
 jest.mock('../lib/filtersRegistry');
 jest.mock('../lib/hooksRegistry');
 jest.mock('../lib/templateConfigValidator');
+jest.mock('../lib/renderer/react.js', () => ({
```
derberg, line 16: we talked during meeting it is no longer needed and can be removed

apps/generator/test/integration.test.js
```diff
@@ -55,4 +57,87 @@ describe('Integration testing generateFromFile() to make sure the result of the
...
+  
+    // Clean up: remove the dynamically created temp.md.js file
+    await fsPromise.unlink(tempJsPath);
+  
+    logSpyDebug.mockRestore();
+  });
+
+  it('check if the temp.md file is not created when compile option is false', async () => {
+    jest.setTimeout(120000); 
+    const logSpyDebug = jest.spyOn(console, 'log').mockImplementation(() => {});
```
derberg, line 107: things like that not needed anymore right?

apps/generator/test/integration.test.js
```diff
@@ -4,6 +4,8 @@
 
 const { readFile } = require('fs').promises;
 const path = require('path');
+const { promises: fsPromise } = require('fs');
+const { readFileSync} = require('fs');
```
derberg, line 8: we talked during meeting is should not be like this
- 2 not needed duplicated requires, as there is one already in line 5
- you already in line 5 get access to `readFile` and now you want to also use the sync version

we need to be consistent and reuse what is already available
- so yes `readFile` always, with `await`
- extend `const { readFile } = require('fs').promises;` like `const { readFile, mkdir, writeFile } = require('fs').promises;`. `unlink` is not needed, as during meeting I said we do not need such a cleanup in test. `mkdir` probably not as well, because tbh I don't know why you create `await fsPromise.mkdir(outputDir, { recursive: true });` if generator does it anyways

derberg (commented):
left few comments, some things we talked about during meeting - you did not take into account

also, you as contributor and maintainer need to proactively check and see that PR tests are failing. Me as maintainer, if I see tests are failing, I do not even start review as my assumption is PR is not ready for review (I reviewed this time only because we talked about PR and I saw you did some commits after)

derberg:
@Gmin2 your tests are still failing

derberg:
@Gmin2 any progress?

Gmin2:
> @Gmin2 any progress?

@derberg resolve the issue now, please check PR [generator-react-sdk/#243](https://github.com/asyncapi/generator-react-sdk/pull/243)

apps/generator/lib/logMessages.js
```diff
@@ -59,5 +63,7 @@ module.exports = {
   templateSuccessfullyInstalled,
   relativeSourceFileNotGenerated,
   conditionalFilesMatched,
+  compileEnabled,
+  compileEnabled,
```
derberg, line 67: doubled export

apps/generator/cli.js
```diff
@@ -84,6 +84,7 @@ program
   })
   .option('-d, --disable-hook [hooks...]', 'disable a specific hook type or hooks from given hook type', disableHooksParser)
   .option('--debug', 'enable more specific errors in the console')
+  .option('--compile', 'compile the template files before rendering')
```
derberg, line 87: after all works remember to remove compile flag from here as we do not enable new flags in local cli

derberg (approved):

derberg:
@Gmin2 all good, I just improved the changeset a bit to make it clear what this release is all about. Please have a look for future education

PR #1186, opened 2024-04-11
feat: introduce monorepo to Generator Repo

ayushnau:
Changes Added. 
1. Turborepo Added. 
2. Converted the generator repo to app,  inside the turborepo. 
3. Added the required scripts of the turborepo to the root package.json file. 

**Related issue(s)**

Resolves https://github.com/asyncapi/generator/issues/1044


aeworxet:
@asyncapi/bounty_team

derberg (commented):
Some stuff needs update because of docs related links and automation:

- doc hooks.md need update of links pointing to GH that would break, change them to relative link from website, which will be `/docs/tools/generator/api`
- workflow update-docs-in-website needs an update of link that now point to `/generator/docs/`, and probably should point to `/generator/apps/generator/docs/`

AyushNautiyalDeveloper:
> Some stuff needs update because of docs related links and automation:
> 
>     * doc hooks.md need update of links pointing to GH that would break, change them to relative link from website, which will be `/docs/tools/generator/api`
> 
>     * workflow update-docs-in-website needs an update of link that now point to `/generator/docs/`, and probably should point to `/generator/apps/generator/docs/`

1. in first point you mean https://github.com/asyncapi/generator/blob/master/docs/api.md to https://github.com/asyncapi/generator/blob/master/apps/generator/docs/api.md

derberg:
@AyushNautiyalDeveloper no, it just should be `/docs/tools/generator/api` as I wrote, a relative link, asyncapi.com relative - add it to asyncapi.com and you will see what I mean -> https://www.asyncapi.com/docs/tools/generator/api

links in docs do not have to work in github in markdown, they need to work in website, where people read them

ayushnau:
@derberg 
1. updated all the links. 
2. The Integration tests are working now.
3. The npm ci is working fine.

package.json
```diff
@@ -1,86 +1,38 @@
...
-    "generator",
-    "markdown",
-    "codegen"
-  ],
-  "author": "Fran Mendez <fmvilas@gmail.com>",
-  "publishConfig": {
-    "access": "public"
+  "devDependencies": {
+    "prettier": "^3.2.5",
+    "turbo": "latest"
```
derberg, line 28: should we really point to latest? isn't it risky and error prone?

package.json
```diff
@@ -1,86 +1,38 @@
...
-    "bump:version": "npm --no-git-tag-version --allow-same-version version $VERSION"
-  },
-  "preferGlobal": true,
-  "bugs": {
-    "url": "https://github.com/asyncapi/generator/issues"
+    "build": "turbo build",
+    "dev": "turbo dev",
+    "test": "turbo run test",
+    "lint": "turbo lint",
+    "format": "prettier --write \"**/*.{ts,tsx,md}\"",
```
derberg, line 9: what we need it for? we do not use ts here anyway?

derberg (commented):
- why we have `apps/generator/.gitignore` instead of having one gitignore in root?

- I believe `apps/generator/assets/readme-banner.png` must stay in the root

- `apps/generator/.releaserc` should also stay in root, right? it is a release config file and should be same for all, no?

- please add some `Development` guide to readme, that explains what I need to do locally to start the project, run tests

- please also explain how pushing to npm will work now. This is current release workflow https://github.com/asyncapi/generator/blob/master/.github/workflows/if-nodejs-release.yml where for example you can see we need `npm test` available in the root, that will trigger all tests of all projects

- also, the plan was filters are not in the PR but another one, so we should not have all these filters scripts in package.json

- did you run all tests on local? are you sure we do not need to make any amendments in `test/test-project/test.sh` ?

package.json
ayushnau: made it to the current latest version.

package.json
ayushnau: removed

ayushnau:
@derberg 
1. We need gitignore in the apps/generator for the  test file that we dont want to be pushed. also adding those commands in root gitignore will not be applicable to all the apps. And also having multiple gitignore files for the turborepo is common pattern
2. updated the readmd file to correct place. Also I guess the readmd file for the generator should move to generator itself. so that we can add general
3. Added the required files to root.

Also i found that the test are not running properly with the current version. the issue is still not fixed 
finding ways to run the tests.correctly.

derberg:
@ayushnau how is it going, do you need anything?

ayushnau:
This pr is moved to https://github.com/asyncapi/generator/pull/1213

PR #1191, opened 2024-04-16
feat: support `noOverwriteGlobs` for templates based on react

### LINKED ISSUE #1128: `noOverwriteGlobs` will not work in majority of cases in templates using react
After we introduced react engine for templates rendering, this code was not changed

```js
async shouldOverwriteFile(filePath) {
    if (!Array.isArray(this.noOverwriteGlobs)) return true;
    const fileExists = await exists(path.resolve(this.targetDir, filePath));
    if (!fileExists) return true;

    return !this.noOverwriteGlobs.some(globExp => minimatch(filePath, globExp));
  }
```

the problem is that input here, `filePath` is always the `filePath` that includes a file name as it is called in the template.

In nunjuck engine things are easy, we practice to name template files exactly as they are named after they are rendered in output. So for example in `html-template` we would call a file in `template` directory by: `index.html`

In react, you have more flexibility, you can follow the same approach as in nunjuck engine, have separate files templated with react, but you will call the same file I mentioned above `index.html.js`

**Problem**

So the code I mentioned above will never work in above mentioned cases as at start the `filePath` will be `index.html.js`.
When I use generator in CLI for example, I want to pass `--no-overwrite="index.html"` as I want to make sure it do not overwrite `index.html`, I do not want to remember to ignore `index.html.js`. And even if I remember, it will not work as `const fileExists = await exists(path.resolve(this.targetDir, filePath));` will anyway return true and override my file cause `index.html.js` do not exists in output.

**Solution**

I think a bigger refactor is needed

- `noOverwriteGlobs` as it is, should accept only patterns/names that I have in output
- `shouldOverwriteFile` should not rely on filenames from template. Fix is not that easy as just remove `.js` from the filename as can be that template just contains `index.js` for example. 

Some investigation is needed. Probable we should move away from `generateFile` the usage of `shouldOverwriteFile` and move it to `renderAndWriteToFile` where for nunjucks we do it old way and for react, we grab final name of file in output somehow from react 🤔

lmgyuan:
**Description**

- add `noOverwriteGlobs` parameters to methods `saveContentToFile` and `saveRenderedReactContent`
- pass `this.noOverwriteGlobs` to `saveContentToFile`
- check the name of file to be written against the `noOverwriteGlobs` to determine whether it should be actually written to the target directory
- add additional comments to relevant codes in the codebase so they are easier for future maintainers and contributors to understand. 

**Related issue(s)**

Fixes #1128


derberg (commented):
@lmgyuan it's looking very good, now we just need a proper test to make sure it works long term. 

Maybe we can add a test to `test/integration.test.js` 🤔 
You could use https://github.com/asyncapi/template-for-generator-templates react-based template as a base for test. Generator template installation is npm based, so all features are included, so installing from GitHub link, pointing to specific commit will work (yes, we should not point to master to have stable tests). 

This template generates a number of files -> https://github.com/asyncapi/template-for-generator-templates/blob/master/template/schemas/schema.js#L13-L20

So your test could generate output with `noOverwriteGlobs` ignoring some of them, and test evaluation would be as simple reading output dir and making sure that some filenames are not there. That should be reliable enough

Thoughts?

lmgyuan:
@derberg just update the implementation!

lib/renderer/react.js
```diff
@@ -23,76 +24,95 @@ reactExport.configureReact = async (templateLocation, templateContentDir, transp
...
+  // check whether the filename should be ignored based on user's inputs
+  const shouldOverwrite = !noOverwriteGlobs.some(globExp => minimatch(finalFileName, globExp));
+
+  // Write the file only if it should not be skipped
+  if (shouldOverwrite) {
+    await writeFile(filePath, content, {
+      mode: permissions
+    });
+  } else {
+    console.log(`Skipping overwrite for: ${filePath}`);
```
derberg, line 99: all is good except of this one. Such logs should only be visible on demand, with debug flag

have a look into `generator.js` as the usage of logger, here you also should log out info with `log.debug`

lmgyuan: I just updated it! I think it should be appropriate to use `const log = require('loglevel');` here, since it's loaded in the generator into the cache already.

test/integration.test.js
```diff
@@ -48,4 +51,27 @@ describe('Integration testing generateFromFile() to make sure the result of the
...
+    if (!await exists(outputDir)) {
+      mkdirSync(outputDir, { recursive: true });
+    }
+    // eslint-disable-next-line sonarjs/no-duplicate-string
+    const testFilePath = path.join(outputDir, 'index.html');
+    await writeFile(testFilePath, '<script>const initialContent = "This should not change";</script>');
+
+    // based on the html-template documentation, the default output is index.html
+    const generator = new Generator('@asyncapi/html-template@0.28.0', outputDir, {
+      forceWrite: true,
```
derberg, line 67: if you also pass `debug: "true` here then in this test we can also validate that proper debug info is shown to the user

```
expect(console.log).toHaveBeenCalledWith(`Skipping overwrite for ${testFilePath}`);
```

yeah, something like that should work but you of course need to check, I just provide pseudo code

lmgyuan:
> @lmgyuan you'll need to solve conflicts first before you amend your integration tests

Thanks! I will come back and take a closer look at this after my exams and projects are over : )

test/integration.test.js
```diff
@@ -55,4 +57,28 @@ describe('Integration testing generateFromFile() to make sure the result of the
...
+
+  it('should ignore specified files with noOverwriteGlobs', async () => {
+    const outputDir = generateFolderName();
+    // Manually create a file to test if it's not overwritten
+    if (!await exists(outputDir)) {
+      mkdirSync(outputDir, { recursive: true });
+    }
+    // eslint-disable-next-line sonarjs/no-duplicate-string
+    const testFilePath = path.join(outputDir, testOutputFile);
+    await writeFile(testFilePath, '<script>const initialContent = "This should not change";</script>');
```
derberg, line 69: isn't it better to add the string to variable and not repeat it in line 82?

test/integration.test.js
derberg: I think you missed this comment

test/integration.test.js
```diff
@@ -55,4 +57,28 @@ describe('Integration testing generateFromFile() to make sure the result of the
     const file = await readFile(path.join(outputDir, testOutputFile), 'utf8');
     expect(file).toMatchSnapshot();
   });
+
+  it('should ignore specified files with noOverwriteGlobs', async () => {
+    const outputDir = generateFolderName();
+    // Manually create a file to test if it's not overwritten
```
derberg, line 63:
```suggestion
    // Manually create an output first, before generation, with additional custom file to validate if later it is still there, not overwritten
```
```suggestion
    // Manually create a file to test if it's not overwritten
```

test/integration.test.js
```diff
@@ -55,4 +57,28 @@ describe('Integration testing generateFromFile() to make sure the result of the
...
+    if (!await exists(outputDir)) {
+      mkdirSync(outputDir, { recursive: true });
+    }
+    // eslint-disable-next-line sonarjs/no-duplicate-string
+    const testFilePath = path.join(outputDir, testOutputFile);
+    await writeFile(testFilePath, '<script>const initialContent = "This should not change";</script>');
+
+    // based on the html-template documentation, the default output is index.html
```
derberg, line 71: not relevant anymore

derberg (commented):
all good, just few minor comments

test/integration.test.js
lmgyuan: is ``Skipping overwrite for ${testFilePath}`` the exact message that will be displayed when we add `debug:true` as an argument to the generator?

test/integration.test.js
```diff
@@ -3,25 +3,38 @@
...
+    // normalize the path for cross-platform compatibility
+    // return path.normalize(path.resolve(mainTestResultPath, crypto.randomBytes(4).toString('hex')));
     return path.resolve(mainTestResultPath, crypto.randomBytes(4).toString('hex'));
   };
 
   jest.setTimeout(60000);
   const testOutputFile = 'test-file.md';
 
+  beforeAll(() => {
+    if (!existsSync(path.join(reactTemplate, 'package.json'))) {
```
derberg, line 33: why we need this?

test/integration.test.js
```diff
@@ -55,4 +68,39 @@ describe('Integration testing generateFromFile() to make sure the result of the
     const file = await readFile(path.join(outputDir, testOutputFile), 'utf8');
     expect(file).toMatchSnapshot();
   });
+
+  it('should ignore specified files with noOverwriteGlobs', async () => {
+    const logSpyDebug = jest.spyOn(log, 'debug').mockImplementation(() => {});
```
derberg, line 73: why not `log.debug = jest.fn();` and then `expect(log.debug)`, then you do not need to do `mockRestore()` imho

test/integration.test.js
```diff
@@ -55,4 +68,39 @@ describe('Integration testing generateFromFile() to make sure the result of the
...
+    });
+
+    await generator.generateFromFile(dummySpecPath);
+
+    // Read the file to confirm it was not overwritten
+    const fileContent = await readFile(testFilePath, 'utf8');
+    // Check if the files have been overwritten
+    await expect(fileContent).toBe(testContent);
+    // Check if the log debug message was printed
+    await expect(logSpyDebug).toHaveBeenCalledWith(logMessage.skipOverwrite(testFilePath));
```
derberg, line 101: expect is not async call, so no need for `await`

test/integration.test.js
```diff
@@ -55,4 +68,39 @@ describe('Integration testing generateFromFile() to make sure the result of the
...
     expect(file).toMatchSnapshot();
   });
+
+  it('should ignore specified files with noOverwriteGlobs', async () => {
+    const logSpyDebug = jest.spyOn(log, 'debug').mockImplementation(() => {});
+    // log.debug = jest.fn(); 
+
+    const outputDir = generateFolderName();
+    // Manually create a file to test if it's not overwritten
+    if (!await exists(outputDir)) {
```
derberg, line 78: actually I don't get why we need to validate if it exists, if a line earlier it is created

test/integration.test.js
lmgyuan: I tried that. jest.fn() does not work but spyOn() does. The former does not actually track the log.debug() calls in other files, but the latter does by creating a [layer of observation around the existing methods](https://medium.com/javascript-journal-unlocking-project-potential/jest-fn-vs-jest-spyon-understanding-the-differences-for-effective-testing-0f5928f7a411).

test/integration.test.js
lmgyuan: Deleted. I used it during development because I could not find the package.json sometimes.

test/integration.test.js
lmgyuan:
```
const generateFolderName = () => {
    // you always want to generate to new directory to make sure test runs in clear environment
    // normalize the path for cross-platform compatibility
    // return path.normalize(path.resolve(mainTestResultPath, crypto.randomBytes(4).toString('hex')));
    return path.resolve(mainTestResultPath, crypto.randomBytes(4).toString('hex'));
  };
```

If I understand it correctly, `generateFolderName()` does not actually generate the directory, but just creates and returns a path string. Thus I manually checks and creates it.

lmgyuan: I see that in other tests you have written, you did not check and create the directory. Did you not have directory or path not found issue? If I comment out line 78-80, I have 
```
ENOENT: no such file or directory, open '/Users/yuanyuan/workspace/generator/test/temp/integrationTestResult/6224bfdd/test-file.md'
```

lmgyuan:
continued and merged in #1234

PR #1202, opened 2024-04-25
test: remove dependency to html-template in integration tests

derberg:
The issue is that all our integration tests depended on html-template

It was fine just until this week when we did generator v2 release - and now our tests will not work until html-template actually starts supporting v2 - it is because we have a functionality that templates can define what generator versions they depend on -> https://github.com/asyncapi/java-template/blob/master/package.json#L102

This PR:
- removes `html-template` from tests
- introduces new local templates, simplistic for react and nunjucks - and now these are used in tests
- local verdaccio (npm registry) is extended to work with all integration tests (not registry tests only) so we have a place where we can during tests publish the local templates (for real production testing)
- eslint had to be reconfigured a bit due to introduction of react template in the code - so we can lint it properly

**Side effect**: we have local test templates that along the way we can extend with more features and extend integration tests to test more functionality.


jonaslagoni (approved):

derberg:
@Florence-Njeri I updated readme in test folder so you need to approve as well

Florence-Njeri (approved):

lmgyuan:
I understand why an alternative simplistic test template is needed to replace HTML template. But my concern is that will using a local template require maintainers always to update it to avoid situations where the tests pass but the generator is not working in the actual usage? If this is temporary, and we will update the HTML template and others to support the generator v2 release, then I think using a local simplistic template for now is understandable. : ) @derberg

derberg:
@lmgyuan the reality is that we did not use html-template in a way that there were some great features to test. So local templates give us more flexibility actually, more can be tested long term.

Do why we didn't do it from the beginning? I added first integration tests when there was no use case for testing installation from nom registry. So there was no use case to figure there is something like verdaccio. So I neglected local templates as templates that do not enable us to test all scenarios. Now with verdaccio it is super easy and we can test basically any scenerio.

/rtm

lmgyuan:
> @lmgyuan the reality is that we did not use html-template in a way that there were some great features to test. So local templates give us more flexibility actually, more can be tested long term.
> 
> Do why we didn't do it from the beginning? I added first integration tests when there was no use case for testing installation from nom registry. So there was no use case to figure there is something like verdaccio. So I neglected local templates as templates that do not enable us to test all scenarios. Now with verdaccio it is super easy and we can test basically any scenerio.
> 
> /rtm

This makes a lot of sense! Thanks for your explanation!

PR #1207, opened 2024-05-06
fix: refactor `asyncapiDocument.server()` to `asyncapiDocument.servers().get()` to fix `conditionalFiles`

### LINKED ISSUE #1206: `conditionalFiles` feature is broken when template uses `server` special parameter
error is in https://github.com/asyncapi/generator/blob/master/lib/generator.js#L882

```
Generator Error: asyncapiDocument.server is not a function
```

It is because there is a new function that should be used

instead of `asyncapiDocument.server(this.templateParams.server)`

should be `asyncapiDocument.servers().get(this.templateParams.server)`

ibishal:
removed `asyncapiDocument.server(this.templateParams.server)` and used `asyncapiDocument.servers().get(this.templateParams.server)`

**Related issue(s)**
Fixes #1206

### DIFF
--- a/lib/generator.js
+++ b/lib/generator.js
@@ -879,7 +879,7 @@ class Generator {
     if (!shouldOverwriteFile) return;
 
     if (this.templateConfig.conditionalFiles && this.templateConfig.conditionalFiles[relativeSourceFile]) {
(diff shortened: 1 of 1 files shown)


derberg (commented):
@ibishal please bring back semicolon as linter validation fails

derberg (approved):

PR #1227, opened 2024-07-03
feat: enable `noOverwriteGlobs` for templates based on react engine

lmgyuan:
Description

add noOverwriteGlobs parameters to methods saveContentToFile and saveRenderedReactContent
pass this.noOverwriteGlobs to saveContentToFile
check the name of file to be written against the noOverwriteGlobs to determine whether it should be actually written to the target directory
add additional comments to relevant codes in the codebase so they are easier for future maintainers and contributors to understand.
Related issue(s)

Fixes https://github.com/asyncapi/generator/issues/1128


derberg (commented):
@lmgyuan please have a look on the PR title, how I changed it - it is always very important to have good meaningful titles as we use them as commit messages and they help a lot long term, with maintenance, when digging into history - bug hunting 😄 

also please remove `.idea` from that PR - add it to `.gitignore` later

although I think you didn't update the PR after our last meeting

derberg:
@lmgyuan  you need to solve conflicts

PR #1234, opened 2024-07-30
feat: enable noOverwriteGlobs for templates based on react engine 

lmgyuan:
**Description**
Description

add noOverwriteGlobs parameters to methods saveContentToFile and saveRenderedReactContent
pass this.noOverwriteGlobs to saveContentToFile
check the name of file to be written against the noOverwriteGlobs to determine whether it should be actually written to the target directory
add additional comments to relevant codes in the codebase so they are easier for future maintainers and contributors to understand.

**Related issue(s)**

Fixes https://github.com/asyncapi/generator/issues/1128


.changeset/enable_noOverwriteGlobs.md
```diff
@@ -0,0 +1,5 @@
+---
+"@asyncapi/generator": minor
+---
+
+Enable noOverwriteGlobs option in the generator.
```
derberg, line 5:
```suggestion
Enable `noOverwriteGlobs` option for templates based on react rendering engine.
```

apps/generator/lib/generator.js
```diff
@@ -194,9 +194,9 @@ class Generator {
    * @return {Promise<void>} A Promise that resolves when the generation is completed.
    */
   async generate(asyncapiDocument, parseOptions = {}) {
-    this.validateAsyncAPIDocument(asyncapiDocument);
+    await this.validateAsyncAPIDocument(asyncapiDocument);
     await this.setupOutput();
-    this.setLogLevel();
+    await this.setLogLevel();
```
derberg, line 199: setLogLevel is not an async function, no need to add await

apps/generator/lib/generator.js
```diff
@@ -194,9 +194,9 @@ class Generator {
    * @return {Promise<void>} A Promise that resolves when the generation is completed.
    */
   async generate(asyncapiDocument, parseOptions = {}) {
-    this.validateAsyncAPIDocument(asyncapiDocument);
+    await this.validateAsyncAPIDocument(asyncapiDocument);
```
derberg, line 197: validateAsyncAPIDocument is not an async function, no need to add await

apps/generator/test/integration.test.js
```diff
@@ -3,6 +3,8 @@
  */
 
 const { readFile } = require('fs').promises;
+const { promises: fsPromise } = require('fs');
+const { readFileSync} = require('fs');
```
derberg, line 7: please don't do it this way, in line 5 you already have a require where readFile in promise version is included, you can extend it with `mkdir` and others if needed, like `const { mkdir, writeFile, readFile } = require('fs').promises;`

lmgyuan:
Updates added. Suggestions resolved.

apps/generator/test/integration.test.js
```diff
@@ -55,4 +55,36 @@ describe('Integration testing generateFromFile() to make sure the result of the
...
+    // Manually create a file to test if it's not overwritten
+    await mkdir(outputDir, { recursive: true });
+    // Create a variable to store the file content
+    const testContent = '<script>const initialContent = "This should not change";</script>';
+    // eslint-disable-next-line sonarjs/no-duplicate-string
+    const testFilePath = path.normalize(path.resolve(outputDir, testOutputFile));
+    await writeFile(testFilePath, testContent);
+    const fileContentBefore = await readFile(testFilePath, 'utf8');
+    // Check if the file was created
+    expect(fileContentBefore).toBe(testContent);
```
derberg, line 70: do we really need line 68-70? seems like we are testing here not generator code but just the FS write function  🤔

lmgyuan: I would think so? The purpose of `should ignore specified files with noOverwriteGlobs` test is to make sure that the files specified by `noOverwriteGlobs` is not overwritten. Thus we need to check the file content after the generation.

Florence-Njeri (dismissed):

apps/generator/test/integration.test.js
derberg: > Thus we need to check the file content after the generation.

exactly, and this is what lines 82-84 are for, where you check if content is still there after generation

in line 68-70 you basically test `fs`, if `writeFile` was successful, which I don't think is necessary

derberg (approved):

PR #1246, opened 2024-08-02
chore: update the readme path in the docker ci

ayushnau:
Changes Added.
1. updated the readme.md path in the docker release ci.
2. Added the git checkout action in the docker ci.

Related issue(s)

Related to: https://github.com/asyncapi/generator/issues/1044


.github/workflows/release-docker.yml
```diff
@@ -10,6 +10,9 @@ jobs:
     name: Docker build and push
     runs-on: ubuntu-latest
     steps:
+      - name: Check out the repository
```
derberg, line 13: make sure it runs only before the step related to readme and use the latest checkout action

ayushnau: done please check.

derberg: now you do checkout twice, and still with old version of the action

derberg (approved):

PR #1247, opened 2024-08-03
chore: use @asyncapi/multi-parser instead of @smoya/multi-parser

smoya:
**Description**

[@smoya/multi-parser](https://github.com/smoya/multi-parser-js) has been deprecated. Now its source code is included and released as part of https://github.com/asyncapi/parser-js monorepo.


derberg (approved):
Thanks man!

PR #1258, opened 2024-08-21
fix: remove `source-map-support` package

### LINKED ISSUE #1067: Installing 'source-map-support' package results in wrong call stack
#### Describe the bug
I originally stumbled upon this bug when I started using `nestjs-asyncapi` package (https://github.com/flamewow/nestjs-asyncapi/issues/516).
Importing one of their decorators resulted in wrong call stack for failing tests.
I narrowed it down a bit to a place where `nestjs-asyncapi` imports `import Generator from '@asyncapi/generator'` - commenting out that line resulted in correct call stacks.
Then I dug deeper and found the actual cause: https://github.com/asyncapi/generator/blob/master/lib/utils.js#L155 - commenting out just that one line resulted in correct call stacks.
Apparently, https://github.com/evanw/node-source-map-support#installation-and-usage is not needed anymore for Node >=12.12.0.
My suggestion is to remove `source-map-support` package from `@asyncapi/generator`, but I don't know whether that would break anything.

P.S.: this is a second bug (https://github.com/asyncapi/generator/issues/1030) which is caused by executing code outside the constructor. From the user/developer perspective, merely importing a function or a decorator (or anything else for that matter) from a package should not DO anything until you call or apply what you imported.

#### How to Reproduce
Run a simple test (I use `jest`):
```
it('test', () => {
  expect(1).toEqual(2)
})
```
the resulting call stack is correct:
```
expect(received).toEqual(expected) // deep equality

    Expected: 2
    Received: 1

      1 | it('test', () => {
    > 2 |   expect(1).toEqual(2)
        |             ^
      3 | })
      4 |

      at Object.<anonymous> (test.spec.ts:2:13)
```

Now run the same test with `require('source-map-support').install()` line added:
```
it('test', () => {
  require('source-map-support').install()
  expect(1).toEqual(2)
})
```
the resulting call stack is incorrect:
```
    expect(received).toEqual(expected) // deep equality

    Expected: 2
    Received: 1

      2 |   require('source-map-support').install()
      3 |   expect(1).toEqual(2)
    > 4 | })
        |   ^
      5 |

      at Object.<anonymous> (test.spec.ts:4:15)
```

#### Expected behavior
Importing `@asyncapi/generator` (or any other library that imports it) should not impact the call stack of failing tests.

bakasmarius:
**Description**

- This PR removes `source-map-support` package which is not needed anymore.

**Related issue(s)**
Fixes #1067

### DIFF
--- a/.changeset/lemon-feet-fail.md
+++ b/.changeset/lemon-feet-fail.md
@@ -1,0 +1,5 @@
+---
+"@asyncapi/generator": minor
+---
+
+Remove `source-map-support` package
--- a/.changeset/lemon-feet-fail.md
+++ b/.changeset/lemon-feet-fail.md
@@ -2,4 +2,4 @@
 "@asyncapi/generator": patch
 ---
 
-Remove `source-map-support` package
+Removed the source-map-support package from the AsyncAPI Generator, as it is no longer required for version 2, which now supports Node.js version 18.12.0 and above.
--- a/.changeset/lemon-feet-fail.md
+++ b/.changeset/lemon-feet-fail.md
@@ -1,5 +1,5 @@
 ---
-"@asyncapi/generator": minor
+"@asyncapi/generator": patch
 ---
 
 Remove `source-map-support` package
--- a/apps/generator/lib/generator.js
+++ b/apps/generator/lib/generator.js
@@ -27,7 +27,6 @@ const {
   fetchSpec,
   isReactTemplate,
   isJsFile,
-  registerSourceMap,
   getTemplateDetails,
   convertCollectionToObject,
 } = require('./utils');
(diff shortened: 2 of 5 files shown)


.changeset/lemon-feet-fail.md
```diff
@@ -0,0 +1,5 @@
+---
+"@asyncapi/generator": minor
```
derberg, line 2:
```suggestion
"@asyncapi/generator": patch
```

.changeset/lemon-feet-fail.md
```diff
@@ -0,0 +1,5 @@
...
+---
+
+Remove `source-map-support` package
```
derberg, line 5:
```suggestion
Removed the source-map-support package from the AsyncAPI Generator, as it is no longer required for version 2, which now supports Node.js version 18.12.0 and above.
```

derberg (commented):
thanks for the PR!

added small adjustments suggestions to the changeset

derberg (approved):

Issue #1264, opened 2024-09-04
docs no longer pushed to website


derberg:
### Describe the bug.

after adding monorepo setup - docs update to website workflow is failing

https://github.com/asyncapi/generator/blob/master/.github/workflows/update-docs-in-website.yml#L35

the job marked with above link needs some paths update, to point to docs folder in new `apps` location

### Expected behavior

make this green: https://github.com/asyncapi/generator/actions/runs/9599491989

### Screenshots

no screen shots

### How to Reproduce

somehow run the workflow in your fork 🤔 

### 🥦 Browser

None

### 👀 Have you checked for similar open issues?

- [X] I checked and didn't find similar issue

### 🏢 Have you read the Contributing Guidelines?

- [X] I have read the [Contributing Guidelines](https://github.com/asyncapi/.github/blob/master/CONTRIBUTING.md)

### Are you willing to work on this issue ?

None

DhairyaMajmudar:
Let me give it a try, can you pls. assign @derberg

DhairyaMajmudar:
> @DhairyaMajmudar there is one more error with path in line 42, with `_section.md` file - can you open a followup PR please?

Sure thingy : )

DhairyaMajmudar:
> @DhairyaMajmudar there is one more error with path in line 42, with `_section.md` file - can you open a followup PR please?

Sure thingy : )

PR #1267, opened 2024-09-09
ci: minor fix in update docs workflow

DhairyaMajmudar:
**Description**
Fixed the path to generator docs at line 42

**Related issue(s)**
related to https://github.com/asyncapi/generator/issues/1264#issuecomment-2337884477


DhairyaMajmudar:
cc: @derberg

derberg (commented):
can you please also add another path to paths list in line 7

`.github/workflows/update-docs-in-website.yml`

so the workflow runs not only when docs are changed but also when workflow is updated?

DhairyaMajmudar:
> can you please also add another path to paths list in line 7
> 
> `.github/workflows/update-docs-in-website.yml`
> 
> so the workflow runs not only when docs are changed but also when workflow is updated?

Sure thingy, I've updated my PR : )

derberg (approved):
Thanks again!

PR #1278, opened 2024-10-03
refactor: improve generate function structure

thebedigupta:
**Description**
In this PR I resolved this [issue](https://sonarcloud.io/project/issuesopen=AZB4sejAgviDxNeZYbKR&id=asyncapi_generator) 
Forget to attach test results file [test-results.txt](https://github.com/user-attachments/files/17262915/test-results.txt)

**The Issue**
The original code used an async function directly in the Promise constructor.

**The Solution**
I've updated the code to wrap the async logic in an immediately invoked async function within the Promise executor.
I am also attching test-results file to get better know what tests I performed

**Related issue(s)**

PR #1282, opened 2024-10-04
Update cli.js   Promise executor functions should not be async

Chayan8837:
**Description**

Promise executor functions should not be async


Gmin2:
Could you also link the sonarcloud issue that the PR aims to solve

derberg:
you always need to make sure you do not duplicate work of others, there was a pr for cli.js already https://github.com/asyncapi/generator/pull/1278#pullrequestreview-2354930328

PR #1288, opened 2024-10-07
fix: overall structure and reliability of the function

thebedigupta:
**Description**
In this PR I worked on that [issue](https://sonarcloud.io/project/issues?open=AZB4sej_gviDxNeZYbKj&id=asyncapi_generator)
This is the test results that I performed [test-results.txt](https://github.com/user-attachments/files/17279940/test-results.txt)

Resolve
- The function now correctly uses Promises without mixing in async/await
- It still performs the same core functionality of walking through the hooks directory and registering each hook file,but in a more error resistant way
- lead to better performance and easier-to-understand control flow.

### DIFF
--- a/apps/generator/cli.js
+++ b/apps/generator/cli.js
@@ -1,19 +1,19 @@
 #!/usr/bin/env node
 
-const path = require("path");
-const os = require("os");
-const program = require("commander");
-const xfs = require("fs.extra");
-const { DiagnosticSeverity } = require("@asyncapi/parser/cjs");
-const packageInfo = require("./package.json");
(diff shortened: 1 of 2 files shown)


derberg (commented):
why there are so many changes in this file, can you please explain?

why function is now sync? wasn't it enough to just remove async from promise?

thebedigupta:
> why there are so many changes in this file, can you please explain?
> 
> why function is now sync? wasn't it enough to just remove async from promise?

sorry for overcomplicating the work you are right removing async is the straight-forward approach

PR #1307, opened 2024-10-18
fix: prevent potential memory leak by properly removing watcher references

Aryan4884:
**Description**

This PR addresses a potential memory leak issue in the Watcher class. When a watcher is closed, the reference to the watcher was previously set to null but not removed from the this.watchers object. Over time, this could lead to unnecessary memory consumption/memory leak.

### DIFF
--- a/apps/generator/lib/generator.js
+++ b/apps/generator/lib/generator.js
@@ -759,7 +759,7 @@ class Generator {
     // Check if the filename dictates it should be separated
     let wasSeparated = false;
     for (const prop in fileNamesForSeparation) {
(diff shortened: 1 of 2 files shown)


DhairyaMajmudar (commented):
There seems to be some unwanted formating changes which can further cause format workflow failure, can you pls. remove them by running the script `npm run lint:fix`

Aryan4884:
@DhairyaMajmudar I have removed the extra formatting changes.

derberg:
was this an issue coming from sonarcloud? of yes, please indicate it in description and provide to sonar report

Aryan4884:
No, it is just an observational fix.

derberg (changes requested):
ok, then I will close this issue, watcher is used in CLI only and we will deprecate it in few days and it doesn't make sense to invest in any changes to it

PR #1304, opened 2024-10-14
feat: add hooks library to generator

derberg:
Resolves https://github.com/asyncapi/generator/issues/1270

I added inline comments for some more context on the PR

What this PR does? from the changeset:
- Package `@asyncapi/generator-hooks` is now part of `generator` repo and won't be released separately. Source code is stored under `apps/hooks` but the package/library name stays as it was originally for backward compatibility,
- By default `@asyncapi/generator-hooks` package, known as package with a lot of different hooks used in templates, is available in the generator and you no longer have to configure it in your `package.json` in `dependencies`. Package `@asyncapi/generator-hooks` will no longer be published to NPM separately and is deprecated. You can still have your own hooks, and you can still store them in a separate package and configure with your template,
- Remember that the fact that hooks package is now included by default, doesn't mean all hooks from it are enabled by default. You still have to enable given hook in the configuration explicitly because some hooks can execute automatically without passing a specific parameter. Also hook's supported parameters also need to be defined in your template's config.

Also, next steps after it is merged:
- archive https://github.com/asyncapi/generator-hooks
- deprecate https://www.npmjs.com/package/@asyncapi/generator-hooks

PR #1308, opened 2024-10-19
fix: updating `turbo.json` file for pipelines

DhairyaMajmudar:
**Description**
-> While running the command `npm run lint:fix` warnings were thrown in the terminal due to deprecated pipeline field in turbo.json file which is now replaced by latest tasks field

-> Run script `npm run lint:fix`

-> Docs reference:- https://turbo.build/repo/docs/reference/configuration#defining-tasks

-> Screenshots

| Before | After |
|-----------|--------|
|image|image|

cc: @derberg

### DIFF
--- a/turbo.json
+++ b/turbo.json
@@ -1,6 +1,6 @@
 {
   "$schema": "https://turbo.build/schema.json",
-  "pipeline": {
+  "tasks": {
     "build": {
       "outputs": []
     },


derberg (commented):
it is not a problem with the wrong key name, it is a problem that we use wrong `$schema` pointer that points to latest (newer version of turborepo without `pipeline`)

migrating to new version of turborepo is a separate issue that you can create. For not I close this PR as the fix is provided in https://github.com/asyncapi/generator/pull/1304/files#r1800918106

PR #1311, opened 2024-10-29
fix: updated sonar test link

toffee-k21:
**Description**

- updated link to test dir in apps/generator in https://github.com/asyncapi/generator/blob/master/.sonarcloud.properties


derberg (commented):
maybe better just do `apps/generator/test/**/*` ?

toffee-k21:
> apps/generator/test/**/*

Hey @derberg, I've updated the directory path as you suggested. Could you take a look and confirm if everything looks correct?

derberg (commented):
you didn't do it as I suggested, also `sonar.test.inclusions` - why this instead of `sonar.exclusions`

PR #1325, opened 2024-12-15
fix: Fixing docs with some replacements

### LINKED ISSUE #1322: [BUG] Cannot run the generator using the commands in the documentation
### Describe the bug.

I get an error that I cannot use the code generator just like described in the [documentation](https://www.asyncapi.com/docs/tools/generator/usage)

### Expected behavior

I expect that copy pasting commands from the documentation does work

### Screenshots

I get the exception

Generator Error: This template is not compatible with the current version of the generator (1.17.25). This template is compatible with the following version range: >=2.0.0 <3.0.0.

Bildschirmfoto 2024-12-04 um 19 12 02

## Locally installed

1. Install newest version of @asyncapi/cli@2.11.1
2. Run
```
asyncapi generate fromTemplate https://bit.ly/asyncapi @asyncapi/html-template -o example
```

## With Docker
1. Verify that there is no old docker image of asyncapi
2. Run
```
docker run --rm -it --user=root asyncapi/cli:latest generate fromTemplate https://bit.ly/asyncapi @asyncapi/html-template -o example
```

### 🥦 Browser

None

- [X] I checked and didn't find similar issue

- [X] I have read the [Contributing Guidelines](https://github.com/asyncapi/.github/blob/master/CONTRIBUTING.md)

No, someone else can work on it

Bharath-dev05:
issue#1322

**Description**
 replaced @asyncapi/html-template with @asyncapi/html-template@3.0.0 --use-new-generator

**Related issue(s)**
Fixes #1322

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,7 +43,7 @@ There is a large number of templates that are ready to use and are officially su
 | `@asyncapi/java-spring-template`              | Generates Java Spring service                                   | [click here](https://github.com/asyncapi/java-spring-template)              |
 | `@asyncapi/java-spring-cloud-stream-template` | Generates Java Spring Cloud Stream service                      | [click here](https://github.com/asyncapi/java-spring-cloud-stream-template) |
 | `@asyncapi/python-paho-template`              | Generates Python service that uses Paho library                 | [click here](https://github.com/asyncapi/python-paho-template)              |
-| `@asyncapi/html-template`                     | Generates HTML documentation site                               | [click here](https://github.com/asyncapi/html-template)                     |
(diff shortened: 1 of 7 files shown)


derberg:
Sorry but to help maintainers in open source you need to make a bit more effort than just simple `find and replace`. 
I will close this PR, if you are still interested in helping, which would be great, please, after doing `find and replace`, actually read the docs that you modified. Not everywhere the change needs to be done. Please take time to understand the problem and then apply the fix.

side not that is not obvious: `api.md` is a generated file, from code comments, so do not modify it, but modify code comments where needed

DhairyaMajmudar:
Giving this issue a try 🚀

PR #1339, opened 2025-01-23
ci:  skip unnecessary test runs for bot-generated and non-critical PRs

### LINKED ISSUE #1335: Improve testing speed by skipping it for certain PRs
We have https://github.com/asyncapi/generator/blob/master/.github/workflows/pr-testing-with-test-project.yml that is really long running.

And it runs sometimes when it is not needed, when there are PRs from bots doing some simple changes, sometimes readme only.

Add to the workflow `if` statement that we already have in https://github.com/asyncapi/generator/blob/master/.github/workflows/if-nodejs-pr-testing.yml#L21

AdityaP700:
### Description
- Updated the pr-testing-with-test-project.yml workflow to refine triggers and exclude irrelevant changes.
Related Issue(s)
Fixes #1335

## Testing
- Verified the solution by making minor changes in README.md and CODE_OF_CONDUCT.md and testing it by making a PR .
- Confirmed that tests for Node versions 18 and 20 were successfully skipped as intended .


AdityaP700:
Hey @derberg ,done with the changes in the required file , do let me know of any further changes . Thanks

ItshMoh:
LGTM :)

ItshMoh:
Hey @derberg  can you run the workflow.

.github/workflows/pr-testing-with-test-project.yml
```diff
@@ -1,13 +1,37 @@
...
+        id: check
+        run: |
+          # Skip for bot PRs with specific titles
+          if [[ "${{ github.actor }}" == "asyncapi-bot" && "${{ github.event.pull_request.title }}" =~ ^(ci:\ update|chore\(release\):) ]] || \
+             [[ "${{ github.actor }}" == "allcontributors[bot]" && "${{ github.event.pull_request.title }}" =~ ^docs: ]]; then
+            echo "run_tests=false" >> $GITHUB_OUTPUT
+          else
+            echo "run_tests=true" >> $GITHUB_OUTPUT
+          fi
+
```
DhairyaMajmudar, line 31: Adding a new job on which the actual job to be done is dependent won't increase job setup time? 

Instead, how's about adding an if clause just like it is done [here](https://github.com/asyncapi/generator/blob/master/.github/workflows/if-nodejs-pr-testing.yml)

.github/workflows/pr-testing-with-test-project.yml
```diff
@@ -1,13 +1,37 @@
...
   test:
-    if: github.event.pull_request.draft == false
+    needs: should-test
+    if: needs.should-test.outputs.run_tests == 'true' && github.event.pull_request.draft == false
```
DhairyaMajmudar, line 34:
```suggestion
    if: needs.should-test.outputs.run_tests == true && github.event.pull_request.draft == false
```

wrapping `true` in single quotes makes it string, although eventually, it results in a true statement (as strings are also considered to be true) so the workflow works as expected. But it's a good practice to use Booleans for truthy and falsy statements.

DhairyaMajmudar (commented):
Amazing PR @AdityaP700 just few suggestions to improve it more.

.github/workflows/pr-testing-with-test-project.yml
```diff
@@ -1,13 +1,37 @@
-#This workflow runs the tests in the test projects to make sure the generator works as a library where it is a Node dependency along with the template.
```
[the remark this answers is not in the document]
AdityaP700: Alright will fix it!!

.github/workflows/pr-testing-with-test-project.yml
AdityaP700: Alright got it !! i took the assumptions so did such changes but anyway will resolve it .

.github/workflows/pr-testing-with-test-project.yml
AdityaP700: I did tried adding but thought of optimising it a bit thus made it likewise ,anyway will look into it .

README.md
```diff
@@ -8,7 +8,7 @@ This is a Monorepo managed using [Turborepo](https://turbo.build/) and contains
 
 1. [Hooks](apps/hooks): This library contains generator filters that can be reused across multiple templates, helping to avoid redundant work. Hooks are designed to let template developers hook into the template generation process. For example, one can create a hook code that will be automatically invoked right after the template generation process has ended.
 
-1. [Nunjucks-filters](apps/nunjucks-filters): This library contains generator filters that can be reused across multiple templates, helping to avoid redundant work. These filters are designed specifically for Nunjucks templates and are included by default with the generator, so there's no need to add them to dependencies separately.
+1. [Nunjucks-filters](apps/nunjucks-filters): This library contains generator filters that can be reused across multiple templates helping to avoid redundant work. The filters are designed specifically for Nunjucks templates and are included by default with the generator, so there's no need to add them to dependencies separately.
```
derberg, line 11: This change is not needed

.github/workflows/pr-testing-with-test-project.yml
```diff
@@ -4,10 +4,17 @@ name: Test using test project
...
+      - 'package-lock.json'
+      - 'apps/**'
+      - 'packages/**'
+      - '.github/workflows/**'
+      - '!**/*.md'
+      - '!docs/**'
 
 jobs:
   test:
-    if: github.event.pull_request.draft == false
```
derberg, line 10: this should stay, we do not want to run this on drafts

.github/workflows/pr-testing-with-test-project.yml
```diff
@@ -16,6 +23,28 @@ jobs:
...
+              startsWith(github.event.pull_request.title, 'chore(release):')
+            )) ||
+            (github.actor == 'allcontributors[bot]' && 
+              startsWith(github.event.pull_request.title, 'docs: add')
+            )
+          )
+        run: echo "shouldrun=true" >> $GITHUB_OUTPUT
+        shell: bash
+
+      - if: steps.should_run.outputs.shouldrun == true
```
derberg, line 46: let's use `'true'` like in other workflows.

technically output of `run: echo "shouldrun=true" >> $GITHUB_OUTPUT` is string -> https://docs.github.com/en/actions/sharing-automations/creating-actions/metadata-syntax-for-github-actions#outputs-for-docker-container-and-javascript-actions

so also technically please compare with true that is represented as string

.github/workflows/pr-testing-with-test-project.yml
```diff
@@ -4,10 +4,17 @@ name: Test using test project
 on:
   pull_request:
     types: [opened, reopened, synchronize, ready_for_review]
+    paths:
```
derberg, line 7: this is an interesting approach, any examples in production? how it works for others? like what was the inspiration? 

so far we used only the "inclusion" part, not "exclusion"

README.md
AdityaP700: Pardon me : will discard the commit of it !!

.github/workflows/pr-testing-with-test-project.yml
AdityaP700: Alright then ,will fix it accordingly.

.github/workflows/pr-testing-with-test-project.yml
AdityaP700: I was confused at first as everywhere i was getting the response to use it as a string but since i was adviced to use it as a boolean expression so thought to implement it . Well will update it soon

.github/workflows/pr-testing-with-test-project.yml
AdityaP700: This approach was inspired by use cases where filtering out certain paths or scenarios is critical to avoid unintended actions. The idea aligns with scenarios where 'exclusion' can complement 'inclusion' for more precise control in the reference of https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#example-including-paths : so thus had the idea of implementing the same here in the workflow.

derberg: thanks for explanation

derberg (approved):
lgtm
thanks a lot and also good proactive idea with paths filtering 👍🏼

derberg:
I adjusted the PR title with `ci` prefix, and `fix` indicates a patch release need - but now in CI setup editing case

derberg:
@all-contributors please add @AdityaP700 for infra

PR #1329, opened 2025-01-06
test: improve logging in complex project tests

achaljhawar:
Related issue(s)
#1102

PR #1344, opened 2025-01-29
refactor: change the version from '*' to their original versions

ItshMoh:
Changed the version of @asyncapi/generator-hooks, @asyncapi/generator-components, @asyncapi/generator-helpers from '*' to their original version. 

**Description**

Replaced wildcard (*) version specifier with explicit version numbers in package dependencies. While the wildcard was intended to resolve packages from local workspaces, it resulted in unnecessary npm registry lookups for unpublished packages. This change ensures proper local workspace resolution and prevents failed registry lookups, improving both reliability and performance.

**Related issue(s)**

## Summary by CodeRabbit

- **Chores**
  - Locked dependency versions to a specific release to promote consistent package behavior.
  
- **Tests**
  - Adjusted template reference paths in test cases to ensure accurate resolution during client generation.


ItshMoh:
@derberg  can you run the workflow?

derberg:
tests are failing

derberg:
please add changeset according to https://github.com/asyncapi/generator/blob/master/Development.md#release-process

ItshMoh:
Hey @derberg in the markdown file for .changeset do i have to include the patch release of those packages also who are not published on the npm registry?

.changeset/version.md
```diff
@@ -0,0 +1,11 @@
+---
+"@asyncapi/generator-hooks": patch
+"@asyncapi/nunjucks-filters": patch
+"@asyncapi/generator-components": patch
+"@asyncapi/generator-helpers": patch
```
derberg, line 5: we should release generator only and here you patch all the other packages which we do not want to include in npm

remember currently it is monorepo still publishing one package

maybe `@asyncapi/generator-components`, `@asyncapi/generator-helpers` in future will also be released separately, but for now they are highly experimental

ItshMoh: @derberg  updated the version.md.

.changeset/version.md
```diff
@@ -0,0 +1,9 @@
+---
+"@asyncapi/generator": patch
+
+---
+
+- Changed the version of `@asyncapi/generator-hooks`, `@asyncapi/generator-components`, `@asyncapi/generator-helpers` and `@asyncapi/nunjucks-filters` from '*' to their original version.
+
+- Replaced asterisk (*) version specifier with explicit version numbers in package dependencies. While the wildcard was intended to resolve packages from local workspaces, it resulted in unnecessary npm registry lookups for unpublished packages. This change ensures proper local workspace resolution and prevents failed registry lookups, improving both reliability and performance.
+
```
derberg, line 9: 😉

ItshMoh: I really don't get the emoji @derberg. Is it because in the markdown I copypasted the PR description. If  yes then this emoji suits this case

ItshMoh:
@derberg  can you run the workflow?

derberg:
@ItshMoh done

ItshMoh:
@derberg  there is still the time out issue .The error is installation failed.

derberg:
@ItshMoh please increase timeouts - we really can't do much here, this damn issues were happening on windows in the past, then we increased timeout, and looks like GH is messing with windows images again

.changeset/version.md
derberg: I don't remember 😄 maybe it was about the empty space - idk

derberg (commented):
https://github.com/asyncapi/generator/pull/1329 is failing with strange installation problems. Please update timeout here and we will first release this PR

derberg:
I finally had time to have a deeper look into logs. I pushed directly to your PR change in test to fix unit tests on windows. It was related to bad handling of paths - when systems must work on cross operating system environment, always `path.resolve` or if needed `path.join` should be used so it resolves path for a given operating system - mainly because of Windows that does paths different way

derberg:
the installation issue on the test project is for different reason

@ItshMoh please update package-lock and push to PR

apps/generator/test/test-project/test-project.test.js
```diff
@@ -18,7 +18,7 @@ const version = '0.0.1';
 console.log = jest.fn();
 
 describe('Testing if markdown was generated with proper version of the template', () => {
-  jest.setTimeout(1000000);
+  jest.setTimeout(5000000);
```
derberg, line 21: please revert

apps/generator/test/test-templates/react-template/package.json
ItshMoh, line 1: It was the pushing of generator components change not the generator-helper. the commit -message is wrong written.

derberg:
@ItshMoh any conclusions?

ItshMoh:
> @ItshMoh any conclusions?

@derberg  the changes to the generator package.json are causing the errors. As we thought that as the error was in the the react_template package.json. But that change didn't give any errors. All other changes go well. The generator changes were causing the issue.

derberg:
@ItshMoh hmm you mean setting hooks to `0.2.0`? cause tests are read - so I guess you did not revert the part that is causing issues?

ItshMoh:
Hey @derberg  I have done the changes as suggested. It's now ready for review

derberg (approved):

Issue #1346, opened 2025-01-29
create component test for existing components in the js websocket client


derberg:
We have tests for components in components library: https://github.com/asyncapi/generator/tree/master/packages/components/test/components

But we do not have tests for https://github.com/asyncapi/generator/tree/master/packages/templates/clients/js/websocket/components that should be located in https://github.com/asyncapi/generator/tree/master/packages/templates/clients/js/websocket/test

DhairyaMajmudar:
Software testing is cool and essential. Let's give it a try : )

Issue #1348, opened 2025-01-29
create test for `Requires.js`


derberg:

Karan-Palan:
Hey @derberg , shall I pick this up?

derberg:
I think PR is already opened, look at the GH info above your comment

Karan-Palan:
Ahh, just checked it. I didn't know it would add a test for all files. Tell me if you have something I can work on

derberg:
solved with https://github.com/asyncapi/generator/pull/1351

PR #1351, opened 2025-01-29
test: add tests for template websockets components

DhairyaMajmudar:
**Description**
Adding testing files for template websockets component: `Requires` and `FileHeaderInfo` 

**Related issue(s)**
fixes: #1346, #1347, #1348

Tests are running file locally

**Screenshots**

Screenshot from 2025-01-30 01-39-43


derberg (commented):
try writing tests without mocks included but reuse maybe https://github.com/asyncapi/generator/blob/master/packages/helpers/test/__fixtures__/asyncapi-websocket-query.yml like this: https://github.com/asyncapi/generator/blob/master/packages/components/test/components/models.test.js#L6


and then, if it works, that we "cross-use" fixtures between projects, maybe the file should be moved to a "generic" test fixtures folder

DhairyaMajmudar:
Updated the PR with suggested changes 👍🏻

packages/templates/clients/js/websocket/test/components/FileHeaderInfo.test.js
```diff
@@ -0,0 +1,45 @@
...
+      title: () => parsedAsyncAPIDocument.info.title,
+      version: () => parsedAsyncAPIDocument.info.version,
+    };
+
+    const serverWithPathname = {
+      protocol: () => parsedAsyncAPIDocument.servers()[3].protocol(),
+      host: () => parsedAsyncAPIDocument.servers()[3].host(),
+      pathname: () => parsedAsyncAPIDocument.servers()[3].pathname(),
+      hasPathname: () => true
+    };
```
derberg, line 28: why do you still want to maintain this code if you already have `parsedAsyncAPIDocument` and can access data using parser? 

you have this
```
const wrapperWithoutPathname = FileHeaderInfo({ info, server: serverWithoutPathname });
```

so why not refactor to this
```
const wrapperWithoutPathname = FileHeaderInfo({ parsedAsyncAPIDocument.info(), parsedAsyncAPIDocument.servers().get('withPathname') });
```

DhairyaMajmudar: Thanks for the suggestions, updated the tests with this approach.

derberg (approved):
nice! thanks

PR #1352, opened 2025-01-29
ci: integrate dorny/paths-filter into PR test workflow

AdityaP700:
### **Description**  
- Integrated `dorny/paths-filter` into the PR test workflow.  
- Updated the CI/CD workflow to conditionally run tests based on file path changes.  
- Ensured compatibility with Node.js versions 18 and 20 in the testing matrix.


AdityaP700:
@derberg Hey there, done changes as suggested by you : alternatively worked on adding https://github.com/dorny/paths-filter as i thought it will be more realiable than git inbuilt functionality and hope it would satisfy the branch protection rules. Do let me know if any further change is needed.

.github/workflows/pr-testing-with-test-project.yml
```diff
@@ -14,8 +14,28 @@ on:
       - '!docs/**'
 
 jobs:
+  changes:
+    runs-on: ubuntu-latest
+    outputs:
+      should_test: ${{ steps.filter.outputs.changes }}
+    steps:
+      - uses: actions/checkout@v3
+      - uses: dorny/paths-filter@v3
```
derberg, line 23: just one thing, for security reasons we always try to remember to not use specific release but commitId

can you switch to something like in the example please: https://github.com/asyncapi/generator/blob/master/.github/workflows/bump.yml#L19

AdityaP700:
Hey @derberg ,can you review the PR and let me know of any other thing needs to be done.

derberg (approved):

PR #1356, opened 2025-02-05
ci: exclude .all-contributorsrc from test workflow triggers

AdityaP700:
**Description**
- Added `.all-contributorsrc` to the paths exclusion list in the test workflow to prevent unnecessary test runs when only contributor list is updated


derberg:
@AdityaP700 I think you committed wrong thing. I suggest each PR you always make from new branch otherwise it remembers old commits and causes issues

AdityaP700:
@derberg my bad , i will surely keep in mind from very next time!!

derberg:
this PR still contains wrong things, please review PR also after pushing

derberg:
@AdityaP700 working on new one?

AdityaP700:
> @AdityaP700 working on new one?

I would be frank ,i did worked on it by adding the `.all-contributorsrc` in the ignored list but while testing it locally ,i wasn't getting the result so that's how it went.

derberg:
sorry, didn't get it? you mean you will not work on a new PR?

Issue #1376, opened 2025-02-27
[Draft] : Replace Deprecated `url.parse()`


Adi-204:
The current implementation of `utils.isFilePath` uses `url.parse()`, which is deprecated in Node.js with stability level 0 as mentioned in the [Node.js documentation](https://nodejs.org/api/url.html#urlparseurlstring-parsequerystring-slashesdenotehost).

Adi-204:
@derberg sir can you please provide feedback thanks : )

derberg:
closing: https://github.com/asyncapi/generator/pull/1378#issuecomment-2693679229

PR #1378, opened 2025-03-01
fix: replace the deprecated function url.parse

Adi-204:
**Description**
Replace the deprecated function url.parse as mention in [Nodejs Docs](https://nodejs.org/api/url.html#urlparseurlstring-parsequerystring-slashesdenotehost)

Test cases 
image

**Related issue(s)**
Fixes: #1376 

## Summary by CodeRabbit

- **Refactor**
  - Improved file path processing reliability by enhancing error handling to more accurately distinguish valid inputs.

### DIFF
--- a/apps/generator/lib/utils.js
+++ b/apps/generator/lib/utils.js
@@ -2,7 +2,7 @@ const fs = require('fs');
 const util = require('util');
 const path = require('path');
 const fetch = require('node-fetch');
-const url = require('url');
+const { URL } = require('url');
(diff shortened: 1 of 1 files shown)


derberg:
please do not proceed with opening PRs until the issue is discussed and accepted. While the deprecation you mentioned is reality since node 18, the thing is that related code is used only in `cli.js` that is anyway deprecated, no longer updated and will be removed around september/october. So we will not proceed further with this PR

PR #1413, opened 2025-03-14
chore: add comments in template code to avoid confusion

### LINKED ISSUE #1408: [Draft]: Ensure Template Requires Server Configuration
The current template implementation inside `packages\templates\clients\js\websocket\template` does not explicitly check for the presence of a server. Without a server, the template will not work as expected. We need to determine the best approach to handle this scenario.

Proposed Solution
1. Check if a hasServers function exists. 
2. If no such function exists, discuss whether:
   a. We should throw an error if no server is present.
   b. We should suggest a workaround (e.g., using custom parameters in templates).

The issue requires discussion.

Adi-204:
**Description**
Add comment in template code to avoid confusion. In below 2 files - 
1. packages\templates\clients\js\websocket\template\client.js.js
2. packages\templates\clients\js\websocket\template\README.md.js

**Related issue(s)**
 Fixes #1408 

## Summary by CodeRabbit

This release introduces internal documentation improvements that clarify how required parameters are automatically validated. These updates help ensure that the system's behavior is well understood during setup and maintenance. No visible changes or alterations to functionality have been made for end users.

- **Documentation**
	- Clarified internal guidelines by highlighting automatic parameter validation.

### DIFF
--- a/packages/templates/clients/js/websocket/template/README.md.js
+++ b/packages/templates/clients/js/websocket/template/README.md.js
@@ -2,6 +2,7 @@ import { File, Text } from '@asyncapi/generator-react-sdk';
 import { getClientName } from '@asyncapi/generator-helpers';
 
 export default function({ asyncapi, params }) {
+  // The generator already includes built-in checks for required parameters during template generation, so there is no need to manually validate them explicitly.
   const server = asyncapi.servers().get(params.server);
   const info = asyncapi.info();
   const clientName = getClientName(info);
--- a/packages/templates/clients/js/websocket/template/client.js.js
+++ b/packages/templates/clients/js/websocket/template/client.js.js
@@ -5,6 +5,7 @@ import { Requires } from '../components/Requires';
 import { ClientClass } from '../components/ClientClass';
 
 export default function ({ asyncapi, params }) {
+  // The generator already includes built-in checks for required parameters during template generation, so there is no need to manually validate them explicitly.
   const server = asyncapi.servers().get(params.server);
   const info = asyncapi.info();
   const title = info.title();


derberg (commented):
as you can see you had to add the same comment twice

this is a signal that better to create a reusable helper where you will have comment just once

Adi-204:
@derberg just to avoid any type of confusion a confirmation before proceeding I am creating new file comment.js inside `packages\helpers\src` which will be returning a reusable comment and we can console.log in both the file.

derberg:
no no, you create there a helper that returns a server basing on provided name

derberg:
ah crap, this is not solving the problem, as then it is not generic and could pick up any other value

sounds like the best is actually to have more protective helper - and not count on "server" functionality

I need a break 😃

Adi-204:
@derberg I think we are not on the same page, I am a bit confuse and we can maybe discuss this issue later on to figure out best possible solution.

Adi-204:
@derberg what is follow up of this PR? What kind of helper function you are thinking of, can you please describe a bit?

derberg:
sorry, need few days to follow up - too many things around

Adi-204:
@derberg ok we can solve it later 👍

derberg:
so my initial suggestion was to basically do helper

```js
function getServer(serverName) {

    // The generator already includes built-in checks for required parameters during template generation, so there is no need to manually validate them explicitly.
    const server = asyncapi.servers().get(serverName);
}
```

but they, the function is more generic, and accepts any name, not only passed as `params.server`, people do not have to use build in `server` param.

> sounds like the best is actually to have more protective helper - and not count on "server" functionality

what I meant is that it looks like the best would be to forget about `The generator already includes built-in checks for required parameters during template generation, so there is no need to manually validate them explicitly.` and perform verification inside the helper (using `hasServer`) independently from the generator core functionality.

so basically changing scope of the PR entirely, instead of adding a comment, we need a helper that is doing proper checking and therefore we do not need the comment

makes sense?

Adi-204:
@derberg my understanding - 

`
function getServer(asyncapi, serverName) {
  if(asyncapi.servers() && asyncapi.servers().has(server)){
    return asyncapi.servers().get(serverName);
  }
};
`

We can write helper something like above that can we reusable across diff templates as it takes asyncapi document and serverName as input and return the server object if it is present in document. 
But what we will return from this function is it doesn't satisfy `if` condition because 2 helpers till now are returning something. Are we planning to throw Error from helper like - `throw new Error(`Server "${serverName}" not found in the AsyncAPI document.`);`

wdyt? does it make sense?

Adi-204:
I thought maybe making a more proactive function by checking if the server is present in the asyncapi document or not. wdyt? Or we can anyways pass asyncapi.severs() directly.

derberg:
I guess you close this PR and open a new one?

Adi-204:
@derberg ok we can close this one and open new.

Issue #1414, opened 2025-03-14
[Test] Adding integration test for js websocket with renaming `snapshot.test.js`


Adi-204:
Currently we have a file name `snapshot.test.js` inside `packages\templates\clients\js\websocket\test` which has to be rename to integration.test.js as it is not meant to add only snapshot test. 
Also adding 2 more testcases in existing test -
1.  without server params
2.  without clientFileName params

Adi-204:
Waiting for approval from maintainers.

ItshMoh:
@Adi-204  you can go ahead and make a PR

PR #1415, opened 2025-03-15
test: rename snapshot.test.js to integration.test.js and add testcases

Adi-204:
**Description**
1. Rename snapshot.test.js to integration.test.js 
2. Added testcases regarding missing params
   a. without clientFileName param
   b. without server param
   
image

**Related issue(s)**
Fixes: #1414

### DIFF
--- a/packages/templates/clients/js/websocket/test/integration.test.js
+++ b/packages/templates/clients/js/websocket/test/integration.test.js
@@ -12,7 +12,7 @@ const template = './';
 
 describe('testing if generated client match snapshot', () => {
   jest.setTimeout(100000);
-  
+
   it('generate simple client for postman echo', async () => {
     const testOutputFile = 'client-postman.js';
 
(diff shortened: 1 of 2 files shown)


packages/templates/clients/js/websocket/test/integration.test.js
```diff
@@ -46,4 +46,37 @@ describe('testing if generated client match snapshot', () => {
...
+      }
+    });
+
+    await generator.generateFromFile(asyncapi_v3_path_hoppscotch);
+
+    const clientOutputFile = path.join(testResultPath, defaultOutputFile);
+    expect(clientOutputFile).toBeDefined();
+
+    const client = await readFile(clientOutputFile, 'utf8');
+    expect(client).toMatchSnapshot();
```
derberg, line 66: no need to do snapshot there 

you check only one thing here: if output client name is `client.js`

so instead of expecting snapshot match - you should just validate if file `client.js` exists

Adi-204: @derberg thanks for the guidance : )

packages/templates/clients/js/websocket/test/integration.test.js
```diff
@@ -2,6 +2,7 @@
  * @jest-environment node
  */
 
+const fs = require('fs');
```
derberg, line 5: nah, async world is better 😛 

see 2 lines below. We already require fs but the promises, so just extend the require and get `exists` and do `await exists`

Adi-204: @derberg fs.exists is completely Deprecated and we can't import also.
Ref - https://nodejs.org/api/fs.html#fsexistspath-callback
We have to either use stat or access.
When I checked the codebase we have used stat in utilis.js
Ref - https://github.com/asyncapi/generator/blob/06a97bc717b1cbe21ee4463a847e387ded742c82/apps/generator/lib/utils.js#L26

So to keep consistency I am proceeding with using stat. 
thanks : )

PR #1422, opened 2025-03-17
docs: add recommendation to run broker locally if public broker is unreachable

### LINKED ISSUE #1406: [📑 Docs]: replace test.mosquitto.org with localhost for better local testing in "creating a template documentation"
### What Dev Docs changes are you proposing?

While following the "Creating a template" tutorial, I ran into an issue when trying to connect to test.mosquitto.org using Docker.

I ran this command:
`docker run hivemq/mqtt-cli sub -t temperature/changed -h test.mosquitto.org`, mentioned in docs
but I kept getting a connection timeout error:

Image
After spending quite some time debugging, I couldn’t get it to work. Eventually, I tried running a local Mosquitto broker and switched the host to localhost, which worked perfectly.

## Why This Should Be Changed
1. test.mosquitto.org is a public broker, and it’s not always reliable. If it's down or unreachable, people following the tutorial will get stuck.
2. The error message isn’t helpful and is also not related to AsyncAPI, and debugging this takes time when the issue isn’t on the user’s end.
3. Setting up a local broker through Docker is easy and avoids these connection issues.

### Suggested Fix
Instead of using test.mosquitto.org, the docs should recommend running a local broker:

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

KunalNasa:
**Description**
This PR adds the recommendation to run a broker locally if a public broker is not reachable. It also provides guidance on the necessary changes users need to make when opting for a local broker.

**Related issue(s)**
Fixes #1406

## Summary by CodeRabbit

- **Documentation**
  - Expanded the tutorial to help users set up a local MQTT broker if the public service is unavailable.
  - Added guidance on replacing the public broker address with a local one, including a Docker command and a manual installation reference.
  - Included a note in the AsyncAPI document regarding the server URL for local broker users.


apps/generator/docs/generator-template.md
```diff
@@ -14,6 +14,18 @@ In this tutorial:
...
+> **Note:** In this tutorial, we are using `test.mosquitto.org` as the public broker. However, sometimes it may not be reachable. If you experience any difficulty connecting to it, you can run a broker on your localhost instead.  
+>  
+> If you choose to run the broker on localhost, then in the further steps, replace all occurrences of `test.mosquitto.org` with `localhost` and run the following Docker command:  
+>  
+> ```sh
+> docker run -d --name mosquitto -p 1883:1883 eclipse-mosquitto
+> ```
+>  
+> This starts an Eclipse Mosquitto broker locally on your machine, listening on port 1883.  
+>
```
ItshMoh, line 26:
```suggestion
>  To verify the broker is running properly, you can check with:
>  ```sh
>  docker ps | grep mosquitto
>  ```
>
```

ItshMoh: @KunalNasa  All looks good to me. just a small suggetion. Thanks for the PR.

KunalNasa: @ItshMoh Thanks for the review. I have merged the changes.

derberg:
lgtm, but you need to add also information that if someone decides to use local broker, then the url `url: test.mosquitto.org` field in asyncapi documents needs an update

KunalNasa:
@derberg I have added that if you choose to use local broker then you need to replace all occurrences of test.mosquitto.org with localhost. Please let me know if any further changes are needed. Thanks.

KunalNasa:
Yes @derberg It is localhost and not localhost:1883 because the default port for mqtt is 1883; therefore, the paho mqtt client assumes it is 1883 by default.

Even if the user wants to use the different port, he wouldn't require to change the url to localhost:1234 (example port), but he needs to pass the port number to the `client. connect()` function, which is a part of the `client.py` code file.

Regarding comments to the line `url : test.mosquitto.org`, it appears about 5-6 times in docs, should I add comments about using localhost at each place or add a comment only once where the line 'url : test.mosquitto.org` first appears in docs? Like url: test.mosquitto.org #in case you're using local mosquitto instance, replace this 'test.mosquitto.org' value with 'localhost' throughout the tutorial.

Adding comments all over the docs would make the docs cluttery, as discussed with @ItshMoh . Would love to hear your thoughts on this as well.

ItshMoh (changes requested):

derberg:
> Regarding comments to the line url : test.mosquitto.org, it appears about 5-6 times in docs, should I add comments about using localhost at each place or add a comment only once where the line 'url : test.mosquitto.org` first appears in docs

the most important is to just add simple comment in the example AsyncAPI document. Other mentions are not that important. But the one in AsyncAPI document is critical - as if someone forgets to do it, they will later on end up in wasting time in debugging why they are not sending/reciving messsages

KunalNasa:
Thank @derberg for the feedback. Working on it.

derberg (approved):

ItshMoh (approved):

derberg (approved):

PR #1441, opened 2025-03-25
feat: add new helper getServer

### LINKED ISSUE #1408: [Draft]: Ensure Template Requires Server Configuration

Adi-204:
**Description**
Add new helper function getServer() also written test for it.

image

**Related issue(s)**
Fixes #1408 

## Summary by CodeRabbit

- **New Features**
  - Introduced a new function for retrieving server configurations with improved error handling.
  - Updated WebSocket client templates for JavaScript and Python to utilize the new server retrieval method, enhancing clarity and efficiency.

- **Tests**
  - Added new tests to verify correct server selection and ensure robust handling of missing or invalid server configurations.

### DIFF
--- a/packages/helpers/src/servers.js
+++ b/packages/helpers/src/servers.js
@@ -27,7 +27,7 @@ const getServerUrl = (server) => {
  * @param {Map<string, object>} servers - A map of server names to AsyncAPI server objects.
  * @param {string} serverName - The name of the server to retrieve.
  * 
- * @throws {Error} If the specified server name is not found in the provided servers map.
+ * @throws {Error} If any of the parameter is missing or invalid.
  *
  * @returns {object} The AsyncAPI server object corresponding to the given server name.
  */


packages/helpers/src/servers.js
```diff
@@ -5,21 +5,42 @@
...
+ *
+ * @param {Map<string, object>} servers - A map of server names to AsyncAPI server objects.
+ * @param {string} serverName - The name of the server to retrieve.
+ * 
+ * @throws {Error} If the specified server name is not found in the provided servers map.
+ *
+ * @returns {object} The AsyncAPI server object corresponding to the given server name.
+ */
+const getServer = (servers, serverName) => {
+  if (serverName && servers.has(serverName)) {
```
derberg, line 35: servers can also be undefined

derberg: no, only this

probably `servers = parsedAsyncAPIDocument.servers();` should not be part of `beforeAll`

Adi-204:
@derberg should I throw different type of errors as suggested by coderabbit or is it ok to throw only 1 type of error.

Adi-204:
@derberg any more suggestion from your side? thanks : )

packages/helpers/src/servers.js
```diff
@@ -5,21 +5,48 @@
...
+ * @param {Map<string, object>} servers - A map of server names to AsyncAPI server objects.
+ * @param {string} serverName - The name of the server to retrieve.
+ * 
+ * @throws {Error} If the specified server name is not found in the provided servers map.
+ *
+ * @returns {object} The AsyncAPI server object corresponding to the given server name.
+ */
+const getServer = (servers, serverName) => {
+  if (!servers) {
+    throw new Error('No server found in AsyncAPI document.');
```
derberg, line 36:
```suggestion
    throw new Error('Provided AsyncAPI document doesn\'t contain Servers object.');
```

Adi-204:
@derberg changes done 👍

PR #1465, opened 2025-04-02
feat: Setup Dockerfile & Docker Compose for project-wide testing

### LINKED ISSUE #1423: [FEATURE] Use Docker Compose for Docker isolated testing
### Why do we need this improvement?

Currently, the [Development.md](https://github.com/asyncapi/generator/blob/master/Development.md#docker-isolated-testing) suggests using a manual `docker run` command for running the tests in docker in isolation. This is simple enough, but isn't ideal due to:
- A big command that needs to be copy-pasted.
- Re-running the tests will require re-installing the npm modules again, wasting time and bandwidth.
- Not easy to extend.

I suggest creating a Dockerfile and a Docker Compose to run the tests instead.

### How will this change help?

Converting to a Docker Compose will:
- Reduce the command needed to test to a simple `docker compose -f <file> up`, which is more intuitive and easier to write.
- The dockerfile caching mechanisms will ensure the npm modules are not re-installed unless required, saving significant time when re-testing.
- The dockerfile will make it more maintainable and easy to extend or configure.

### Screenshots

_No response_

### How could it be implemented/designed?

1- Dockerfile

Creating a Dockerfile at the root of the project. The dockerfile doesn't need to be complex for now, just:
- copy the package.json files of the project
- run npm install. This will now ensure that npm install will be cached.
- copy the rest of the project
- configure CMD to be npm test

2- Docker Compose

Creating a Docker Compose at the root of the project, could be named `test.compose.yaml`.
- will have a build field
- will have a command field set to `npm test`. Sounds redundant but in case the dockerfile's CMD is changed later.

3- .dockerignore

This will list the files to not be copied over to the image. An example would be node_modules.

4- Update Development.md

### 🚧 Breaking changes

No

- [x] I checked and didn't find a similar issue

### Are you willing to work on this issue?

nightknighto:
**Description**

This PR improves the Dockerized testing workflow by simplifying the command and making use of caching.

- Reduce the command needed to test to a simple `docker compose up`.
- Make use of caching mechanisms in Dockerfile, ensures the npm modules are not re-installed unless required, saving significant time when re-testing.
- Allowing the dockerized testing workflow to be more maintainable and easy to extend or configure.

**Screenshots**

**Design Choices**

Writing the dockerfile in a correct way to cache npm modules was the most difficult part of the PR, due to the monorepo structure. Some possible ways were:

1- Manually copying each package.json file from the project. Ex:
```docker
COPY package*.json ./
COPY apps/generator/package*.json ./apps/generator/
...
COPY packages\templates\clients\js\websocket\package.json packages\templates\clients\js\websocket\package.json
...
```

Problem: While manually copying package.json of each package could be ok, the folder structure of the package.json must be written manually and would need manual edit if changed. This problem is very apparent in the js websocket client, which has a deeply nested package.json. Refused this approach.

2- Using `turbo prune` command. Turborepo CLI has a command that can prune a project out of modules and divide it into a folder of package.json with respecting the nested structure, and a folder of source code. The first folder will be used to install npm modules, and the second will be to add source code.
```docker
RUN turbo prune @asyncapi/generator @asyncapi/template-js-websocket-client @asyncapi/generator-components --docker --out-dir /out
```

A small problem is that the command requires a scope (a package to prune as a target), and there is no `all` option, so we have to specify all package names. Some packages include other packages as dependencies so some names can be skipped. I've opened an issue anyways in their github: https://github.com/vercel/turborepo/discussions/10267

This approach gave great results, and is the chosen one.

**Related issue(s)**

Fixes #1423

### DIFF
--- a/Development.md
+++ b/Development.md
@@ -47,21 +47,10 @@ To run tests in an isolated Docker environment:
 2. Run the following command from the project root:
 
 ```bash
-docker run --rm -v ${PWD}:/app -w /app node:18 sh -c "
-cp -r /app /tmp/app &&
-cd /tmp/app &&
-npm install &&
-npm test
-"
+docker compose up
 ```
 
-This command above does the following:
-- Mounts the current directory to `/app` in the container
-- Copies the project to a temporary directory
-- Installs dependencies
-- Runs all tests
-
-Note: This approach ensures a clean environment for each test run by removing any existing `node_modules`.
+Note: This approach ensures a clean environment for each test run by cleanly installing dependencies and running tests in a Docker container.
 
 ### Manually testing with test templates
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,28 @@
+# Stage 1: Prepare package.json files
+FROM node:18-alpine AS base
+
+WORKDIR /app
+
+# Install turbo
+RUN npm install -g turbo@1.13.3
+
+# Mount the project temporarily to use turbo prune to get all package.json files of the project
+# We have to specify the package names. Some packages are included as dependencies in others, they will be automatically included
+RUN --mount=type=bind,source=.,target=. turbo prune @asyncapi/generator @asyncapi/template-js-websocket-client @asyncapi/generator-components --docker --out-dir /out
+
+
+# ----------------------------------------
+# Stage 2: Install dependencies
+FROM base AS final
+
+# Copy package.json files extracted by turbo prune
+COPY --from=base /out/json/ .
+COPY --from=base /out/package-lock.json ./package-lock.json
+
+# Install dependencies
+RUN npm ci
+
+# Copy the rest of the source code
+COPY --from=base /out/full/ .
+
+CMD ["npm", "test"]
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,28 +1,38 @@
-# Stage 1: Prepare package.json files
 FROM node:18-alpine AS base
 
 WORKDIR /app
 
+# ----------------------------------------
+# Stage 1: Prepare package.json files
+FROM base AS installer
+
 # Install turbo
 RUN npm install -g turbo@1.13.3
 
-# Mount the project temporarily to use turbo prune to get all package.json files of the project
-# We have to specify the package names. Some packages are included as dependencies in others, they will be automatically included
-RUN --mount=type=bind,source=.,target=. turbo prune @asyncapi/generator @asyncapi/template-js-websocket-client @asyncapi/generator-components --docker --out-dir /out
+# COPY the whole project to the container
+COPY . .
 
+# Run turbo prune to prune the project down to just package.json files of the project
+# This creates a new directory called /out with the following structure:
+# /out
+# ├── json -> package.json files of the project
+# ├── full -> full source code of the project
+# └── package-lock.json -> package-lock.json of the project
+# We have to specify the package names. Some packages are included as dependencies in others, they will be automatically included
+RUN turbo prune @asyncapi/generator @asyncapi/template-js-websocket-client @asyncapi/generator-components --docker --out-dir /out
 
 # ----------------------------------------
 # Stage 2: Install dependencies
 FROM base AS final
 
 # Copy package.json files extracted by turbo prune
-COPY --from=base /out/json/ .
-COPY --from=base /out/package-lock.json ./package-lock.json
+COPY --from=installer /out/json/ .
+COPY --from=installer /out/package-lock.json ./package-lock.json
 
-# Install dependencies
+# Install dependencies only with package.json files to make use of cache
 RUN npm ci
 
 # Copy the rest of the source code
-COPY --from=base /out/full/ .
+COPY --from=installer /out/full/ .
 
 CMD ["npm", "test"]
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,37 @@
+# Node modules
+**/node_modules
+npm-debug.log
+yarn-debug.log
+yarn-error.log
+
+# Environment variables
+.env
+.env.*
+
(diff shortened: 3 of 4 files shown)


nightknighto:
Regarding SonarCube notes, I believe the notes aren't significant due to this Docker workflow being used only for testing, and not production.

nightknighto:
@derberg What do you think about the new dev workflow?

derberg:
the problem is that SonarCloud report is red and blocking merging

and actually, Dockerfile is not only used by us for testing, it is also used by people that still use `ag` CLI

derberg:
nevermind about my comment about Dockerfile, now I get you don't want to use https://github.com/asyncapi/generator/blob/master/apps/generator/Dockerfile but have one for entire monorepo for dev purpose only

Dockerfile
```diff
@@ -0,0 +1,38 @@
...
+COPY . .
+
+# Run turbo prune to prune the project down to just package.json files of the project
+# This creates a new directory called /out with the following structure:
+# /out
+# ├── json -> package.json files of the project
+# ├── full -> full source code of the project
+# └── package-lock.json -> package-lock.json of the project
+# We have to specify the package names. Some packages are included as dependencies in others, they will be automatically included
+RUN turbo prune @asyncapi/generator @asyncapi/template-js-websocket-client @asyncapi/generator-components --docker --out-dir /out
```
derberg, line 22: instead of `RUN npm install -g turbo@1.13.3` you could probably use npx directly to mitigate one security issue

Dockerfile
```diff
@@ -0,0 +1,38 @@
...
+# ----------------------------------------
+# Stage 2: Install dependencies
+FROM base AS final
+
+# Copy package.json files extracted by turbo prune
+COPY --from=installer /out/json/ .
+COPY --from=installer /out/package-lock.json ./package-lock.json
+
+# Install dependencies only with package.json files to make use of cache
+RUN npm ci
```
derberg, line 33: this issue is pretty easy to address, just add `--ignore-scripts`

Dockerfile
```diff
@@ -0,0 +1,38 @@
...
+# /out
+# ├── json -> package.json files of the project
+# ├── full -> full source code of the project
+# └── package-lock.json -> package-lock.json of the project
+# We have to specify the package names. Some packages are included as dependencies in others, they will be automatically included
+RUN turbo prune @asyncapi/generator @asyncapi/template-js-websocket-client @asyncapi/generator-components --docker --out-dir /out
+
+# ----------------------------------------
+# Stage 2: Install dependencies
+FROM base AS final
```
derberg, line 26: probably enough to add `USER nonroot`

Dockerfile
```diff
@@ -0,0 +1,38 @@
...
+
+# ----------------------------------------
+# Stage 1: Prepare package.json files
+FROM base AS installer
+
+# Install turbo
+RUN npm install -g turbo@1.13.3
+
+# COPY the whole project to the container
+COPY . .
```
derberg, line 13: maybe there is a way to just ignore this line for sonarcube

Dockerfile
nightknighto: Using NPX works, but I don't understand why it works.

When it copies all files `COPY . .`, it ignores node_modules folder, and I don't install the modules before running the command, so what I see is that it should fail and say that it didn't find `turbo` package installed. However it works. Do you know why?

Dockerfile
nightknighto: Looks like it has to be done from their UI: https://community.sonarsource.com/t/nosonar-in-docker-file/95235

nightknighto: Could you please clarify what to do with the first link? I don't understand how the specified function can be used in our case

derberg:
```suggestion
COPY . . // NOSONAR
```

Dockerfile
```diff
@@ -0,0 +1,41 @@
...
+
+WORKDIR /app
+
+# ----------------------------------------
+# Stage 1: Prepare package.json files
+FROM base AS installer
+
+# COPY the whole project to the container
+COPY . . // NOSONAR
+
```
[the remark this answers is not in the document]
nightknighto: Found out that comments must begin in new lines, and can't be placed after a command or it will be interpreted as part of it. Fixed it in the next commit

Dockerfile
nightknighto: Unfortunately the NOSONAR didn't work, I have tried multiple syntax versions of it.
Going now with the third method discussed in our meeting, adding a comment for clarification and manually approving it in SonarQube UI.
@derberg kindly check

derberg (commented):
Screenshot 2025-04-14 at 16 32 20

please look into the chat we had in slack. 

wdyt?

nightknighto:
Added running lint optionally based on an environment variable `LINT`, with the default being not to run it. Documented it as well for windows and linux.

Development.md
```diff
@@ -47,21 +47,14 @@ To run tests in an isolated Docker environment:
...
-- Mounts the current directory to `/app` in the container
-- Copies the project to a temporary directory
-- Installs dependencies
-- Runs all tests
+You can also opt in to run the lint checks after the tests, by setting an environment variable `LINT` with any value before the command:
+   - Windows: `set LINT=true && docker compose up`
+   - Linux/macOS: `LINT=true docker compose up`
 
-Note: This approach ensures a clean environment for each test run by removing any existing `node_modules`.
+Note: This approach ensures a clean environment for each test run by cleanly installing dependencies and running tests in a Docker container.
```
derberg, line 57:
```suggestion
> This approach ensures a clean environment for each test run by cleanly installing dependencies and running tests in a Docker container.
```

better use native markdown note

derberg (approved):

Issue #1449, opened 2025-03-27
Enable alternative to `package.json` approach for storing generator config


derberg:
The work we do now on putting templates in one dir in generator monorepo shows we duplicate a lot of work putting package.json all over the place. To stop doing it we need to start supporting alternative to `package.json` as a location for the config: https://github.com/asyncapi/generator/blob/master/apps/generator/docs/configuration-file.md

My suggestion is to enable to support for `.something` file that follows YAML, just like other do with eslint or npm or any other library that has config. I don't thing at this stage we have use case for more sophisticated config that would be a JS file or something - no use case, then doesn't make sense to enable it.

suggested config file name to be consistent with other packages in JS ecosystem: `.ageneratorrc` or `.asyncapigenrc` or `.asyncapigeneratorrc`

Detailed scope:
- refactor https://github.com/asyncapi/generator/blob/master/apps/generator/lib/generator.js#L970 to support other locations of the config (also let us stop using sync functions `fs.existsSync`)
- small update in https://github.com/asyncapi/generator/blob/master/apps/generator/lib/filtersRegistry.js#L19 to mention config and not package.json explicitly
- update https://github.com/asyncapi/generator/blob/master/apps/generator/lib/generator.js#L138 to mention config in general, not package.json 
- well of course tests and docs

derberg:
@ItshMoh please take this one up

ItshMoh:
@derberg Thanks , I will be taking it

ItshMoh:
hey @derberg here the config file for example `.ageneratorrc` file should be in every template folder. As the config of the generator config can be different as user will change it for his use case. He can include some `conditionalFiles` or `nonRenderableFiles` feature also. 
So am i going in the right direction that the `.ageneratorrc` would be present in every template folder which user can change for his usecase and it will be replace this in package.json.
```
 "generator": {
    "renderer": "react",
    "apiVersion": "v3",
    "generator": ">=1.3.0 <3.0.0",
    "parameters": {
      "server": {
        "description": "The name of the server described in AsyncAPI document",
        "required": true
      },
      "clientFileName": {
        "description": "The name of the generated client file",
        "required": false,
        "default": "client.js"
      }
    }
  }
```

derberg:
yes, user should be able to put the same config in that file as he can now in `package.json`.

`generator` key can be omitted if this file is used, so it will look like
```yml
renderer: react
apiVersion: v3
generator: '>=1.3.0 <3.0.0'
parameters:
[...]

PR #1485, opened 2025-04-09
feat: add the .ageneratorrc file for setting the generator Config.

ItshMoh:
**Description**
In this PR i have added the feature for .ageneratorrc file which can be defined in the template directory and here we can write the generator config. By this file we can replace the generator key from the package.json. I have changed the `loadTemplateConfig` function to allow the .ageneratorrc file. It is just optional to add .ageneratorrc file you can also do the config settings in package.json also. 
**Related issue(s)**
#1449 

## Summary by CodeRabbit

- **New Features**
  - Introduced support for template configuration via a dedicated `.ageneratorrc` YAML file, which now takes precedence over the previous `generator` section in `package.json`.
  - Added `.ageneratorrc` configuration files to Dart, Python, and JavaScript WebSocket client templates, as well as the React test template.

- **Bug Fixes**
  - Improved error messages and guidance when required template parameters are missing, now referencing "Generator Configuration" instead of `package.json`.

- **Documentation**
  - Updated configuration documentation to reflect the new `.ageneratorrc` file approach, with clear examples and precedence rules.
  - Enhanced inline comments for better clarity regarding filter registration.

- **Chores**
  - Removed generator configuration sections from `package.json` files in affected templates, consolidating settings into `.ageneratorrc`.


ItshMoh:
hey @derberg  here I have allowed the both `.ageneratorrc` file and `package.json` for setting the config. I am thinking of also including it in `.changeset` . What's your thoughts? 
Here any other changes you expect

apps/generator/lib/generator.js
```diff
@@ -968,16 +968,30 @@ class Generator {
...
+      // First, check for .ageneratorrc file
+      const rcConfigPath = path.resolve(this.templateDir, '.ageneratorrc');
+      
+      try {
+        await fs.promises.access(rcConfigPath);
+        const yaml = await readFile(rcConfigPath, { encoding: 'utf8' });
+        const yamlConfig = require('js-yaml').load(yaml);
+        this.templateConfig = yamlConfig || {};
+      } catch (accessError) {
+        // File doesn't exist, fallback to package.json
```
nightknighto, line 980: IMHO, it would be better to explicitly check for the presence of the file and doing conditional code, instead of relying on throwing errors and catching them.

Reasons:
1- You have enclosed multiple operations in the try block, not only the `access` function. If one of the other operations produce an error such as `readFile` (unrelated to file not existing) or the `require().load()`, it could be wrongly interpreted as file not existing. This is also for the second try/catch below.

2- There are multiple nested try/catch, and I believe this reduces readability and makes it harder to maintain. It would be much easier if it was a linear flow with conditionals instead.

Proposal: replace `.access` with a method that checks for the presence of file and returns a boolean.

derberg (commented):
@ItshMoh please look into the review from @nightknighto and also the complaints from sonar

apps/generator/lib/generator.js
ItshMoh: Hey @nightknighto Sorry for late response mate. Been busy with uni exams. 
So here you are saying to remove the `.access` for checking the file and directly handling the error when reading the file.

derberg:
@ItshMoh you're having some merge conflicts

Adi-204 (commented):
@ItshMoh can you please add `.ageneratorrc` for Dart websocket client also.

ItshMoh:
> @ItshMoh can you please add `.ageneratorrc` for Dart websocket client also.

@Adi-204  done

ItshMoh:
@derberg  It's ready for the review.

apps/generator/lib/generator.js
```diff
@@ -135,7 +135,7 @@ class Generator {
         enumerable: true,
         get() {
           if (!self.templateConfig.parameters?.[key]) {
-            throw new Error(`Template parameter "${key}" has not been defined in the package.json file under generator property. Please make sure it's listed there before you use it in your template.`);
+            throw new Error(`Template parameter "${key}" has not been defined in the .ageneratorrc file. Please make sure it's listed there before you use it in your template.`);
```
derberg, line 138: I bet in future more formats will be added, wouldn't it be better to do:

```suggestion
            throw new Error(`Template parameter "${key}" has not been defined in the Generator configuration. Please make sure it's listed there before you use it in your template.`);
```

apps/generator/lib/generator.js
```diff
@@ -967,19 +967,34 @@ class Generator {
    * @private
    */
   async loadTemplateConfig() {
```
derberg, line 969: please pull it outside `generator.js`, just like we have for example template config validation in [templateConfigValidator.js](https://github.com/asyncapi/generator/blob/master/apps/generator/lib/templateConfigValidator.js)

with each PR touching we should try to move code outside this gigantic `generator.js`

maybe we need new folder? like templateConfig,  where we will have `validator.js` and `loader.js`

or maybe this should be a follow-up to cleanup?

package.json
```diff
@@ -44,5 +44,8 @@
   "workspaces": [
     "apps/*",
     "packages/**"
-  ]
+  ],
+  "dependencies": {
+    "js-yaml": "^4.1.0"
```
derberg, line 49: why in main `package.json` if this is just generator dependency?

apps/generator/docs/configuration-file.md
```diff
@@ -3,30 +3,61 @@ title: "Configuration file"
 weight: 90
 ---
 
-The `generator` property from `package.json` file must contain a JSON object that may have the following information:
+You can configure your AsyncAPI Generator template using either a dedicated `.ageneratorrc` YAML file or through the `generator` property in your `package.json` file.
```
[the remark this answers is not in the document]
ItshMoh: Done

ItshMoh:
@derberg  ready for review.

apps/generator/docs/configuration-file.md
```diff
@@ -3,30 +3,61 @@ title: "Configuration file"
 weight: 90
 ---
 
-The `generator` property from `package.json` file must contain a JSON object that may have the following information:
+You can configure your AsyncAPI Generator template using either a dedicated `.ageneratorrc` YAML file or through the `generator` property in your `package.json` file.Previously, generator configuration had to be defined in the `package.json` file. Now, you can define the configuration in a separate `.ageneratorrc` file. The configuration defined in the `.ageneratorrc` file will override any configuration defined in `package.json`. The generator will first check for the `.ageneratorrc` file in the template's root directory, and if not found, it will look for the generator config in `package.json`.
```
derberg, line 6:
```suggestion
You can configure your AsyncAPI Generator template using either a dedicated `.ageneratorrc` YAML file or through the `generator` property in your `package.json` file. Previously, generator configuration had to be defined in the `package.json` file. Now, you can define the configuration in a separate `.ageneratorrc` file. The configuration defined in the `.ageneratorrc` file will override any configuration defined in `package.json`. The generator will first check for the `.ageneratorrc` file in the template's root directory, and if not found, it will look for the generator config in `package.json`.
```

PR #1492, opened 2025-04-14
refactor: move `example.py` out of `test` folder and update README instructions

### LINKED ISSUE #1460: be consistent with examples and move `example.py` out from `test` folder
move `/packages/templates/clients/websocket/python/test/example.py` to `/packages/templates/clients/websocket/python/example.py`

as a result update:
- instructions in `/packages/templates/clients/websocket/python/README.md` accordingly
- import from https://github.com/asyncapi/generator/blob/master/packages/templates/clients/websocket/python/test/example.py#L1 must also be updated as example file will be one level up

AdeebaNizam404:
**Description**

- Moved `example.py` from `packages/templates/clients/websocket/python/test/` to `packages/templates/clients/websocket/python/`.
- Updated the import path in `example.py` to reflect the new location.
- Updated the instructions in `packages/templates/clients/websocket/python/README.md` accordingly to match the new structure.
- Tested the example script to ensure it runs correctly after the path changes

**Related issue(s)**
  Fixes #1460

## Summary by CodeRabbit

- **Documentation**
  - Updated installation instructions to reflect revised dependency paths and streamline setup guidance.
  
- **New Features**
  - Enhanced the demo script’s configuration to automatically include the correct module directory for smoother execution.


packages/templates/clients/websocket/python/example.py
```diff
@@ -1,4 +1,10 @@
+import sys
+import os
+
+sys.path.append(os.path.join(os.path.dirname(__file__), 'test'))
```
derberg, line 4: why this? wasn't it enough to update below by adding `test` like `test.temp.snapshotTestResult.client_postman`?

derberg: yeah, let is leave it like this, 

other solution would be to put empty `__init__.py` in test folder, but in case of our node project I think it is worse than what we have now in the example

derberg (approved):

Issue #1510, opened 2025-04-22
Improve `getClientName` helper to make `Client` suffix optional


derberg:
Related implementation: https://github.com/asyncapi/generator/blob/master/packages/helpers/src/utils.js

`info.title` can be `Hoppscotch Echo WebSocket Client` so the resulting class name will be `HoppscotchEchoWebSocketClientClient` which is weird 😄 

It would be best to make the `Client` suffix optional. Simple boolean switch that controls if suffix should be added or not.

But then we also need to make it possible for the users to decide, during code generation, if they want to see the suffix or not. 

Scope:
- modify existing `getClientName` helper so by default it adds `Client` suffix but with proper option passed, the suffix is not added.
- modify current clients implementation to allow user to pass a custom parameter during generation, to specify if suffix should be deleted

While writing this issue, I actually figured that the best would be to allow user to specify entirely custom client name, so they either get the one from `getClientName` or they use a custom parameter to specify a different client name.

Adi-204:
@derberg confirming the idea in package.json we are planning to add 1 more optional parameter `client` and if it have value the getClientName can return it instantly or else we return what is returned right now in code just remove Client from the back. Am I missing something?

KunalNasa:
@derberg something like this?
```/**
 * Get client name from AsyncAPI info.title
 *
 * @param {object} info - The AsyncAPI info object.
 * 
 * return {string} - The client name with "Client" appended at the end.
 */
module.exports.getClientName = (info, flag = true) => {
  const title = info.title();
  if (flag) {
    // Remove spaces, make the first letter uppercase, and add "Client" at the end
    return `${title.replace(/\s+/g, '') // Remove all spaces
      .replace(/^./, char => char.toUpperCase()) // Make the first letter uppercase
    }Client`;
  }
  // don't add Client at the end if flag is false
  return `${title.replace(/\s+/g, '') // Remove all spaces
    .replace(/^./, char => char.toUpperCase()) // Make the first letter uppercase
  }`;
};
  
  ```
or am I missing something extra as @Adi-204 is mentioning about something to add in package.json?

rohithyarramala:
@derberg  sure i will work on it

rohithyarramala:
@Adi-204  & @KunalNasa  , as i worked with a previous issue related to refactor and changing some naming convention, its a result of it , i will forward the thing , thankyou for your response

derberg:
[...]

Issue #1507, opened 2025-04-17
[BUG]: Acceptance tests for both python and javascript are failing


Adi-204:
Acceptance tests for both python and javascript are failing because of outdated paths and imports in the codebase. Start all the test containers with command `podman compose -f ./microcks-setup/microcks-podman.yml --profile tests up -d` and then run below commands.

1. Run command `podman logs websocket-acceptance-tester-js`  it will show module not found screenshot attached for ref.

![Image](https://github.com/user-attachments/assets/f827ce9f-05f2-4bb6-b6b9-55b5b723e7f8)

2. Run command `podman logs websocket-acceptance-tester-py`  it will show can't cd to `packages/templates/clients/websocket/python/test/temp/snapshotTestResult` screenshot for ref.

![Image](https://github.com/user-attachments/assets/b7aea0c0-21f8-4125-87d0-4298b40aa01a)

Adi-204:
Waiting for approval from maintainers thanks : )

Adi-204:
One more suggestion - 
Current in `microcks-podman.yaml` in tester-py in the command we are running `cd packages/templates/clients/websocket/python/test/temp/snapshotTestResult` so we are ASSUMING that `npm run test` must be run before and python generated clients must be present. If they are not present then also this test will fail directly.
Two approach - 
1. Add npm run test before cd command to make sure it always exist
2. Just give NOTE in README.md about out assumption.
Which one should we proceed with?

Adi-204:
solved #1552

PR #1513, opened 2025-04-23
refactor: improve `getClientName` helper

Adi-204:
**Description**
improve `getClientName` helper by adding 2 new params in template generation process.
1. appendClientSuffix - Add Client at the end of the class name
2. customClientName - The custom name for the generated client class

**Related issue(s)**
Fixes - #1510

## Summary by CodeRabbit

- **New Features**
  - Added options to customize the generated client class name for WebSocket clients in JavaScript, Python, and Dart templates, allowing specification of a custom client name or control over appending the "Client" suffix.
  - Updated example projects and documentation to reflect Hoppscotch as the default echo service client, with instructions for switching to Postman.
- **Bug Fixes**
  - Enhanced test coverage to validate the new client naming options and ensure correct client name generation across different configurations.


Adi-204:
@derberg any more suggestion? 
What are we planning for `js` client generations structure. In `python` and `dart` (which not merge yet) both have different folders inside `temp\snapshotTestResult` but in `js` we have directly files. For consistency purpose should we make changes in js also. It is not scope of the issue but I can do it 🙂 in this PR or maybe we can create other issue and someone can work on it. wdyt 🤔?

packages/templates/clients/websocket/javascript/package.json
```diff
@@ -67,6 +67,15 @@
         "description": "The name of the generated client file",
         "required": false,
         "default": "client.js"
+      },
+      "appendClientSuffix": {
+        "description": "Add Client at the end of the class name",
```
nightknighto, line 72:
```suggestion
        "description": "Add 'Client' suffix at the end of the class name",
```
I think the original version could be confuse people in the sense that they may not quickly understand that "Client" here refers to the english word, and not a code object.
Same for the python package.

packages/helpers/test/utils.test.js
```diff
@@ -13,12 +13,36 @@ describe('getClientName integration test with AsyncAPI', () => {
...
+
+  it('should return customClientName', () => {
+    const info = parsedAsyncAPIDocument.info();
+    const appendClientSuffix = false;
+    const customClientName = 'GeminiClient';
+
+    const clientName = getClientName(info, appendClientSuffix, customClientName);
+
+    // Example assertion: Check if the name is formatted correctly
+    expect(clientName).toBe('GeminiClient');
```
nightknighto, line 46:
```suggestion
    expect(clientName).toBe(customClientName);
```
Using the constant instead of a hard-coded value, in case someone changes the constant's value.

Adi-204:
@nightknighto changes done! @derberg any suggestion from your side?

packages/templates/clients/websocket/javascript/test/integration.test.js
```diff
@@ -20,7 +20,8 @@ describe('testing if generated client match snapshot', () => {
       forceWrite: true,
       templateParams: {
         server: 'echoServer',
-        clientFileName: testOutputFile
+        clientFileName: testOutputFile,
+        appendClientSuffix: false
```
derberg, line 24: if false, no need to explicitly pass it as default is false (provided in the config - it is used by the generator)

derberg: probably one example should have `appendClientSuffix: true`

packages/templates/clients/websocket/javascript/package.json
```diff
@@ -67,6 +67,15 @@
...
         "default": "client.js"
+      },
+      "appendClientSuffix": {
+        "description": "Add 'Client' suffix at the end of the class name",
+        "required": false,
+        "default": false
+      },
+      "customClientName": {
+        "description": "The custom name for the generated client class",
+        "required": false
```
derberg, line 78: with current implementation, these are mutually exclusive, so explain it in description

so please know that if `customClientName` is used, then `appendClientSuffix` do not take any effect

Adi-204: @derberg what about something like 
"description": "Add 'Client' suffix at the end of the class name. This option has no effect if 'customClientName' is specified.",
ok something more explicit?

derberg: sounds good

packages/templates/clients/websocket/javascript/test/integration.test.js
Adi-204: @derberg added appendClientSuffix: true in one of the testcase.

derberg: it's ok it is now

derberg (approved):

derberg (commented):
ah one more thing, I think it makes sense to now start using the generated client called `HoppscotchClient` in the examples, and in microcks

you know, it is much nicer, shorter, etc, wdyt?

derberg:
can you update this PR then? or you prefer a followup?

Adi-204:
@derberg yes I can make your changes and can I also make channges in dart client adding this 2 parameters in that one as it is merged. wdyt 🤔?

derberg:
sure

btw, another thing that must be replicated in 3 different places - something to think about

Adi-204:
@derberg summary of my last commit changes - 
1. Add 2 params in dart client also. Things are working fine.

image

2. Changes done in `example.js` of js client. Their was also one issue as `sendEchoMessage` was changed to static method but it was called like a normal class function so changed that code to - `await WSClient.sendEchoMessage(message, wsClient.websocket);`

image

3. Changes done in `example.py` of python client. The location of 1 import statement `from temp.snapshotTestResult.custom_client_hoppscotch.client import HoppscotchClient` was wrong so changed it. But we still have an error because in generated client inside function `send_message_static` we have written HoppscotchEchoWebSocketClient hardcoded and import is not written in generated file for that. Not sure why hence I didn't change anything for now.

image

Didn't change anything in microcks for now because for now we should wait to solve issue - https://github.com/asyncapi/generator/issues/1507 so that I can check properly. thanks :) 

I hope this makes your review life easy 😄!

packages/templates/clients/websocket/javascript/example.js
```diff
@@ -27,7 +27,7 @@ async function main() {
 
     while (true) {
       try {
-        await wsClient.sendEchoMessage(message);
+        await WSClient.sendEchoMessage(message, wsClient.websocket);
```
derberg, line 30: can you explain why this changed if the implementation of send message picks websocket instance if one is not provided

Adi-204: image

@derberg 
we set `this.websocket` on the instance (wsClient.websocket) when call `connect()`, but we never set it on the class itself. In a static method `this` refers to the `HoppscotchClient` itself, not to wsClient instance. 
So basically `HoppscotchClient.websocket` = undefined but `wsClient.websocket` is not undefined.

Adi-204: so | | operator doesn't really make sense because this.websocket will always be undefined and if socket not passed it will also be undefined. Should we remove this.websocket or some other approach to take?
wdyt 🤔?

derberg: we could do something like 

```
  static sendEchoMessage(message, socket) {
    socket.send(JSON.stringify(message));
  }

  sendEchoMessage(message) {
    PostmanEchoWebSocketClientClient.sendEchoMessage(message, this.websocket);
  }
```

but then you need to make sure `PostmanEchoWebSocketClientClient` value is generated.

---

let us leave what you have already and please just create a followup issue

derberg:
@allcontributors please add @Adi-204 for bug, code, ideas, infra, review, test, doc

PR #1516, opened 2025-04-28
ci: updated PR testing to run correctly based on updates

ssala034:
**Description**
A possible solution to skipping long pr test on simple doc changes. Added filter that was not there previously and some logs that will remove later.  

**Related issue(s)**
Related to #1486


.github/workflows/pr-testing-with-test-project.yml
```diff
@@ -39,7 +49,8 @@ jobs:
       - name: Determine if tests should run
         id: should_run
         if: >
-          !(
+          needs.changes.outputs.should_test == 'true' &&
```
derberg, line 52: for now please do not add this condition here but rather add a step above to log the value and the condition result, just, similar to what you did in lines 33-36

ssala034:
@derberg  the added log
Screenshot 2025-04-28 124138

derberg:
yeah, and this is weird as the config says workflows modification shoud trigger tests so modified files should be `true`

I don't know, mabe my and initial contributor english is bad and we misunderstood `predicate-quantifier` and should remove it from config

ssala034:
> yeah, and this is weird as the config says workflows modification shoud trigger tests so modified files should be `true`
> 
> I don't know, mabe my and initial contributor english is bad and we misunderstood `predicate-quantifier` and should remove it from config

Its very much possible. From my understanding, The ` predicate-quantifier: 'every' ` means that all modified files in the pull request must match the specified filters for the modified_files output to be considered true. If even one modified file does not match the filters, the output will be false.

We could change it to be ` predicate-quantifier: 'some' ` or don't use it at all. But I think the default of that is still `some`

ssala034:
I think now it works. Looking at the forked repo, I had to make sure to check if explicity true with ` variable == 'true'`. 
Screenshot 2025-04-29 084447
Screenshot 2025-04-29 084617

ssala034:
> > If even one modified file does not match the filters, the output will be false.
> 
> but this PR modifies one file and it match the filter

Yes, that is true. So the `every` filter wouldn't affect this PR. But in the future, if we have some PR that modify `.md` files and other files that modify `.js` it my not pick up the filter cuz we used `every`.  That why I set it to `some` or the default.

What made this work was explicitly does ` x == true`, which we weren't doing before.

.github/workflows/pr-testing-with-test-project.yml
```diff
@@ -8,12 +8,16 @@ on:
 jobs:
   changes:
     runs-on: ubuntu-latest
+    outputs:
+      should_test: ${{ steps.filter.outputs.modified_files == 'true' }} # might be able to access this variable ???, not sure about what it returns but check the logs, might do != '' or == 'true' or something like that
```
derberg, line 12: I think below should also work

```suggestion
      should_test: ${{ steps.filter.outputs.modified_files }}
```

ssala034: @derberg It should theoretically work, but when I tried [this commit 8c0e920](https://github.com/asyncapi/generator/commit/8c0e92097dee18a044068d06086716580ab16a43), it didn't behave as expected. We got returned `false`. 

After looking at the forked repo with the information, I say an example where it uses my current one. And since it seems to work right now. I think thats how it works. 
Screenshot 2025-04-29 122648

derberg: please commit `should_test: ${{ steps.filter.outputs.modified_files }}` as I'd like to see what step `Log the modified files output` will log out

ssala034:
@derberg added final changes

derberg (approved):

derberg:
@allcontributors please add @ssala034 for test,docs,ci

derberg:
@allcontributors please add @ssala034 for infra

PR #1520, opened 2025-04-29
test: add support for Python in FileHeaderInfo tests

### LINKED ISSUE #1447: [Test]: Add `FileHeaderInfo.test.js` for python websocket client
FileHeaderInfo component has conditions in implementation so we can include test for it.

AdeebaNizam404:
**Description**
This PR extends the FileHeaderInfo.test.js test suite to include validation for the Python WebSocket client template.

**Related issue(s)**
 Resolves #1447


packages/components/src/components/FileHeaderInfo.js
```diff
@@ -0,0 +1,45 @@
...
+      {commentChar} Protocol: {server.protocol?.() || 'Unknown'}
+      </Text>
+
+      <Text>
+      {commentChar} Host: {server.host?.() || 'Unknown'}
+      </Text>
+
+      +  {server.hasPathname?.() && (
+        <Text>
+           {commentChar} Path: {server.pathname?.() || ''}
```
Adi-204, line 36: please improve the message if `server.pathname() `doesnot exist to something more explicit like 'Not specified' rather than leaving it empty.

packages/components/src/components/FileHeaderInfo.js
```diff
@@ -0,0 +1,45 @@
...
+      <Text>{line}</Text>
+
+      <Text>{commentChar}</Text>
+
+      <Text>
+      {commentChar} {info.title?.() || 'Untitled'} - {info.version?.() || 'No version'}
+      </Text>
+
+      <Text>
+      {commentChar} Protocol: {server.protocol?.() || 'Unknown'}
```
Adi-204, line 27: Unknown might confuse users, better is to write something like 'Not specified'.

packages/components/src/components/FileHeaderInfo.js
```diff
@@ -0,0 +1,45 @@
...
+  }
+
+  return (
+    <Text>
+      <Text>{line}</Text>
+
+      <Text>{commentChar}</Text>
+
+      <Text>
+      {commentChar} {info.title?.() || 'Untitled'} - {info.version?.() || 'No version'}
```
Adi-204, line 23: please replace 'Untitled' with something more explicit like - 'No title specified'. Also in version add No version specified.

packages/components/test/components/FileHeaderInfo.test.js
```diff
@@ -1,38 +1,63 @@
...
-      />
-    );
-    const actual = result.trim();
-    expect(actual).toMatchSnapshot();
-  });
-});
+import path from 'path';
+import { render } from '@asyncapi/generator-react-sdk';
+import { Parser, fromFile } from '@asyncapi/parser';
+import { FileHeaderInfo } from '../../../src/components/FileHeaderInfo';
```
Adi-204, line 4: please update this path to '../../src/components/FileHeaderInfo.js' because test are failing.

packages/components/test/components/FileHeaderInfo.test.js
```diff
@@ -1,38 +1,63 @@
...
+
+const parser = new Parser();
+const asyncapi_websocket_query = path.resolve(__dirname, '../../../../../../helpers/test/__fixtures__/asyncapi-websocket-query.yml');
```
Adi-204, line 7: not sure if we should use fixtures from helpers. the FileHeaderInfo is used in templates. I think better might be to use fixtures from https://github.com/asyncapi/generator/tree/master/packages/templates/clients/websocket/test/__fixtures__
@derberg wdyt which fixtures should we use ? currently fixtures inside components/test is pretty basic so can't use it.

Adi-204: please correct the path should be ../../../helpers/test/__fixtures__/asyncapi-websocket-query.yml and if npm run test is successful also run npm run test:update to update snapshot before pushing code. thanks : )

Issue #1524, opened 2025-04-29
[Enhancement]: Componentize readme.md.js file completely


KunalNasa:
We have componentized only the operations part of [readme.md.js](https://github.com/asyncapi/generator/blob/master/packages/templates/clients/websocket/javascript/template/README.md.js) file. 

Need to break the leftover part into components for better readability.

Also, we have two different components under the components folder of the WebSocket JS client. One for readme file and another for client file. What should the folder hierarchy be?

Should we go with 
``` 
   components
          |
          |--client
          |      |--client components here
          |
          |-- readme
          |      |-- readme components here

```

Or is our current structure fine?

ItshMoh:
Hey @KunalNasa  
What i found is that for the README.md part three component files are used which are `Availableoperations` , `MessageExamples` and `OperationHeader`. We can organize them in the proper folder as you have suggested, but to keep in mind that if there will be any common component between client and README.md then what should we do?

KunalNasa:
Yes @ItshMoh, you are right. There can be similar components between clients and readme files. Also, there is one more problem with this structure: we will be duplicating components of readme from different protocol or language clients.

For example, there can be many components that would be similar for the Python WS readme and the JS WS readme and we don't want to duplicate them.

Therefore, in the last weekly meeting, @derberg suggested finding a solution so that we can put these readme components inside a [ packages/components ](https://github.com/asyncapi/generator/tree/master/packages/components) folder rather than creating another folder for readme under the `websocket/javascript/components` folder. 

This way we can ensure readme component reusability across different clients of different protocols and languages.

But after going through readme again I found that we can not put all the components of readme inside `packages/components` as there are so many components in readme that are hard-coded and meant particularly for js ws client and these will be different for different languages and protocols.

So, I am planning to structure it something like this. 
* Put reusable readme components inside `packages/components` folder
* Put readme components specific to client and language inside `language/components/readme`
* And if there are some components that are shared across readme and client of same language then I believe we would have to make another folder,`language/components/common`, to keep common components.

So the structure would look something like this.

```packages
    |----components
    |           |--- readme
    |                    |--- component-files-here
    |
    |
    |----template/client/websocket
                 |
                 |---- javascript
                            |--- components
                                      |
                                      |--client
                                      |      |--client components here
                                      |
                                      |-- readme
                                      |      |-- readme components here
                                      |
                                      |-- common
                                             |-- common to readme and client of language or protocol 
```

wdyt about this new structure? 🤔 
[...]

PR #1519, opened 2025-04-29
refactor: remove old FileHeaderInfo and add shared version for both JS and Python templates

AdeebaNizam404:
**Description**

Removed the old FileHeaderInfo.js and FileHeaderInfo.test.js from the JavaScript client folders.

Added a new, shared version of FileHeaderInfo.js in the components/src/components directory.

Added the corresponding test file FileHeaderInfo.test.js in the components/test/components directory.

Updated both the JavaScript (client.js.js) and Python (client.py.js) templates to use the new shared FileHeaderInfo.js.

This ensures that the same FileHeaderInfo.js is used in both the JavaScript and Python client templates.
  

Testing:

Verified that the new FileHeaderInfo.js works correctly in both templates (client.js.js and client.py.js).

All tests pass successfully.
Screenshot 2025-04-29 122137

Screenshot 2025-04-29 121636

**Related issue(s)**
 Fixes #1447 

## Summary by CodeRabbit

- **New Features**
  - Introduced a reusable file header component that formats API and server information for different programming languages.

- **Refactor**
  - Centralized the file header component for easier maintenance and consistency across JavaScript and Python templates.
  - Updated usage of the file header component to specify the target language.
  - Removed local file header components from JavaScript and Python websocket client templates.

- **Tests**
  - Updated tests to support the new language prop in the file header component.

- **Chores**
  - Added a new Git submodule for generator tooling with corresponding `.gitignore` configuration.
  - Updated package dependencies to include the latest generator packages and utilities.

PR #1526, opened 2025-05-02
chore: render all examples in ws js generated readme

### LINKED ISSUE #1522: [Enhancement]: Render all examples in generated README file.
In the generated README file of the JS WebSocket client, only the first message example from the list of available examples is being rendered. This limits visibility of other examples defined in the AsyncAPI document.

Reference Code:
[MessageExamples.js](https://github.com/asyncapi/generator/blob/master/packages/templates/clients/websocket/javascript/components/MessageExamples.js)

All message examples defined under a message object should be rendered in the README output and not just the first one.

KunalNasa:
Screenshot 2025-05-02 at 8 05 09 PM

Render all examples instead of just the first one.

Fixes #1522

## Summary by CodeRabbit

- **New Features**
	- Now displays multiple message examples instead of just one, allowing users to view all available examples for a given operation.

- **Bug Fixes**
	- Improved handling of cases with missing messages or examples, preventing display issues.

- **Tests**
	- Added comprehensive tests to ensure correct rendering of multiple message examples under various scenarios.


derberg (commented):
`MessageExamples` is not tested at the moment, so we do not know if this code works or not.

I suggest you add `MessageExamples` component dedicated tests under test/components

test cases:
- operation has no messages defined = no example should be rendered
- there are no message examples = no example should be rendered
- there is one example = one example is rendered
- there are multiple examples = all are rendered
- there are multiple messages, some with example some without = all available should be rendered

KunalNasa:
@derberg, added tests for `MessageExamples` components.
This is the only good solution that came to my mind for writing tests for this component, and it covers all the test cases you mentioned. 
Please provide feedback for tests.

Also, added the `key` prop in  `asyncAPI React's Text` component in `MessageExamples component`. Earlier I didn't add this to keep code clean, as I thought this wouldn't lead to any optimizations, but I was getting warnings at several places, so I added it to maintain consistency.

derberg (commented):
no no please remove `key`. This is one thing rabbit tries to constantly complain on. We do not do frontend here, key's not needed

and regarding test - please adjust to follow https://github.com/asyncapi/generator/blob/master/packages/templates/clients/websocket/javascript/test/components/FileHeaderInfo.test.js so we use real fixture instead of mocked  data. Such test are more reliable - no temptation to write a mock data inside test that make test pass

KunalNasa:
Thank you for the suggestions. I will remove the key prop.

Regarding the tests, I understand the preference for using real fixtures instead of mocked data. However, the current real fixture doesn't seem to cover all five of the suggested test cases (cases with no messages, no examples, or mixed message states).

Would you recommend extending the existing fixture or creating a separate one specifically designed to cover these scenarios? Or should I just append a snapshot test similar to `Fileheader.test.js` using current fixture at the end of  'MessageExamples.test.js` file? This way we can have both (use a real fixture and cover test cases with mocked data).

derberg (approved):
looks great!

also this component is another example of component where we can really have it generic for any language and do something like https://github.com/asyncapi/generator/pull/1519

but yeah, separate issue and PR

KunalNasa:
> looks great!
> 
> also this component is another example of component where we can really have it generic for any language and do something like #1519
> 
> but yeah, separate issue and PR

Hi @derberg, we already have a separate issue opened for this.  
Please have a look at #1524 issue's discussion. I have dropped some suggestions regarding the refactoring of components. Would love to hear your thoughts as well. 😊

PR #1555, opened 2025-05-09
test: add test for python constructor

### LINKED ISSUE #1542: add missing snapshot test for python `Constructor.js`
like in title

thulieblack:
Add snapshots test for the Python constructo.js

closes #1542

## Summary by CodeRabbit

- **Tests**
  - Added new test cases for the Constructor component to verify its behavior with different AsyncAPI document configurations.


Adi-204:
@thulieblack I think it would be better to keep test for constructor inside `test/components` folder. So basically you can create a component folder inside current test folder and move constructor.test.js in it.

We are also doing real world AsyncAPI document testing in templates, so I think rather than using mock testing you can use document - https://github.com/asyncapi/generator/blob/master/packages/helpers/test/__fixtures__/asyncapi-websocket-query.yml
The document has channels with query so it would be perfect for testing. I think you can also use helpers `getServerUrl` and `getQueryParams` to get serverURL and query rather that doing it manually.

@derberg any suggestions from your side?

derberg (commented):
yeah, basically follow the same I provided as feedback in https://github.com/asyncapi/generator/pull/1526

thulieblack:
AAAH okay thanks @Adi-204 and Lukasz

thulieblack:
please share your review @Adi-204 @derberg

packages/templates/clients/websocket/python/test/components/constructor.test.js
```diff
@@ -0,0 +1,48 @@
...
+    const serverUrl = getServerUrl(server);
+    const result = render(<Constructor serverUrl={serverUrl} />);
+    expect(result.trim()).toMatchSnapshot();
+  });
+
+  test('renders with serverUrl and query parameters', () => {
+    const server = getServer(servers, 'withVariables');
+    const serverUrl = getServerUrl(server);
+    const channels = parsedAsyncAPIDocument.channels();
+    const queryParams = getQueryParams(server, channels);
```
Adi-204, line 43: `getQueryParams` has only one parameter channels why are we passing server? I think error is because of this.

thulieblack: Thanks

derberg (commented):
For consistency name the file with upper case like we already do in https://github.com/asyncapi/generator/tree/master/packages/templates/clients/websocket/javascript/test/components

just rerequest review once done and we can merge it

packages/templates/clients/websocket/python/test/components/constructor.test.js
```diff
@@ -0,0 +1,48 @@
+import path from 'path';
+import { render } from '@asyncapi/generator-react-sdk';
+import { Parser, fromFile } from '@asyncapi/parser';
+import { getServer, getServerUrl, getQueryParams } from '@asyncapi/generator-helpers';
+import { Constructor } from '../../components/Constructor.js';
+
+const parser = new Parser();
+const asyncapiFilePath = path.resolve(__dirname, '../../../../../../helpers/test/__fixtures__/asyncapi-websocket-query.yml');
```
derberg, line 8:
```suggestion
const parser = new Parser();

const asyncapiFilePath = path.resolve(__dirname, '../../../../../../helpers/test/__fixtures__/asyncapi-websocket-query.yml');
```

derberg (approved):

PR #1567, opened 2025-05-19
chore: add new helper function `getInfo()`

### LINKED ISSUE #1558: add new helper function `getInfo()`
In template development we are using helpers to ensure and throw useful error messages to users. 
Introduce new helper `getInfo()` and use it in all websocket clients.

Adi-204:
**Description**
Add new helper function `getInfo()`

**Related issue(s)**
Fixes #1558 

## Summary by CodeRabbit

- **New Features**
  - Introduced a utility to validate and retrieve the `info` object from AsyncAPI documents.

- **Bug Fixes**
  - Enhanced error handling for missing or invalid `info` objects in AsyncAPI documents.

- **Tests**
  - Added tests to verify correct functionality and error handling of the new utility.

- **Refactor**
  - Updated template clients to utilize the new utility for accessing the `info` object.

### DIFF
--- a/packages/helpers/src/utils.js
+++ b/packages/helpers/src/utils.js
@@ -45,10 +45,10 @@ const listFiles = async (dir) => {
  */
 const getInfo = (asyncapi) => {
   if (!asyncapi) {
-    throw new Error('Provided AsyncAPI document is invaild.');
+    throw new Error('Make sure you pass AsyncAPI document as an argument.');
   }
   if (!asyncapi.info) {
-    throw new Error('Provided AsyncAPI document doesn\'t contain info.');
+    throw new Error('Provided AsyncAPI document doesn\'t contain Info object.');
   }
   const info = asyncapi.info();
   if (!info) {
--- a/packages/helpers/src/utils.js
+++ b/packages/helpers/src/utils.js
@@ -35,8 +35,25 @@ const listFiles = async (dir) => {
     .map(dirE => dirE.name);
 };
 
+/**
+ * Validate and retrieve the AsyncAPI info object.
+ *
+ * Throws an error if the provided AsyncAPI document has no `info` section.
+ *
+ * @param {object|null|undefined} info - The `info` object from an AsyncAPI document.
+ * @throws {Error} When `info` is `null` or `undefined`.
+ * @returns {object} The validated `info` object.
+ */
+const getInfo = (info) => {
+  if (!info) {
+    throw new Error('Provided AsyncAPI document doesn\'t contain info.');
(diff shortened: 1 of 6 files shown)


packages/helpers/src/utils.js
```diff
@@ -35,8 +35,25 @@ const listFiles = async (dir) => {
...
+ *
+ * Throws an error if the provided AsyncAPI document has no `info` section.
+ *
+ * @param {object|null|undefined} info - The `info` object from an AsyncAPI document.
+ * @throws {Error} When `info` is `null` or `undefined`.
+ * @returns {object} The validated `info` object.
+ */
+const getInfo = (info) => {
+  if (!info) {
+    throw new Error('Provided AsyncAPI document doesn\'t contain info.');
```
derberg, line 49: `getInfo(asyncapi.info())` - this is how this function is used, so it doesn't really bring much value, except for the error - but error is not accurate, as it talks about AsyncAPI not containing info, although the argument of function is info object, so might be that template developer makes an errror and do not pass the info object - which means we provide wrong error to the user

imho this is how this helper should be used, and wrap parser code: `getInfo(asyncapi)`

Adi-204:
@derberg thanks for the suggestion made some changes you can check.

packages/helpers/src/utils.js
```diff
@@ -35,8 +35,31 @@ const listFiles = async (dir) => {
...
+ * Validate and retrieve the AsyncAPI info object from an AsyncAPI document.
+ *
+ * Throws an error if the provided AsyncAPI document has no `info` section.
+ *
+ * @param {object} asyncapi - The AsyncAPI document object.
+ * @returns {object} The validated info object from the AsyncAPI document.
+ */
+const getInfo = (asyncapi) => {
+  if (!asyncapi) {
+    throw new Error('Provided AsyncAPI document is invaild.');
```
derberg, line 48:
```suggestion
    throw new Error('Make sure you pass AsyncAPI document as an argument.');
```

missing asyncapi document doesn't mean it is invalid

packages/helpers/src/utils.js
```diff
@@ -35,8 +35,31 @@ const listFiles = async (dir) => {
...
+  }
+  if (!asyncapi.info) {
+    throw new Error('Provided AsyncAPI document doesn\'t contain info.');
```
derberg, line 51:
```suggestion
    throw new Error('Provided AsyncAPI document doesn\'t contain Info object.');
```

Adi-204:
@derberg changes done thanks :)

PR #1568, opened 2025-05-21
chore: add the `getTitle` helper

### LINKED ISSUE #1559: add new helper function `getTitle()`
In template development we are using helpers to ensure and throw useful error messages to users.
Introduce new helper `getTitle()` and use it in all websocket clients.

ItshMoh:
**Description**
It adds the new helper in `utils.js` which is getTitle. It fetches the title parameter from asyncapi info object. 
It gives an error when the title is undefined or null or empty string. 

**Related issue(s)**
fixes #1559 

## Summary by CodeRabbit

- **New Features**
  - Added utilities to validate and retrieve the title and info sections from AsyncAPI documents.
  - Updated client templates to use these utilities for obtaining titles and client names.

- **Bug Fixes**
  - Enhanced error handling for missing or invalid title and info sections in AsyncAPI documents.

- **Tests**
  - Added tests verifying correct behavior and error handling of the new title retrieval utility.


ItshMoh:
currently for the test of `getTitle` in the `utils.test.js` . I have mocked for the emptyTitle case and the  noTitle case.

Adi-204 (commented):
@ItshMoh currently your `getTitle()` helper function implementation assumes that info is present in asyncapi document but it maybe case info is not present. 

In our implementation of all websocket clients we call `getInfo()` first and then `getTitle()` and code is executed line by line so it guarantee that info will exist, but I think we should not consider this assumption and better is to first call `getInfo()` from `getTitle()` so that proper error is thrown when someone is using only `getTitle()`  wdyt 🤔 ?

@derberg share your thoughts also.

ItshMoh:
> but I think we should not consider this assumption and better is to first call `getInfo()` from `getTitle()` so that proper error is thrown when someone is using only `getTitle()` wdyt 🤔 ?
@Adi-204  you are right, the `getTitle()` handles our current implementation, but we should add the check of `getInfo` before accessing the `info.title()` in the `getTitle` function. 
so first let `getInfo` be merged and then change `getTitle` function here.

packages/templates/clients/websocket/javascript/template/README.md.js
```diff
@@ -5,7 +5,7 @@ import { AvailableOperations } from '../components/AvailableOperations';
 export default function({ asyncapi, params }) {
   const server = getServer(asyncapi.servers(), params.server);
   const info = asyncapi.info();
-  const clientName = getClientName(info);
+  const clientName = getClientName(asyncapi);
```
Adi-204:
```suggestion
  const clientName = getClientName(asyncapi, params.appendClientSuffix, params.customClientName);
```

packages/helpers/src/index.js
```diff
@@ -1,6 +1,6 @@
 const { getMessageExamples, getOperationMessages } = require('./operations');
 const { getServerUrl, getServer }  = require('./servers');
-const { getClientName, listFiles, getInfo, toSnakeCase } = require('./utils');
+const { getClientName, listFiles, getInfo, toSnakeCase ,getTitle} = require('./utils');
```
derberg:
```suggestion
const { getClientName, listFiles, getInfo, toSnakeCase, getTitle} = require('./utils');
```

packages/helpers/test/utils.test.js
```diff
@@ -1,6 +1,6 @@
 const path = require('path');
 const { Parser, fromFile } = require('@asyncapi/parser');
-const { getClientName, getInfo, toSnakeCase } = require('@asyncapi/generator-helpers');
+const { getClientName, getInfo,getTitle, toSnakeCase } = require('@asyncapi/generator-helpers');
```
derberg:
```suggestion
const { getClientName, getInfo, getTitle, toSnakeCase } = require('@asyncapi/generator-helpers');
```

packages/templates/clients/websocket/javascript/template/README.md.js
```diff
@@ -5,7 +5,7 @@ import { AvailableOperations } from '../components/AvailableOperations';
 export default function({ asyncapi, params }) {
   const server = getServer(asyncapi.servers(), params.server);
   const info = asyncapi.info();
-  const clientName = getClientName(info);
+  const clientName = getClientName(asyncapi,params.appendClientSuffix, params.customClientName);
```
derberg:
```suggestion
  const clientName = getClientName(asyncapi, params.appendClientSuffix, params.customClientName);
```

derberg (commented):
few minors
I wonder why eslint is not picking this up

ItshMoh:
@derberg  done

PR #1569, opened 2025-05-21
test: add test for AvailableOperations component

### LINKED ISSUE #1562: add snapshot test for javascript `AvailableOperations`
In javascript client websocket template we have a component `AvailableOperations`  add snapshot testing for it - https://github.com/asyncapi/generator/blob/master/packages/templates/clients/websocket/javascript/components/AvailableOperations.js

KunalNasa:
Screenshot 2025-05-21 at 2 43 47 PM

* Add tests for AvailableOperations component. file: [AvailableOperations](https://github.com/asyncapi/generator/blob/master/packages/templates/clients/websocket/javascript/components/AvailableOperations.js)
* Updates component to handle the null and empty arrays as Operations object.

Resolves #1562

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved the AvailableOperations component to avoid rendering when no operations are available.
  
- **Tests**
  - Added tests to verify correct rendering of operations and ensure no output when the operations list is empty or null.

### DIFF
--- a/packages/templates/clients/websocket/javascript/components/AvailableOperations.js
+++ b/packages/templates/clients/websocket/javascript/components/AvailableOperations.js
@@ -3,6 +3,9 @@ import OperationHeader from './OperationHeader';
 import MessageExamples from './MessageExamples';
 
 export function AvailableOperations({ operations }) {
+  if (!operations || operations.length === 0) {
+    return null;
+  }
   return (
     <>
       <Text newLines={2}>### Available Operations</Text>
--- a/packages/templates/clients/websocket/javascript/template/README.md.js
+++ b/packages/templates/clients/websocket/javascript/template/README.md.js
@@ -46,7 +46,7 @@ Registers a callback to handle WebSocket errors.
 #### \`close()\`
 Closes the WebSocket connection.`}
       </Text>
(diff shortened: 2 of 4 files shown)


packages/templates/clients/websocket/javascript/components/AvailableOperations.js
```diff
@@ -3,6 +3,9 @@ import OperationHeader from './OperationHeader';
 import MessageExamples from './MessageExamples';
 
 export function AvailableOperations({ operations }) {
+  if (!operations || operations.length === 0) {
+    return null;
+  }
```
Adi-204, line 8: @KunalNasa because we have now return checks inside `AvailableOperations` component only. I think we can remove checks from `README.md.js` here - https://github.com/asyncapi/generator/blob/master/packages/templates/clients/websocket/javascript/template/README.md.js#L49
and make code of more simpler.

KunalNasa: Makes sense. Removed the redundant null check from the readme.md.js file.
Also, I had to force-push this time because of difference in git heads of the remote and local branches, as I had changed the title from the GitHub app and have no track in local.

PR #1572, opened 2025-05-21
refactor: replace hardcode `sendEchoMessage` method in `py` websocket client

### LINKED ISSUE #1517: refactor for python websocket client
Currently inside method - `async def send_message_static(message, socket):` the line `await HoppscotchEchoWebSocketClient._send(message, socket)` is written which means for all the generated client `HoppscotchEchoWebSocketClient` will be written but we need it to be dynamic and extract from asyncapi document. 

For ref -  https://github.com/asyncapi/generator/blob/20d10dc1014f686145c092542dad629fa6ac46c7/packages/templates/clients/websocket/python/components/SendEchoMessage.js#L31

Adi-204:
**Description**
Replace hardcode `sendEchoMessage` method in `py` websocket client.

image

**Related issue(s)**
Fixes #1517


packages/templates/clients/websocket/python/components/SendOperation.js
```diff
@@ -0,0 +1,48 @@
...
+        socket (websockets.WebSocketCommonProtocol): The WebSocket to send through.
+
+    Raises:
+        Exception: If sending fails or the socket is not connected.
+    """
+    await ${clientName}._send(message, socket)
+`}
+            </Text>
+          );
+        })
```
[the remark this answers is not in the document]
Adi-204: man this is wrong suggestion stop doing it.

Adi-204:
@derberg test for new component added in this PR - `SendOperation` is left but I thought maybe you review first current changes and suggest changes (if any) then I will add testcases.

packages/templates/clients/websocket/python/components/ClientClass.js
```diff
@@ -22,7 +23,8 @@ export function ClientClass({ clientName, serverUrl, title, queryParams }) {
       <RegisterOutgoingProcessor />
       <HandleMessage />
       <HandleError />
-      <SendEchoMessage />
+      <SendOperation sendOperations={sendOperations} clientName={clientName} />
```
derberg, line 26: should we call the component `SendOperation` influenced by the name of object it operates on, or rather `SendMethods` or `SendFunctions` or event `OutgoingFunctions`

packages/templates/clients/websocket/python/components/ClientClass.js
```diff
@@ -22,7 +23,8 @@ export function ClientClass({ clientName, serverUrl, title, queryParams }) {
...
+      {sendOperations.length > 0 && <Send />}
```
derberg, line 27: I know `Send` do not require any arguments, but wouldn't it be easier and cleaner to use if it actually accepts `sendOperations` as input and do `sendOperations.length > 0` internally ?

derberg (commented):
left few comments, also one small change is needed, in python, methods are called with `_` as words separation, not camel case

packages/templates/clients/websocket/python/components/ClientClass.js
Adi-204: So according to me if someone is developing template or want to modify our template code, I think they might refer to their own AsyncAPI document frequently in which word "operations" will be mentioned so to keep name like SendOperations I think it is more intuitive and one can easily understand.
But you have a lot of experience and talk with companies which are using AsyncAPI. wdyt about my point does it make sense? 🤔 interested to know your thoughts.

Adi-204:
> left few comments, also one small change is needed, in python, methods are called with `_` as words separation, not camel case

So I read about operations id from - https://www.asyncapi.com/docs/reference/specification/v3.0.0#operationsObject
the statement does not explicitly mention which casing (e.g., camelCase, PascalCase etc.) should be used for the operationId.
But I think majority users will use either camelCase or PascalCase. 
So I am adding a function which will convert either of them to snake_case but if someone is naming like sendmessage nothing will happen and same string will be returned. wdyt 🤔 ? should I proceed with function I mentioned?

packages/templates/clients/websocket/python/components/SendOperation.js
```diff
@@ -0,0 +1,56 @@
+import { Text } from '@asyncapi/generator-react-sdk';
+
+export function SendOperation({ sendOperations, clientName }) {
+  if (!sendOperations || sendOperations.length === 0) {
+    return null;
+  }
+
+  const toSnakeCase = (camelStr) => {
```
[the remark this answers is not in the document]
Adi-204: oh ok I will add a new helper

Adi-204:
@derberg I had to add 2 new operations in file for testing my toSnakeCase and hence integration test of some other part is also updated. I could change existing ones but I thought because multiple contributors are working on project I will break someone else testing code.

Issue #1538, opened 2025-05-06
add unit tests for `bindings.js#getQueryParams`


derberg:
like in the title

PR #1576, opened 2025-05-24
Added tests for bindings.js#getQueryParams

achaljhawar:
**Description**

- Added unit tests for the getQueryParams function in `packages/helpers/src/bindings.js`.
- Covered various scenarios, including missing WebSocket bindings, empty query parameters, and different channel states.

**Related issue(s)**
#1538

### DIFF
--- a/packages/helpers/test/bindings.test.js
+++ b/packages/helpers/test/bindings.test.js
@@ -1,0 +1,97 @@
+const { getQueryParams } = require('../src/bindings');
+
+class MockChannel {
+  constructor(bindings) {
+    this._bindings = bindings;
+  }
+
+  bindings() {
+    return {
+      has: (key) => key === 'ws' && !!this._bindings,
+      get: (key) => (key === 'ws' ? this._bindings : undefined),
+    };
+  }
+}
+
+class MockChannelsMap extends Map {
+  isEmpty() {
+    return this.size === 0;
+  }
+
+  all() {
+    return this;
+  }
+}
+
+describe('getQueryParams', () => {
+  it('should return null if there is no WebSocket binding', () => {
+    const channels = new MockChannelsMap();
+    channels.set('test/channel', new MockChannel(null));
+    expect(getQueryParams(channels)).toBeNull();
+  });
+
+  it('should return null if WebSocket binding exists but has no query parameters', () => {
+    const channels = new MockChannelsMap();
+    const wsBinding = { value: () => ({}) };
+    channels.set('test/channel', new MockChannel(wsBinding));
+    expect(getQueryParams(channels)).toBeNull();
+  });
+
+  it('should return null if WebSocket binding query exists but has no properties', () => {
+    const channels = new MockChannelsMap();
+    const wsBinding = { value: () => ({ query: {} }) };
+    channels.set('test/channel', new MockChannel(wsBinding));
+    expect(getQueryParams(channels)).toBeNull();
+  });
+
+  it('should extract query parameters with and without default values', () => {
+    const channels = new MockChannelsMap();
+    const wsBinding = {
+      value: () => ({
+        query: {
+          properties: {
+            foo: { default: 'bar' },
+            baz: {},
+            num: { default: 123 },
+          },
+        },
+      }),
+    };
+    channels.set('test/channel', new MockChannel(wsBinding));
+    const params = getQueryParams(channels);
+    expect(params).not.toBeNull();
+    expect(params.get('foo')).toBe('bar');
+    expect(params.get('baz')).toBe('');
+    expect(params.get('num')).toBe('123');
+  });
+
+  it('should return null for an empty channels map', () => {
+    const channels = new MockChannelsMap();
+    expect(getQueryParams(channels)).toBeNull();
+  });
+
+  it('should return null if channel bindings is undefined', () => {
+    const channels = new MockChannelsMap();
+    channels.set('test/channel', { bindings: () => undefined });
+    expect(getQueryParams(channels)).toBeNull();
+  });
+
+   it('should return null if channel itself is not as expected (e.g. missing bindings function)', () => {
+    const channels = new MockChannelsMap();
+    channels.set('test/channel', {});
+    expect(getQueryParams(channels)).toBeNull();
+  });
+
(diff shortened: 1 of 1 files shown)


achaljhawar:
@ItshMoh I have fixed the linting error

derberg (commented):
please refactor test and follow approach from other helper tests where we do not mock data inside test but reuse a test fixture like this https://github.com/asyncapi/generator/blob/master/packages/helpers/test/servers.test.js#L6

packages/helpers/test/bindings.test.js
```diff
@@ -0,0 +1,107 @@
...
+    const channels = parsedAsyncAPIDocument.channels();
+
+    const filteredChannels = new Map();
+    const channelWithoutBinding = channels.get('marketDataV1NoBinding');
+    if (channelWithoutBinding) {
+      filteredChannels.set('marketDataV1NoBinding', channelWithoutBinding);
+    }
+
+    filteredChannels.isEmpty = function() { return this.size === 0; };
+    filteredChannels.all = function() { return this; };
```
derberg, line 37: instead of doing this, and need for mocking

how about just using `delete` on the map? deleting channels that are not needed for given test case?

achaljhawar:
@derberg I have implemented the changes you asked me to make

packages/helpers/test/bindings.test.js
derberg: hey, I see you did some updates but you did not answer the question

achaljhawar: > instead of doing this, and need for mocking
> 
> how about just using `delete` on the map? deleting channels that are not needed for given test case?

Yeah, this can be done

derberg: awesome, than please proceed

achaljhawar: can you check my latest commit I used Map.delete() to implement the fix

achaljhawar: I have changed the variable names accordingly

packages/helpers/test/bindings.test.js
```diff
@@ -0,0 +1,89 @@
...
+  });
+
+  it('should return null if WebSocket binding query exists but has no properties', () => {
+    const channels = parsedAsyncAPIDocument.channels();
+    const filteredChannels = createChannelsWithOnly(['wsBindingEmptyQuery'], channels);
+    const params = getQueryParams(filteredChannels);
+    expect(params).toBeNull();
+  });
+
+  it('should return null if channel bindings is undefined', () => {
```
derberg: we should avoid mocking here too

packages/helpers/test/bindings.test.js
```diff
@@ -0,0 +1,89 @@
+const path = require('path');
+const { Parser, fromFile } = require('@asyncapi/parser');
+const { getQueryParams } = require('../src/bindings');
+
+const parser = new Parser();
+const asyncapi_v3_path = path.resolve(__dirname, './__fixtures__/asyncapi-websocket-query.yml');
+
+function createChannelsWithOnly(keysToKeep, originalChannels) {
```
derberg, line 7: why we need to create a new map here, and still do some mocking basically of `all` and `isEmpty`

derberg:
you pinged me for review again, but I see no changes, and no answers to questions

achaljhawar:
@derberg sorry I pinged you before , I made some changes that remove mocking and refactored the `createChannelsWithOnly` function to return a `Map` object directly, instead of an object that mocks the `isEmpty` and `all` methods.

packages/helpers/test/bindings.test.js
derberg: hey are you still working on something?

achaljhawar: No

achaljhawar: I am done ig

derberg: > why we need to create a new map here, and still do some mocking basically of all and isEmpty

achaljhawar: ohk I'll fix this

achaljhawar: should I create a second fixture file to fix this problem?
or should I test with the full channel object and verify its behaviour?

derberg: let's not decide yet, first answer my question 😄 why new map is needed, why you cannot operate in the helper on the map you get as input

achaljhawar: The new Map is needed to create a copy so the original channels map isn't mutated, which would affect other tests that need the data.

packages/helpers/test/bindings.test.js
```diff
@@ -0,0 +1,85 @@
+const path = require('path');
+const { Parser, fromFile } = require('@asyncapi/parser');
+const { getQueryParams } = require('../src/bindings');
```
derberg:
```suggestion
const { getQueryParams } = require('@asyncapi/generator-helpers');
```

packages/helpers/test/bindings.test.js
derberg: oh, now much better. Remember reviewer did not code and do not see what issues you had during development. Sometimes we just ask question not to play smart, but to really get an answer 😄 thanks for the context - now it makes sense why you've done so, and why later you still do mocking of `all` and `isEmpty` - please add comment to tested code to explain why this mocking is needed

packages/helpers/test/bindings.test.js
```diff
@@ -0,0 +1,85 @@
...
+
+    expect(params).not.toBeNull();
+    expect(params.get('heartbeat')).toBe('false');
+    expect(params.get('top_of_book')).toBe('false');
+    expect(params.get('bids')).toBe('true');
+    expect(params.get('offers')).toBe('');
+  });
+
+  it('should return null for channel without WebSocket binding', () => {
+    const channels = parsedAsyncAPIDocument.channels();
```
derberg: I don't think it must be repeasted with each test case

achaljhawar:
@derberg I have made some changes according to your suggestions

Issue #1514, opened 2025-04-24
change coderabbit's config to be less annoying


nightknighto:
### Why do we need this improvement?

CodeRabbit runs on every PR and writes comments that summarizes and explains the changes made by the PR. It is generally useful, but it has several pain points that make it a bit annoying in many cases, such as being too verbose or too repetitive, having dummy sections, responding to other bots, etc.

Points that can be improved:
- Removing Poem section from the initial comment.
- Removing suggested reviewers. Redundant because reviewers are assigned automatically.
- Remove suggested labels. Redundant because does it ever suggest anything other than "ready-to-merge"?
- Preventing replying to other bots, especially @asyncapi-bot 
- Making sequence diagram optional.
Usually, we make small frequent PRs. These PRs are usually simple and direct enough. The amount of sections CodeRabbit produces for these small PRs is overkill, being too long and repeating the same information. These sections are: Summary (in the PR original message), Walkthrough, Changes, and Sequence Diagram. Sequence diagram is too long, and often not that useful unless for a major PR. Would be great to disable by default and optionally create it via a command or something.

This greatly involves people's preferences, so it will be great to get thoughts about each of these points by the active contributors.

### How will this change help?

This will decrease the size of the messages CodeRabbit adds to each PR, and thus it will be easier to read the discussion, less information to digest, fewer distractions, and more direct feedback.

### Screenshots

Bots talking to each other:

![Image](https://github.com/user-attachments/assets/a40ccf84-f235-47ad-aa22-3320fb5d835e)

Huge message added to each PR:

![Image](https://github.com/user-attachments/assets/8c18bcfa-cc1a-4c88-ae35-87e34bf0827f)

![Image](https://github.com/user-attachments/assets/332382b8-8dd8-49f7-993f-a5f7772cc894)

### How could it be implemented/designed?

- Changing the configuration of CodeRabbit to reflect the desired changes.
- Exploring whether the sequence diagram can be created on demand with a command inside a PR.

### 🚧 Breaking changes

No

### 👀 Have you checked for similar open issues?

- [x] I checked and didn't find a similar issue

### 🏢 Have you read the Contributing Guidelines?

- [x] I have read the [Contributing Guidelines](https://github.com/asyncapi/.github/blob/master/CONTRIBUTING.md)

### Are you willing to work on this issue?

None

derberg:
> bots talking to each other

[...]

PR #1577, opened 2025-05-25
ci: add coderabbit configuration file

Siddhant-Pragyan-Sinha:
**Description**
Here is a concise summary of what was agreed upon to improve CodeRabbit, preserving the original meaning:

* **Remove the poem section** from CodeRabbit’s initial comment, as it’s not useful.
* **Remove suggested reviewers and labels** since these are redundant—reviewers are assigned automatically, and label suggestions are typically not helpful.
* **Make the sequence diagram optional**, as it is often too verbose and not always useful. Ideally, it should be generated on demand via a command.
* **Retain the summary section**, as it is helpful when PR descriptions are lacking.

**Related issue(s)**
Related to #1514 

## Summary by CodeRabbit

- **Chores**
  - Introduced a new configuration file to customize CodeRabbit system settings, including language preferences, review features, summary generation, and chat behavior.
  - Enabled automatic summarization of changed files, linked issues, related issues, and pull requests.
  - Activated auto review with incremental support and status tracking.
  - Set knowledge base to automatically scope learnings, issues, and pull requests.

### DIFF
--- a/.coderabbit.yaml
+++ b/.coderabbit.yaml
@@ -2,35 +2,25 @@
 
 language: en-US
 reviews:
-  # Disable poem generation in walkthrough comments
   poem: false
   suggested_reviewers: false
   suggested_labels: false
   sequence_diagrams: false
-
-  # Keep high level summary (helpful when PR descriptions are lacking)
   high_level_summary: true
   high_level_summary_placeholder: "@coderabbitai summary"
-
-  # Standard review settings
   profile: chill
   auto_review:
     enabled: true
     auto_incremental_review: true
-
-  # Additional useful settings
   review_status: true
   commit_status: true
   changed_files_summary: true
   assess_linked_issues: true
   related_issues: true
   related_prs: true
-
-# Chat configuration
 chat:
   auto_reply: true
 
-# Knowledge base configuration
 knowledge_base:
   learnings:
     scope: auto
--- a/.coderabbit.yaml
+++ b/.coderabbit.yaml
@@ -19,7 +19,7 @@ reviews:
   related_issues: true
   related_prs: true
 chat:
-  auto_reply: true
+  auto_reply: false
 
 knowledge_base:
   learnings:
--- a/.coderabbit.yaml
+++ b/.coderabbit.yaml
@@ -1,0 +1,40 @@
+# CodeRabbit Configuration File
+
+language: en-US
+reviews:
+  # Disable poem generation in walkthrough comments
+  poem: false
+  suggested_reviewers: false
+  suggested_labels: false
+  sequence_diagrams: false
+
+  # Keep high level summary (helpful when PR descriptions are lacking)
+  high_level_summary: true
+  high_level_summary_placeholder: "@coderabbitai summary"
+
+  # Standard review settings
+  profile: chill
+  auto_review:
+    enabled: true
+    auto_incremental_review: true
+
+  # Additional useful settings
+  review_status: true
+  commit_status: true
+  changed_files_summary: true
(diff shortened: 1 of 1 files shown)


.coderabbit.yaml
```diff
@@ -0,0 +1,30 @@
...
+    enabled: true
+    auto_incremental_review: true
+  review_status: true
+  commit_status: true
+  changed_files_summary: true
+  assess_linked_issues: true
+  related_issues: true
+  related_prs: true
+chat:
+  auto_reply: true
```
Adi-204, line 22: hey @Siddhant-Pragyan-Sinha can you please explain why you set `auto_reply: true` because according to docs true is default value only so we don't need to write it. But I think maybe we can keep it false so that when user tag coderabbit then only it will reply in chat. wdyt 🤔? 
But don't change anything for now we can discuss it first.

Siddhant-Pragyan-Sinha:
Hey @derberg @Adi-204 , please suggest me improvements and stuff that we can change or add to the configuration.
Btw, looking over the website repo [.coderabbit.yaml](https://github.com/asyncapi/website/blob/a63abed1c135003a75a589ae860aa106cddaf73c/.coderabbit.yaml#L4) file, can we add these tools as well:-
`tools:
    markdownlint:
      enabled: true
    biome:
      enabled: true
    github-checks:
      enabled: true
      timeout_ms: 180000
    languagetool:
      enabled: true
      enabled_only: false
      level: default
    hadolint:
      enabled: true
    yamllint:
      enabled: true
    gitleaks:
      enabled: true
    eslint:
      enabled: true
    actionlint:
      enabled: true
    semgrep:
      enabled: true`

.coderabbit.yaml
Siddhant-Pragyan-Sinha: I actually started a discussion in the slack to get a workaround that so that all except asyncapi-bot gets responded to.

Adi-204: > Although I get what you mean to say i.e. to prevent bots talking to each other (if I am not wrong). Sure, we can keep it like that, lets see what @derberg has to say.


No, so actually if we keep `auto_reply: false` it won't solve bot issue because `asyncapi-bot` actually tag coderabbit in comment so coderabbit will still reply. It was suggestion which we didn't discuss in issue. I guess we should better discuss it in issue can contributors can provide inputs thier rather than slack channel.

Adi-204:
> Hey @derberg @Adi-204 , please suggest me improvements and stuff that we can change or add to the configuration.
> 
> Btw, looking over the website repo [.coderabbit.yaml](https://github.com/asyncapi/website/blob/a63abed1c135003a75a589ae860aa106cddaf73c/.coderabbit.yaml#L4) file, can we add these tools as well:- `tools: markdownlint: enabled: true biome: enabled: true github-checks: enabled: true timeout_ms: 180000 languagetool: enabled: true enabled_only: false level: default hadolint: enabled: true yamllint: enabled: true gitleaks: enabled: true eslint: enabled: true actionlint: enabled: true semgrep: enabled: true`


@Siddhant-Pragyan-Sinha for now I don't think we need to add this stuff rather we can update the file in future if we face any problems in PR by coderabbit.

.coderabbit.yaml
derberg: yeah, I guess we can set it to false

I don't see a problem that we will have to explicitly tag rabbit to provide answer, as it is normal flow with humans, you also need to tag them so they are aware you're talking to them

Siddhant-Pragyan-Sinha: Sure.

PR #1583, opened 2025-05-29
chore: add fields in acceptance testing package.json

Adi-204:
**Description**
Added important fields in package.json to solve the releases error.

**Related issue(s)**
As per discuss in slack - https://asyncapi.slack.com/archives/C072JMTJ85Q/p1748461109464129

### DIFF
--- a/packages/templates/clients/websocket/test/javascript/package.json
+++ b/packages/templates/clients/websocket/test/javascript/package.json
@@ -1,4 +1,7 @@
 {
+    "name": "@asyncapi/template-acceptance-test-js",
+    "version": "0.0.1",
+    "description": "This is an acceptance test for the JavaScript websocket client template",
     "scripts": {
         "test": "jest --coverage",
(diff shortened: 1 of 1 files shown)


packages/templates/clients/websocket/test/javascript/acceptance.test.js
```diff
@@ -2,7 +2,7 @@
  * @jest-environment node
  */
 
-const WSClient = require('../../javascript/test/temp/snapshotTestResult/client-hoppscotch');
+const WSClient = require('../../javascript/test/temp/snapshotTestResult/client-hoppscotch.js');
```
derberg, line 5:
```suggestion
const WSClient = require('../../javascript/test/temp/snapshotTestResult/client-hoppscotch');
```

derberg:
@Adi-204 sorry for directly committing to your branch but we need to fix it fast. The solution was to unplug test run from turbo, so renaming to something less obvious, which in this case makes sense to call these test scritp `test:acceptance` - this way turbo won't pick it up

Adi-204:
yeah @derberg your solution make complete sense and great learning about Turbo-repo thanks : )

PR #1603, opened 2025-06-04
chore: unify `test:update` script

### LINKED ISSUE #1540: unify `npm run test:update` across the whole project
unify `npm run test:update` across the whole project. Basically make sure that whenever we have `jest --updateSnapshot` in `package.json`, the script is called `test:update` and instead of calling jest directly, it calls regular test script like so: `npm run test -- -u`. We also need `test:update` in main `package.json`

thulieblack:
unify npm run test:update across the whole project

Resolves #1540

## Summary by CodeRabbit

- **Chores**
	- Updated and standardized snapshot update scripts across multiple packages to ensure consistency when updating test snapshots.
	- Added missing trailing comma in the root package configuration for improved syntax validity.
	- Introduced new snapshot update scripts in select packages for easier test maintenance.


derberg (commented):
please look into requirements again

what about:
- `apps/generator/package.json` where `test:integration` should be called
- `packages/templates/clients/websocket/dart/package.json`
- also requirement `We also need test:update in main package.json`

packages/templates/clients/websocket/test/javascript/package.json
```diff
@@ -4,6 +4,7 @@
     "description": "This is an acceptance test for the JavaScript websocket client template",
     "scripts": {
         "test:acceptance": "jest --coverage",
+        "test:update": "npm run test -- -u",
```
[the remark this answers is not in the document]
derberg: valid point

package.json
```diff
@@ -31,7 +31,8 @@
     "helpers:lint": "turbo run lint --filter=@asyncapi/generator-helpers",
     "templates:test": "turbo run test --filter=@asyncapi/template*",
     "packages:test": "turbo run test --filter=./packages/** --only",
-    "hooks:test": "turbo run test --filter=@asyncapi/generator-hooks"
+    "hooks:test": "turbo run test --filter=@asyncapi/generator-hooks",
+    "test:update": "npm run test -- -u"
```
Adi-204: @thulieblack I don't think this `test:update` should be working. We are using turbo so it should be `turbo run test -- -u`
But before applying my suggestion please do check running `npm run test:update` globally, it should throw error. 
thanks :)

thulieblack: you are correct Adi, thanks

packages/helpers/package.json
```diff
@@ -4,6 +4,7 @@
   "description": "Package with reusable helpers that make it easier to work with AsyncAPI structures.",
   "scripts": {
     "test": "jest --coverage",
+    "test:update": "npm run test -- -u",
```
Adi-204: @thulieblack we are not performing snapshot testing in helpers for now, so I don't think their is need to add `test:update` script.

thulieblack: But its part of the packages.json unless I understood the issue requirements?

Adi-204: yes but in issue it is mention whenever we have `jest --updateSnapshot` in `package.json` we had to replace it with `npm run test -- -u` for helpers we are not doing "Snapshot testing" so we don't require `test:update` script. 
Does it make sense?

packages/templates/clients/websocket/test/javascript/package.json
```diff
@@ -4,6 +4,7 @@
     "description": "This is an acceptance test for the JavaScript websocket client template",
     "scripts": {
         "test:acceptance": "jest --coverage",
+        "test:update": "npm run test:acceptance -- -u",
```
derberg: can be removed, no snapshots here, just like in case of `hooks`

packages/helpers/package.json
```diff
@@ -4,6 +4,7 @@
   "description": "Package with reusable helpers that make it easier to work with AsyncAPI structures.",
   "scripts": {
     "test": "jest --coverage",
+    "test:update": "jest --updateSnapshot",
```
Adi-204: @thulieblack sorry if I confuse you with my comment [here](https://github.com/asyncapi/generator/pull/1603#discussion_r2131671792) but what I meant was that it can be 
removed no snapshots here,  just like in case of `hooks`

PR #1618, opened 2025-06-18
chore: add new package `@asyncapi/keeper`

### LINKED ISSUE #1565: explore how we can perform validation before sending messages
currently in javascript websocket client we have `SendOperation` component we dynamically extract all send type operations from AsyncAPI document and directly send message. But we can take this to another level by performing validation of message being send with help of libraries like ajv. So the client user do not have to learn how to validate messages before sending them.

Ref - https://github.com/asyncapi/generator/blob/master/packages/templates/clients/websocket/javascript/components/SendOperation.js

Adi-204:
**Description**
This is really basic setup for adding new package under `generator/app` which internally uses @hyperjump/json-schema. We are introducing it to abstract complexity of @hyperjump/json-schema and also it will be reusable across all clients message-validation so we don't need to write same code again and again. 

**Related issue(s)**
Fixes #1565 

## Summary by CodeRabbit

* **New Features**
  * Introduced the `@asyncapi/keeper` package for validating message payloads against JSON Schema Draft-07.
  * Added asynchronous validation functions, including validation by operation ID from an AsyncAPI document.
  * Provided utility functions for parsing AsyncAPI documents and generating schema URIs.

* **Documentation**
  * Added a comprehensive README with installation instructions, API documentation, and usage examples.

* **Tests**
  * Added integration tests to verify message validation functionality.
  * Included an AsyncAPI specification fixture for testing message validation.

* **Chores**
  * Added configuration files and scripts for building, testing, and linting the new package.
  * Updated `.gitignore` to exclude build artifacts.
  * Added Jest setup for structuredClone polyfill to support testing environments.


Adi-204:
The current approach of using babel is most probably not going to work according to me because babel is compiling and converting the code for `@hyperjump/json-schema/draft-07` but it's internally implementation is written in pure ESM and this library also doesn't provide support for CommonJS like `@asyncapi/generator-react-sdk` and `@asyncapi/modelina`.

Tests are also not written for now because Jest is not working with `@hyperjump/json-schema/draft-07` as if you run `npm run test` it will throw error directly.

Adi-204:
@derberg I tried to do something similar to `@asyncapi/modelina` and now it is throwing below error. 

image

I think main problem is  - 
1. Jest Assumes code in node_modules is CommonJS. 
2. Skips transforming ESM code from dependencies like `@hyperjump/json-schema`

So we are transforming code inside index.js successfully but code present in node_modules of @hyperjump/json-schema is still in pure ESM hence it will throw error. wdyt 🤔? Does it make sense? 
I have also dropped message in JSON schema slack channel about some kind of workaround to solve this without going for "type": "module"

Current I am trying to add `transformIgnorePatterns` but it is also not working for now.

Adi-204:
I am currently trying to add 
```js
"transformIgnorePatterns": [
      "/node_modules/(?!@hyperjump/json-schema|json-pointer|pact)"
    ]
```
in package.json but it is also failing not sure why. People are having similar issues with other libraries also - https://github.com/jestjs/jest/issues/8272 not sure now whether it is problem with jest or hyperjump.

Adi-204:
simplify configurations a bit. I am currently using jest-esm-transformer instead of babel-jest. Now it is throwing error - 

image

Now it is no more throwing error of unable to identify import statement. But now it is throwing error of unable to find `@hyperjump/browser/jref` module which is strange because I installed latest version of library, so all the other dependencies must be included. I am not using `@hyperjump/browser/jref` explicitly in my code and it is used in library [here](https://github.com/search?q=repo%3Ahyperjump-io%2Fjson-schema%20%40hyperjump%2Fbrowser%2Fjref&type=code). I can also see `@hyperjump/browser` is peerDependencies when you installed `@hyperjump/json-schema`. 

@derberg can you please look into it. This is strange error because dependency is already installed. Not sure why this error.

Adi-204:
1. `app/message-validation/` node_modules

Screenshot 2025-06-24 095919

2. Global node_modules

Screenshot 2025-06-24 095949

I think this is classic problem of dependency version conflict in Turborepo. I tried to use `moduleNameMapper` to reference `@hyperjump/browser/jref` but it is also not working.

derberg:
@Adi-204 the problem is that `@hyperjump/browser` is not declared inside `@hyperjump/json-schema` as normal dependency but as a `peerDependency` which means you need to explicitly install it on your own. I guess the best version to install would be the one defined in `peerDependency`: https://github.com/hyperjump-io/json-schema/blob/main/package.json#L77

btw, didn't you think about some more fancy name for the library? `payload-police` or `message-guard` or `events-shield` 😄 `message-validation` is super boring 😆

Adi-204:
image


@derberg I tried this solution before also but when I am trying to install normally like npm i @hyperjump/browser but it is not adding explicitly to node_modules of message-validation and the error is still same. Do I need to install it differently 😕

package-lock.json
```diff
@@ -2433,6 +2495,37 @@
       "integrity": "sha512-k2Ty1JcVojjJFwrg/ThKi2ujJ7XNLYaFGNB/bWT9wGR+oSMJHMa5w+CUq6p/pVrKeNNgA7pCqEcjSnHVoqJQFw==",
       "license": "MIT"
     },
+    "node_modules/@hyperjump/browser": {
```
Adi-204, line 4699: @derberg it is added here when I installed `@hyperjump/json-schema`

Adi-204: The only thing I wonder is when it is mention as `peerDependencies` with version 1.1.0. How the latest version is installed?

derberg: It's because of `^` symbol, it's ok.

derberg:
ok, the problem is jest related.

so explicit installation of browser package solves only one part of the issue

another issue is that it is installed in root node_modules, while json-schema package in message-validation. Looks like Jest do not understand `export` defined in `package.json` of browser lib, when package is not installed in the same node_modules as json-schema

```
"exports": {
    ".": "./lib/index.js",
    "./jref": "./lib/jref/index.js"
  }
```

just add `"^@hyperjump/browser/jref$": "<rootDir>/../../node_modules/@hyperjump/browser/lib/jref/index.js"` to `moduleNameMapper` in the `apps/message-validation/package.json`

Adi-204:
@derberg I also tried that approach as mention in [comment](https://github.com/asyncapi/generator/pull/1618#issuecomment-2998822617) it is also throwing internal error of the library.

image


and actually `structuredClone` is only mentioned in the file it is not even import but because it is internal implementation so it might be handle differently I searched in [@hyperjump/json-schema](https://github.com/hyperjump-io/json-schema) codebase also https://github.com/search?q=repo%3Ahyperjump-io%2Fjson-schema%20structuredClone&type=code it is use only once. Should I again contact to Jason Desrosiers? wdyt 🤔 ?

derberg:
well the error is valid, `structuredClone` is rally not defined: https://github.com/hyperjump-io/json-schema/blob/main/lib/schema.js#L42, reach out to him, try slack or just create GH issue

Adi-204:
@derberg I looked into latest error throw found a discussion on stackoverflow - https://stackoverflow.com/questions/73607410/referenceerror-structuredclone-is-not-defined-using-jest-with-nodejs-typesc

Many solutions are written to solve the error we are getting.

1. I tried adding `global.structuredClone` mention on stackoverflow. This is working but it doesn't seem to be great idea to me.

image

2. I tried simple solution adding `"testEnvironment": "node",` because it is mention in official docs also that structuredClone should work with Node > 17 but I think jest@27.3.1 is not supporting this fully. so this really simple and nice approach failed. I tried to upgrade jest but it was throwing different error which I will try to solve later.

So, we do have a working solution but it is not best hence I am exploring more to find best and simple solution.

Adi-204:
@derberg as I mention in last comment, we had 2 solution - 

1. Use https://www.npmjs.com/package/@ungap/structured-clone and we can remove the error of `structuredClone` as I have done in this PR. We are using `@ungap/structured-clone` only for testing and jest config, so I think it should not be a problem. But not sure, I think you insights are needed.

2. Upgrade Jest version to 29 and it was written at some places that in v.29 wont throw error of `structuredClone`. I tried this but only problem is if we upgrade to v29 it won't support `jest-esm-transformer` which is currently able to transform code. So we would have to go for `babel-jest` I tried again but it is unable to transform node_modules not sure why?

So, wdyt 🤔 ?

Adi-204:
@derberg if you didn;t like the package name we can change it maybe payload-police 😆

derberg:
how about `lagertha` as the name of the shield-maiden from Vikings tv series? You know, shield, event/message shield :) you know :D or `valkyrie` that were responsible for `choose which warriors would die, and escort the worthy dead to Valhalla` - you know, like validating against the schema who can pass the post :D 

now depends which movies you are a fan off, Vikings or Avengers-Thor

derberg:
so @Adi-204 ? `valkyrie` or basically `@asyncapi/valkyrie`? 😃

Adi-204:
> so @Adi-204 ? `valkyrie` or basically `@asyncapi/valkyrie`? 😃

@derberg actually to be honest I'm not big fan of any type of movies 😅

derberg:
any type of what I mentioned or in general? `valkyrie` is more  related to nordic old school mythology, like greek or roman mythology that is very popular in mainstream. Good name that resonates well with this package scope. If not movies no than what is your thing? some specific gaming, or just code 🥰 😃

Adi-204:
> any type of what I mentioned or in general? `valkyrie` is more related to nordic old school mythology, like greek or roman mythology that is very popular in mainstream. Good name that resonates well with this package scope. If not movies no than what is your thing? some specific gaming, or just code 🥰 😃

@derberg in general no movies. majority is code 🚀 but outside it any sports like chess, football and many more.

derberg:
then `@asyncapi/goalie`? goalie as short for goalkeeper, wdyt?

derberg:
ok, it is either `@asyncapi/payload-police` or `@asyncapi/goalie` or `@asyncapi/keeper` or `@asyncapi/goalkeeper`

Pick one and rename it accordingly

Adi-204:
@derberg package name updated.

apps/keeper/README.md
```diff
@@ -0,0 +1,56 @@
...
+
+#### Parameters
+
+| Parameter | Type               | Description                                                                 | Required |
+|-----------|--------------------|-----------------------------------------------------------------------------|----------|
+| `message` | `any`           | The message payload to validate (any non-null value)                                      | Yes      |
+| `schemas` | `Array<Object>`    | Array of JSON Schema Draft-07 objects representing valid message structures  | Yes      |
+
+#### Returns
+`Promise<boolean>` — Resolves to true if the message is valid against any schema, otherwise false.
```
derberg: what about technical info, in case of `false` that indicates what made validation fail?

apps/keeper/test/index.test.js
```diff
@@ -0,0 +1,86 @@
...
+  let schemas;
+
+  beforeAll(async () => {
+    const parseResult = await fromFile(parser, asyncapi_v3_path).parse();
+    parsedAsyncAPIDocument = parseResult.document;
+    schemas = parsedAsyncAPIDocument.schemas().all();
+  });
+
+  test('should validate a correct message against schemas and return true', async () => {
+    const jsonSchemas = schemas.map(schema => schema.json());
```
derberg: this is my only concern

user needs to know parser, and know how to extract all schemas, and only schemas that should given message be validated agains, not all

shouldn't api of this library accept 'asyncapi' document as input, and we internally use parser?

did you look at message-validation package and it's API, and how it is used in https://github.com/asyncapi/nodejs-template?

Adi-204: @derberg I thought in templates we will write extract logic inside file only 
https://github.com/asyncapi/nodejs-template/blob/master/template/src/lib/message-validator.js 
But yeah I also like how they are using in above file ok, I will do some refactor.

Adi-204: @derberg so right now main purpose of building this is before sending message in a operation we need to validate message. so should we make something like `validateOperationPayload(asyncapiDoc, operationId, payload)` 
wdyt 🤔 ?

Adi-204: you can suggest name of function which is more intuitive and users are not confuse.

derberg: but `validateMessage` is completely fine, a good name, nothing against it. User just wants to validate a message.

maybe try now (without real integration really) to see how would you use this function in existing JS client we generate - so you see how it fits, and what data you have access in component when you use the comonent in given context - maybe I'm wrong and you do not need to support `asyncapi` as input. My opinion is abstract here and affected by the fact that keeper is now standalone library, so I focus on the library and not our main use case. So yeah, check now on template level

Adi-204: @derberg I understand your point and if we take `asyncapiPath` file as input we can also avoid `schema.js` https://github.com/asyncapi/generator/pull/1615/files#diff-38eae824ef67eb8caab5faeb2b1f423acd9a693ab3d03da8d45374e4db467a9c what I did in previous PR.  It was file only having schemas with operation. 

The only thing I don't like much is in `nodejs-template` here - https://github.com/asyncapi/nodejs-template/blob/master/template/src/lib/message-validator.js#L12 
They are copying input `asyncapiFile` and created `asyncapi.yml` in generated output. We would also have to do something similar.
Other than that approach use in `nodejs-template` is nice.

new solution can be `validateMessage(asyncapiFilePath, message, operationId)` now complete logic will be moved to our library `keeper` and folks can even use this as standalone library. 
wdyt 🤔 ?

derberg: > The only thing I don't like much is in nodejs-template here - https://github.com/asyncapi/nodejs-template/blob/master/template/src/lib/message-validator.js#L12 They are copying input asyncapiFile and created asyncapi.yml in generated output.

why you do not like it?

> now complete logic will be moved to our library

what logic you have in mind

Adi-204: > what logic you have in mind

if we proceed with something similar in Nodejs-template. Than in `@asyncapi/keeper` we can have `.validateMessage()` which it already has but also `.validateByOperationId()` and many more. 
When it comes to coding logic than if we have input asyncapi file path than inside library we can use `@asyncapi/parser` and get asyncapi doc object 
`asyncapi doc object` -> `operations.get(operationId)` -> `getOperationMessages(operation)` -> `messagePayloads` -> `schemas`

I hope above is not confusing please do let me know if something seems to be wrong.

apps/keeper/src/index.js
```diff
@@ -0,0 +1,96 @@
...
+    asyncapi = parseResult.document;
+  } catch (error) {
+    throw new Error(`Failed to parse AsyncAPI document: ${error.message}`);
+  }
+
+  const operation = asyncapi.operations().get(operationId);
+  const messages = operation.messages().all();
+  const messagePayloads = messages.map((message) => message.payload());
+  const schemas = messagePayloads.map(payload => payload.json());
+
```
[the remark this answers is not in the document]
Adi-204: @derberg this is interesting. I was thinking we can add new helpers in `@asyncapi/helper` and we will use them here. So that our `@asyncapi/helper` also provide wider support will create follow-ups for that.

Adi-204:
@derberg how you think of modified version. we can easily add many more validateBy in future if needed.

apps/keeper/README.md
Adi-204: @derberg but currently in implementation of validateMessage if any of schema valid true against payload we return true. As in our usecase - If an AsyncAPI “send” operation declares multiple possible message payloads, you should consider the payload valid if it matches any one of those schemas. Please correct me if I'm wrong. 
So if we return false it means payload failed against all the schemas so should we than throw technical info for all?

Adi-204: if we want to throw technical info also it has to be done by `@hyperjump/json-schema/draft-07` only. I read docs it is mention - Only the FLAG output format is part of the Stable API. Additional [output formats](https://github.com/hyperjump-io/json-schema#output-formats) are included as part of the Experimental API. So we can give tech info also but than we will have to use "Experimental API".

apps/keeper/src/index.js
```diff
@@ -0,0 +1,93 @@
...
+  try {
+    const parseResult = await fromFile(parser, asyncapiFilepath).parse();
+    asyncapi = parseResult.document;
+  } catch (error) {
+    throw new Error(`Failed to parse AsyncAPI document: ${error.message}`);
+  }
+
+  const operation = asyncapi.operations().get(operationId);
+  const messages = operation.messages().all();
+  const messagePayloads = messages.map((message) => message.payload());
```
derberg: only if `hasPayload()` I guess?

apps/keeper/package.json
```diff
@@ -0,0 +1,56 @@
+{
+  "name": "@asyncapi/keeper",
+  "version": "1.0.0",
```
derberg: 0.0.1 please, we do not release it yet

apps/keeper/README.md
derberg: yes, we should use info from `@hyperjump/json-schema`

yeah, I don't see issues with fact we will use `Experimental API`, let's test it

apps/keeper/package.json
```diff
@@ -0,0 +1,56 @@
...
+  "description": "AsyncAPI message payload validation library that validates messages against JSON Schema (Draft-07)",
+  "scripts": {
+    "build": "babel src --out-dir lib",
+    "test": "npm run build && jest --coverage",
```
derberg: please also add `"pretest": "npm run build"`

apps/keeper/src/index.js
```diff
@@ -0,0 +1,93 @@
...
+export async function validateMessage(message, schemas) {
+  if (!Array.isArray(schemas)) {
+    throw new Error(`Invalid "schemas" parameter: expected an array of JSON Schemas, but got ${typeof schemas}`);
+  }
+  if (!schemas || schemas.length === 0) {
+    throw new Error('Empty "schemas" array: at least one JSON Schema must be provided for validation.');
+  }
+  if (message === null || message === undefined) {
+    throw new Error(`Invalid "message" parameter: expected a non-null object to validate, but received ${message}`);
+  }
```
derberg: we should not throw errors, 

AsyncAPI documents may be tiny, or maybe super detailed - we do not dictate that

so there can be AsyncAPI documents that have no schemas for example, which is valid. We should not throw errors there, basically log why we skip message validation. Except for the message, this is always passed, so throwing error makes sense

Adi-204:
@derberg I implemented error tech info after so many challenges and then when finally it was implemented I saw the error message logged not useful because you can see - https://github.com/hyperjump-io/json-schema/issues/56#issuecomment-2000539074 
so we won't able to give tech error info to users for now at least 😞. I have not pushed that code because error message are not useful you can see the message share by users in the same issue I shared above. not sure what to do know I have not updated `README.md` yet just waiting for your confirmation.
Error example - 
```
"errors":[{
"keyword":"https://json-schema.org/evaluation/validate",
"absoluteKeywordLocation":"http://example.com/schemas/string#/items",
"instanceLocation":"#/0",
"valid":false
}]
```

apps/keeper/test/index.test.js
```diff
@@ -0,0 +1,93 @@
...
+
+    beforeAll(async () => {
+      compiledSchemas = await compileSchemasByOperationId(asyncapi_v3_path, 'sendHelloMessage');
+    });
+
+    test('should validate a correct message against operation schemas and return true', async () => {
+      const validMessage = {
+        content: 'This is a operation test message'
+      };
+      const result = validateMessage(compiledSchemas, validMessage);
```
derberg, line 81: test q

derberg:
don't worry @Adi-204 
I'm aware that in ecosystem there is no widely adopted standard for "human readable" errors. So the current error is completely fine. For initial implementation completely technical error log is enough.

there are projects like https://www.npmjs.com/package/@segment/ajv-human-errors or https://www.npmjs.com/package/better-ajv-errors but no need to use them anyway

Issue #1564, opened 2025-05-19
Refactoring of props for ClientClass in python template


Adi-204:
currently inside `client.py.js` of python websocket client we have a component ClientClass which passes all the props down to respective components. This list can grow in future and hard to manage, their is potential of refactoring this code and passing complete `asyncapi`  object and extracting from it in their respective components. 

Ref - https://github.com/asyncapi/generator/blob/master/packages/templates/clients/websocket/python/template/client.py.js

Adi-204:
anyone can take up the issue.

achaljhawar:
ohk 👍🏼

Adi-204:
@achaljhawar are you working on the issue? Please mention it explicitly.

achaljhawar:
Yes, I did not specifically mention it because I am still figuring out how to work on it.

> [@achaljhawar](https://github.com/achaljhawar) are you working on the issue? Please mention it explicitly.

Adi-204:
@achaljhawar no it is just simple refactoring what I mean is in ClientClass directly pass asyncapi obj and call helpers in ClientClass for now. Start with simpler PR will look into it. just code refactoring no need to add new code.

PR #1628, opened 2025-06-25
refactor: pass asyncapi object in ClientClass

achaljhawar:
fixes: #1564 

Refactored the Python Websocket client template to pass the entire `asyncapi` object and `params` to the `ClientClass` component instead of individual props.

## Summary by CodeRabbit

- **Refactor**
  - Improved client configuration by consolidating multiple settings into a single object for easier management.
  - Simplified internal code structure without affecting user-facing features.

No visible changes to functionality or user experience.


achaljhawar:
@Adi-204 does this work?

packages/templates/clients/websocket/python/template/client.py.js
```diff
@@ -1,26 +1,21 @@
...
 export default function ({ asyncapi, params }) {
   const server = getServer(asyncapi.servers(), params.server);
   const info = getInfo(asyncapi);
-  const title = getTitle(asyncapi);
   const queryParams = getQueryParams(asyncapi.channels());
-  const clientName = getClientName(asyncapi, params.appendClientSuffix, params.customClientName);
-  const serverUrl = getServerUrl(server);
-  const operations = asyncapi.operations();
   return (
-    // The clientFileName default values can be found and modified under the package.json
```
Adi-204, line 16: Please don't remove this comment and actually in comment change package.json to .ageneratorrc

packages/templates/clients/websocket/python/template/client.py.js
```diff
@@ -1,26 +1,21 @@
...
     <File name={params.clientFileName}>
       <FileHeaderInfo
         info={info}
         server={server}
       />
       <Requires query={queryParams} />
-      <ClientClass clientName={clientName} serverUrl={serverUrl} title={title} queryParams={queryParams} operations={operations} />
+      <ClientClass asyncapi={asyncapi} params={params} />
```
Adi-204, line 20: and maybe we can pass `query={queryParams}` also so that we don't need to call `getQueryParams` again inside ClientClass. It is also the case for `getServer` called twice. So technically we should pass it also but then again their will be too many parameters passed so this need to be discussed @achaljhawar wdyt 🤔 ? Don't change it directly let's wait for @derberg input also.

@derberg can you please confirm this?

derberg: it's ok if `getQueryParams` is used in `ClientClass`. You don't know if client file template will always have access to query params. This is cleaner as it is implemented now

Adi-204 (commented):
@achaljhawar please update the snapshot by running `npm run test:update`

Adi-204 (approved):
LGTM!

derberg (approved):

derberg:
@achaljhawar sorry, you'll have to solve merge conflict issues

achaljhawar:
@derberg I have resolved the merge conflicts.

derberg (approved):
thanks @achaljhawar and @Adi-204 🙏🏼

PR #1630, opened 2025-06-30
test: remove code duplication of `integration.test.js`

### LINKED ISSUE #1549: check if we can have one `integration.test.js` with snapshots for all websocket client under `websocket/test` with as less code duplication as possible
if you look into `python/test/integration.test.js` you see that once I introduced `listFiles`, now basically all tests are the same, and only input arguments are different. I believe we can create a helper for generating snapshots that basically accepts different `templateParams` and `fixtures` and `testsResults`. There is too much boilerplate code. Also investigate, maybe we should have integration tests only in `websocket/test` and not in separate language-based folders

Adi-204:
**Description**
Remove `integration.test.js` from all websocket clients and move complete integration test setup in `/test` folder.

image

**Related issue(s)**
Fixes #1549


packages/templates/clients/websocket/dart/package.json
```diff
@@ -3,9 +3,8 @@
   "version": "0.0.1",
   "description": "This is a template generating Dart websocket client",
   "scripts": {
-    "test": "npm run test:cleanup && jest  --coverage",
+    "test": "jest  --coverage --passWithNoTests",
```
Adi-204, line 6: @derberg I added `--passWithNoTests` so that CI passes and don't throw error because Dart client has only integration test file which I removed in this PR and no component testing for now. We will remove this flag when any component testing code inside Dart.

Adi-204:
@derberg wdyt 🤔 about approach? I think it is following DRY approach and also it is now extremely simple to add new language integration test. I think writing additional testcases will also be super easy for new-commers and not hard to understand. 

//TODO
If we do merge this approach we have to update - https://github.com/asyncapi/generator/tree/master/packages#assumptions-and-principles before merging.  
Also need to update `README.md` of all the clients https://github.com/asyncapi/generator/tree/master/packages/templates/clients/websocket/javascript

packages/templates/clients/websocket/test/integration-test/integration.test.js
```diff
@@ -0,0 +1,107 @@
+
+const path = require('path');
+const { readFile, stat } = require('fs').promises;
+const { rm } = require('fs/promises');
```
derberg:
```suggestion
const { readFile, stat, rm } = require('fs/promises');
```

packages/templates/clients/websocket/test/integration-test/package.json
```diff
@@ -0,0 +1,43 @@
...
+        "@babel/core": "^7.26.0",
+        "@babel/preset-env": "^7.26.0",
+        "@babel/preset-react": "^7.25.9",
+        "jest-esm-transformer": "^1.0.0",
+        "@asyncapi/generator": "*",
+        "eslint": "^6.8.0",
+        "eslint-plugin-jest": "^23.8.2",
+        "eslint-plugin-react": "^7.34.1",
+        "eslint-plugin-sonarjs": "^0.5.0",
+        "rimraf": "^3.0.2"
```
derberg: looks like you're not use it anymore and rely on test-based cleanup, so remove this dependency

packages/templates/clients/websocket/test/integration-test/integration.test.js
```diff
@@ -0,0 +1,107 @@
+
```
derberg:
```suggestion
```

packages/templates/clients/websocket/test/integration-test/integration.test.js
```diff
@@ -0,0 +1,107 @@
...
+    testResultPath: path.resolve(__dirname, '../../javascript/test/temp/snapshotTestResult'),
+    template: path.resolve(__dirname, '../../javascript')
+  },
+  dart: {
+    clientFileName: 'client.dart',
+    testResultPath: path.resolve(__dirname, '../../dart/test/temp/snapshotTestResult'),
+    template: path.resolve(__dirname, '../../dart')
+  }
+};
+
```
derberg, line 30: love it! it clearly shows custom tests that indicate use cases that are not yet implemented in all clients - like out of the box TODO list. Eventually all tests shoudl be common

packages/templates/clients/websocket/python/package.json
```diff
@@ -3,9 +3,8 @@
   "version": "0.0.1",
   "description": "This is a template generating Python websocket client",
   "scripts": {
-    "test": "npm run test:cleanup && jest --coverage",
+    "test": "jest --coverage",
```
derberg, line 6: I guess https://github.com/asyncapi/generator/tree/master/packages/templates/clients/websocket/python/test/__snapshots__ should be removed too

Adi-204: @derberg I deleted it but I don't what went wrong but it is considering it as renamed new snapshot of `integration-test` I added. But in my local branch you can see their is no test/snapshot inside python 
https://github.com/Adi-204/generator/tree/integrationTest/packages/templates/clients/websocket/python

so we should be good after merging.

derberg: ok, sounds good then

Adi-204:
@derberg readme of all clients also updated but for Assumptions and Principles I think need your help 😅.

Adi-204:
@derberg yeah I think adding README for now inside https://github.com/asyncapi/generator/tree/master/packages/templates/clients/websocket makes more sense than `Development.md` because we are experimenting this right now so let's keep info inside `/packages` only later in future I am sure will we move up this README.md when we have more templates and protocols.

Adi-204:
Or wait actually we have https://github.com/asyncapi/generator/blob/master/packages/templates/clients/websocket/test/README.md but for now it is completely focusing on acceptance testing. wdyt 🤔? maybe add headings in this readme about acceptance test and integration test and write inside it?
More options lead to more confusion 😅

derberg:
good point, extend https://github.com/asyncapi/generator/blob/master/packages/templates/clients/websocket/test/README.md byt also add info in new readme in `/packages/templates/clients/websocket/` short info about tests basically and link to https://github.com/asyncapi/generator/blob/master/packages/templates/clients/websocket/test/README.md

Adi-204:
@derberg hope you like new testing docs 😄! any suggestions?

derberg (approved):
awesome job!

PR #1641, opened 2025-07-14
ci: add new pr review checklist workflow

### LINKED ISSUE #1639: Add CI workflow to auto-comment PR review checklist
Add a new GitHub Actions workflow that automatically posts an initial comment on each newly opened pull request.
The comment should include our PR review checklist, helping contributors know what maintainers look for, and making it easy for reviewers to quickly refer to the checklist.

Adi-204:
**Description**
Add new pr review checklist workflow

**Related issue(s)**
Fixes #1639

### DIFF
--- a/.github/pr-review-checklist.md
+++ b/.github/pr-review-checklist.md
@@ -1,0 +1,22 @@
+### What reviewer looks at during PR review
+The following are ideal points maintainers look for during review. 
+
+1. **PR Title**: Use a concise title that follows our [Conventional Commits](https://github.com/asyncapi/generator/blob/master/CONTRIBUTING.md#conventional-commits) guidelines and clearly summarizes the change using [imperative mood](https://cbea.ms/git-commit/#imperative) (it means `spoken or written as if giving a command or instruction`, like "add new helper for listing operations")
+    > **Note** - In Generator, prepend `feat:` or `fix:` in PR title only when PATCH/MINOR release must be triggered.
+
+1. **PR Description**: Clearly explain the issue being solved, summarize the changes made, and mention the related issue.
+    > **Note** - In Generator, we use [Maintainers Work board](https://github.com/orgs/asyncapi/projects/50) to track progress. Ensure the PR Description includes `Resolves #<issue-number>` or `Fixes #<issue-number>` this will automatically close the linked issue when the PR is merged and helps automate the maintainers workflow.
+
+1. **Documentation**: Update the relevant [Generator documentation](https://www.asyncapi.com/docs/tools/generator) to accurately reflect the changes introduced in the PR, ensuring users and contributors have up-to-date guidance.
+
+1. **Comments and JSDoc**: Write clear and consistent JSDoc comments for functions, including parameter types, return values, and error conditions, so others can easily understand and use the code.
+
+1. **DRY Code**: Ensure the code follows the **Don't Repeat Yourself** principle. Look out for duplicate logic that can be reused.
+
+1. **Test Coverage**: Ensure the new code is well-tested with meaningful test cases that pass consistently and cover all relevant edge cases.
+
+1. **Commit History**: Contributors should avoid force-pushing as much as possible. It makes it harder to track incremental changes and review the latest updates.
+
+1. **Template Design Principles Alignment**: While reviewing template-related changes in the `packages/` directory, ensure they align with the [Assumptions and Principles](https://github.com/asyncapi/generator/blob/master/packages/README.md#assumptions-and-principles). If any principle feels outdated or no longer applicable, start a discussion these principles are meant to evolve with the project.
+
+1. **Reduce Scope When Needed**: If an issue or PR feels too large or complex, consider splitting it and creating follow-up issues. Smaller, focused PRs are easier to review and merge.
--- a/.github/workflows/pr-review-checklist.yml
+++ b/.github/workflows/pr-review-checklist.yml
@@ -1,0 +1,17 @@
+name: PR Review Checklist
+
+on:
+  pull_request:
+    types: 
+      - opened
+
+jobs:
+  pr-review-list:
+    name: Post PR Review Checklist
+    runs-on: ubuntu-latest
+    steps:
+    - uses: peter-evans/create-or-update-comment@v4
+      with:
+        token: ${{ secrets.GH_TOKEN }}
+        issue-number: ${{ github.event.pull_request.number }}
+        body-path: '.github/pr-review-checklist.md'


Adi-204:
@derberg when I'm using PAT `secrets.GH_TOKEN` it is showing error of token maybe empty (this is currently in PR). I tried using default `secrets.GITHUB_TOKEN` also but it is then giving pr write permission error. I include permission field also but still it was throwing error, I think it must be change from repo settings also maybe. 
I also notice some workflows are use PAT and some not can you please explain this secrets ci config of asyncapi? 
so main problem I'm facing right now is write permission on PR in github.

.github/workflows/pr-review-checklist.yml
```diff
@@ -0,0 +1,20 @@
...
+jobs:
+  pr-review-list:
+    name: Post PR Review Checklist
+    runs-on: ubuntu-latest
+    steps:
+    - name: Checkout repository
+      uses: actions/checkout@v4
+
+    - name: Post initial PR checklist
+      uses: peter-evans/create-or-update-comment@v4
```
derberg: for security reasons it is critical to not point to specific tab release, but instead a SHA of the committ associated with this release: https://www.stepsecurity.io/blog/pinning-github-actions-for-enhanced-security-a-complete-guide

.github/workflows/pr-review-checklist.yml
```diff
@@ -0,0 +1,20 @@
+name: PR Review Checklist
+
+on:
+  pull_request:
```
derberg: the issues you see with tokens are related to the event type the action reacts to: https://runs-on.com/github-actions/pull-request-vs-pull-request-target/

in short, current workflow can be modified on a PR level and it runs in scope of this repo in given PR. In other words, anybody can modify such workflow in their PR and add a script that will send any token to remote location, and later use token for bad things. This is why GITHUB_TOKEN used in such workflow is read-only, and our custom GH_TOKEN is empty.

you should switch to `pull_request_target` that always run in `master` scope. So for example you will not see how it works after modifications on PR-level - using "write" tokens in such workflow is always safe. In other words workflow that reacts on `pull_request_target` is always ran on your PR using the workflow from `master` not from your PR/fork

.github/pr-review-checklist.md
```diff
@@ -0,0 +1,22 @@
+### What reviewer looks at during PR review
+The following are ideal points maintainers look for during review. 
+
```
derberg, line 3: add here a note that we discussed, like "yo, this is what maintainers do, so to speed up review process and time-to-merge, just go through the list and improve your PR accordingly

.github/pr-review-checklist.md
```diff
@@ -0,0 +1,22 @@
...
+1. **DRY Code**: Ensure the code follows the **Don't Repeat Yourself** principle. Look out for duplicate logic that can be reused.
+
+1. **Test Coverage**: Ensure the new code is well-tested with meaningful test cases that pass consistently and cover all relevant edge cases.
+
+1. **Commit History**: Contributors should avoid force-pushing as much as possible. It makes it harder to track incremental changes and review the latest updates.
+
+1. **Template Design Principles Alignment**: While reviewing template-related changes in the `packages/` directory, ensure they align with the [Assumptions and Principles](https://github.com/asyncapi/generator/blob/master/packages/README.md#assumptions-and-principles). If any principle feels outdated or no longer applicable, start a discussion these principles are meant to evolve with the project.
+
+1. **Reduce Scope When Needed**: If an issue or PR feels too large or complex, consider splitting it and creating follow-up issues. Smaller, focused PRs are easier to review and merge.
```
derberg: I think we also need to somehow address bots, like coderabit and sonarqubecloud

like that comments from these should not be ignored, but they are definitely discussable and if there is a comment "you do not agree with" add a comment with proper information

?

Adi-204: @derberg wait you mean like at end of file we tag coderabbit and sonarqubecloud and tell them to consider this points while review. so basically kind of prompt eng 🤔? idea seems to be nice, we can give it a try.

derberg: oh, no, that is now what I meant, but with coderabbit it might be interesting - although wouldn't it read it anyway? isn't it reading comments?

my point was that we just need another point - as reviewers, we need to check if the comments from bots were respected by contributor and taken into accoutn, and react or resolve comments from bots that we do not agree with

Adi-204: oh 😆 I overthink so hard. classic case of Overengineering

derberg: 😆

derberg: but the idea you had is not bad, only a matter of checking if coderabbit can do it.

once we merge this, and see first comment in action - we will see if coderabbit read it, and if not, we can play with coderabbit prompt support to see if "when asked explicitly" it will do something

Adi-204:
@derberg i think we would see action in play after merge should I remove synchronize and any additional changes you want?

derberg:
> i think we would see action in play after merge

yes, 100%
this is unfortunately the case with GH Actions. There is this tool `act` to test them but it is not the best for our cases. I basically always create a new workflow in my test repo and check how it works: https://github.com/derberg/test-experiment

> should I remove synchronize and any additional changes you want

what do you mean?

Adi-204:
> what do you mean?

@derberg We need to run this workflow only once when PR is open so why do we need "synchronize" inside pull_request_target.

derberg:
we don't, you can remove it, just this new workflow will not work with already existing PRs, but that is fine

.github/workflows/pr-review-checklist.yml
```diff
@@ -0,0 +1,21 @@
...
+jobs:
+  pr-review-list:
+    name: Post PR Review Checklist
+    runs-on: ubuntu-latest
+    steps:
+    - name: Checkout repository
+      uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683
+
+    - name: Post initial PR checklist
+      uses: peter-evans/create-or-update-comment@71345be0265236311c031f5c7866368bd1eff043
```
derberg: sorry, I forgot to share the practice we do here, when we use SHA, we also add a comment what release it is related too, so please look at other workflows how such comment looks like

btw, did you test it in sample repo as I mentioned?

Adi-204: I couldn't test it because I don't think secrets.GH_TOKEN will work I will have to config that test-experiment repo also right? I will set it up soon and will use always in future.

Adi-204: @derberg done have a look at - https://github.com/Adi-204/ci-testing/pull/2

derberg:
@Adi-204 as always, all well done 👏🏼 thanks for manual testing

@Florence-Njeri would you like to have a look, no only to review but also maybe you have some input to add to the review checklist? The idea is to always drop https://github.com/Adi-204/ci-testing/pull/2#issuecomment-3074414561 comment automatically to every PR

Florence-Njeri (approved):

PR #1647, opened 2025-07-22
refactor: move `FileHeaderInfo` component to `@asyncapi/generator-components`

### LINKED ISSUE #1447: [Test]: Add `FileHeaderInfo.test.js` for python websocket client

Adi-204:
**Description**
Move FileHeaderInfo component to `@asyncapi/components`.

**Related issue(s)**
Fixes #1447 

## Summary by CodeRabbit

* **New Features**
  * Introduced a unified FileHeaderInfo component that formats file header comments for multiple languages (JavaScript, Python, Dart, and others) using AsyncAPI metadata.

* **Improvements**
  * Updated templates to use the new shared FileHeaderInfo component from a central package, ensuring consistent header formatting and language support.

* **Bug Fixes**
  * Corrected export paths for Models and added export for FileHeaderInfo in the components package.

* **Tests**
  * Added comprehensive tests for the new FileHeaderInfo component covering various languages and scenarios.
  * Removed outdated tests related to the old FileHeaderInfo implementations.


packages/components/src/components/fileheaderinfo.js
```diff
@@ -0,0 +1,64 @@
...
+ *
+ * @param {Object} props - Component props.
+ * @param {object} props.info - Info object from the AsyncAPI document.
+ * @param {object} props.server - Server object from the AsyncAPI document.
+ * @param {Language} props.language - Programming language used for comment formatting.
+ * @returns {JSX.Element} Rendered file header.
+ */
+export function FileHeaderInfo({ info, server, language }) {
+  const commentChar = commentStyles[language] || '//';
+  const line = `${commentChar}${'/'.repeat(70)}`;
```
derberg: in case of python we should respect `#` and use it not `/`

derberg (commented):
looks great, just one thing to fix

I noticed we are not consistent in components and do all lower case and you followed. Can you in this PR make sure that components in components library, have filenames that are one to one representing the name of the component? the same in tests? so `FileHeaderInfo.js` I mean, not all lower case

packages/components/src/components/FileHeaderInfo.js
```diff
@@ -0,0 +1,64 @@
...
+ *
+ * @param {Object} props - Component props.
+ * @param {object} props.info - Info object from the AsyncAPI document.
+ * @param {object} props.server - Server object from the AsyncAPI document.
+ * @param {Language} props.language - Programming language used for comment formatting.
+ * @returns {JSX.Element} Rendered file header.
+ */
+export function FileHeaderInfo({ info, server, language }) {
+  const commentChar = commentStyles[language] || '//';
+  const line = `${commentChar.repeat(50)}`;
```
derberg: in case of dart it is a lot, and goes in 2 lines basically

if you want to keep it simple then maybe do 25? 

or maybe we do it wrong and it should look completely different in case of dart - I see they just do `///` in general

it wont be complex if in case of `line` you will also keep a language specific config, like for `commentStyles`

Adi-204: @derberg thanks for the idea do you like it now? I'm bad with design 😆

PR #1655, opened 2025-08-01
refactor: move out the `loadTemplateConfig` from `generator.js`

### LINKED ISSUE #1582: refactor: move out the loadTemplateConfig from Gigantic generator.js
After the merging of #1485 . We have to move the loadTemplateConfig from generator.js. We have to do it like done for template config validation i(https://github.com/asyncapi/generator/blob/master/apps/generator/lib/templateConfigValidator.js)

We will be having a folder `templateConfig` in which there iwill be `validator.js` and `loader.js`

Adi-204:
**Description**
Moved the loadTemplateConfig and loadDefaultValues function to separate file for clean code.

**Related issue(s)**
Fixes #1582 

## Summary by CodeRabbit

- New Features
  - Support template configuration via .ageneratorrc (YAML), with fallback to package.json.
  - Applies default parameter values without overriding user-specified inputs.

- Refactor
  - Centralized template configuration loading into a dedicated module for more reliable behavior.

- Tests
  - Added comprehensive tests for config loading and defaults.
  - Updated existing tests to align with the new loading mechanism.

- Chores
  - Updated internal import paths and mocks to reflect the new configuration layout.


apps/generator/lib/templates/config/loader.js
```diff
@@ -0,0 +1,65 @@
+const path = require('path');
+const { readFile } = require('../../utils');
+const log = require('loglevel');
+
```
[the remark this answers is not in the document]
Adi-204: @coderabbitai I think I just need to do refactoring and in original code it was written in this way only. 
Lets wait for @derberg review he is 100 times better than you.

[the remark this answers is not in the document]
derberg: yeah, I think it is cleaner to have all requires at the top

apps/generator/lib/templates/config/loader.js
```diff
@@ -0,0 +1,65 @@
...
+const CONFIG_FILENAME = 'package.json';
```
derberg: use the opportunity to refactro this too, so we now have 2 config files
- `PRIMARY_CONFIG_FILE` which is `.ageneratorrc`
- `SECONDARY_CONFIG_FILE` which is `package.json`

apps/generator/lib/generator.js
```diff
@@ -286,7 +286,8 @@ class Generator {
     this.templateName = templatePkgName;
     this.templateContentDir = path.resolve(this.templateDir, TEMPLATE_CONTENT_DIRNAME);
 
-    await this.loadTemplateConfig();
+    this.templateConfig = await loadTemplateConfig(this.templateDir);
+    loadDefaultValues(this.templateConfig, this.templateParams);
```
derberg: any reasons why we need to call `loadDefaultValues` here instead of calling it inside `loadTemplateConfig`

Adi-204: yes this was main thing why previous contributor was doing weird this binding stuff. So the code inside `loadDefaultValues` 
```js
Object.defineProperty(templateParams, dv, {
            enumerable: true,
            get() {
                return parameters[dv].default;
            }
        })
 ```
 this is doing changes in object inplace so if we pass through loadTemplateConfig apparently it was not able to perform inplace changes so I thought to get it out and call directly and then it was working.

Comment (derberg) [reply apps/generator/lib/generator.js]: did you try changing

```
get() {
                return parameters[dv].default;
            }
```

to

```
get: (() => {
                    const key = dv;
                    return function () {
                        return parameters[key].default;
                    };
                })()
```

Comment (Adi-204) [reply apps/generator/lib/generator.js]: yes I did tried something with arrow function also but let me try once again.

Comment (Adi-204) [reply apps/generator/lib/generator.js]: image

image

Yes same all the test of generator passes but than integration test of websocket clients all are failing. Should I push code if you want to see in your local 🤔

Comment (derberg) [reply apps/generator/lib/generator.js]: yeah, push all the code you have to call `loadDefaultValues` from inside `loadTemplateConfig` so later it is easier to revert for you

**Comment by Adi-204** [comment]
@derberg I tried some different approaches but it is not working if pass params through the function. Should we go to original approach which I did or wait for some time?

**Comment by derberg** [comment]
@Adi-204 please revert to the "working" version

**Comment by derberg** [comment]
thanks for the revert, 

I took your code and produced a working version: https://github.com/derberg/generator/pull/5/files

the problem is I cannot open a PR from my fork against your fork, you are somehow not on the list of forks, which is weird

Screenshot 2025-08-20 at 14 55 08

**Comment by Adi-204** [comment]
@derberg I also added testing. The testing for loadDefaultValues is same as original code and loadTemplateConfig I tried. It is ready to review.

File: apps/generator/test/generator.test.js [hunk 4]
```diff
@@ -209,8 +211,8 @@ describe('Generator', () => {
 
       expect(xfsMock.mkdirpSync).toHaveBeenCalledWith(__dirname);
       expect(gen.installTemplate).toHaveBeenCalledWith(true);
-      expect(gen.loadTemplateConfig).toHaveBeenCalled();
-      expect(gen.loadTemplateConfig).toHaveBeenCalled();
+      expect(templateConfigLoader.loadTemplateConfig).toHaveBeenCalled();
```
derberg, line 214: Weird that we have this duplicated, same in line 238-39
I did it in my PR but don't remember why :D did you try to change it and test once?

Adi-204: yup it is working without duplicate. I will remove duplicates.

PR #1738, opened 2025-10-02
docs: prevent </dd> from breaking into newline in api.md

DivitJain26:
**Description**
- Fixed the `</dd>` being pushed to a new line in generated `api.md`.
- Updated the Handlebars template `params-table.hbs` to correctly inline the closing `</dd>`.
- Regenerated `docs/api.md` using the updated template.

**Related issue(s)**

- Fixes #791 (Extending Asset with Contract or Clause causes Archive error)
Screenshot 2025-10-02 122507

## Summary by CodeRabbit

* **Documentation**
  * Reformatted API docs for improved readability: updated markup, inline styling, and reflowed descriptions and examples.
  * Added a new global-index layout that presents globals with signatures and descriptions in a definition-list style for easier scanning.
  * Updated parameter tables to render as definition-list containers for clearer structure.
  * No changes to public API signatures or behavior.

Issue #1737, opened 2025-10-01
modify api.md related templates to not break website build


derberg:
changes like this: https://github.com/asyncapi/generator/pull/1734/files

if pushed to https://github.com/asyncapi/website - break website

reason is that `</dd>` is in newline, and should not be.

it cannot be fixed in https://github.com/asyncapi/generator/blob/master/apps/generator/docs/api.md as this document is generated using these templates https://github.com/asyncapi/generator/tree/master/apps/generator/docs/jsdoc2md-handlebars

so goal is to:
- investigate which handlebar file is responsible for generation of `</dd>` and that this element is not pushed to newline
- fix the handlebar file 
- PR shoudl also include fixed api.md - just run proper generate assets npm script to run the generation locally

nitinawari:
@derberg can i work on this ?

dhruthi-ty:
Hey @derberg 

I would like to work on the issue related to the newline problem for </dd> in the generated api.md file on the AsyncAPI website.

Problem Breakdown:
The api.md file is generated automatically using Handlebars templates located in apps/generator/docs/jsdoc2md-handlebars.
Currently, the closing tag </dd> is placed on a new line.
This breaks the website layout.
The file api.md itself cannot be fixed directly, because it is auto-generated from these templates.

Proposed Approach:
Identify which Handlebars template generates the <dd> element and is responsible for the newline.
Update the template so that the </dd> closing tag stays on the same line as the content.
Run the npm generate script locally to regenerate api.md with the corrected formatting.
Submit a PR including both the fixed Handlebars template and the regenerated api.md.

 I'll start working on it once assigned! Thank you!

nehaaprasad:
Hello!
@derberg 
now empty <dd></dd> tags in api.md have been eliminated by adding conditional rende

lightning-sagar:
@derberg can i work on this??

batchu5:
I have investigated that this change comes from dmd folder in node_modules. so to avoid this </dd> in new line,  we need to create the new .hbs file in the docs folder.
If you assign me, I would like to work on it

Adi-204:
@varshitha-besthu we don't assign issues, please feel free to create a PR.

PR #1739, opened 2025-10-06
fix: prevent empty <dd> tags in api.md from breaking website builds

naaa760:
**Description**

- Fix empty `<dd></dd>` tags in generated api.md that break website builds
- Add missing docs.hbs template referenced in npm script
- Create custom main-index-dl.hbs and global-index-dl.hbs templates with conditional rendering
- Update docs generation script to include new template partials
- Regenerate api.md with fixed templates to eliminate empty description tags

**Related issue(s)**
Fixes #1734

### DIFF
--- a/apps/generator/docs/jsdoc2md-handlebars/docs.hbs
+++ b/apps/generator/docs/jsdoc2md-handlebars/docs.hbs
@@ -1,0 +1,5 @@
+{{>header~}}
+{{>body}}
+{{>member-index~}}
+{{>separator~}}
+{{>members~}}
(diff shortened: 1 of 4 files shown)


aasim-maverick:
Hey @naaa760, just saw your commits, the OP of gh-1737 expects the PR to include the fixed `api.md`. 

Try running `npm run docs` and see if this is preventing `</dd>` from breaking into a new line. If yes, then add that to a commit too.  

Edit: I think the tag still breaks into a new line.

derberg:
Please do not open PR if there is already another one opened and in review

https://github.com/asyncapi/generator/pull/1738

Issue #1741, opened 2025-10-15
[BUG] client java channel name gives invalid java syntax


lfvJonas:
### Describe the bug.

I have a AsyncAPI that has the following channels

```
"channels": {
    "se.lfv.sysdev.swim.swim-opmet-consumer.METAR": {
      "address": "se.lfv.sysdev.swim.swim-opmet-consumer.METAR",
      "messages": {
        "HeartbeatMessage": {
          "$ref": "#/components/messages/HeartbeatMessage"
        },
        "MetarData": {
          "$ref": "#/components/messages/MetarData"
        },
        "ProviderException": {
          "$ref": "#/components/messages/ProviderException"
        }
      },
      "bindings": {}
    },
    "se.lfv.sysdev.swim.swim-opmet-consumer.SIGMET": {
      "address": "se.lfv.sysdev.swim.swim-opmet-consumer.SIGMET",
      "messages": {
        "HeartbeatMessage": {
          "$ref": "#/components/messages/HeartbeatMessage"
        },
        "ProviderException": {
          "$ref": "#/components/messages/ProviderException"
        },
        "SigmetData": {
          "$ref": "#/components/messages/SigmetData"
        }
      },
      "bindings": {}
    },
    "se.lfv.sysdev.swim.swim-opmet-consumer.TAF": {
      "address": "se.lfv.sysdev.swim.swim-opmet-consumer.TAF",
      "messages": {
        "HeartbeatMessage": {
          "$ref": "#/components/messages/HeartbeatMessage"
        },
        "ProviderException": {
          "$ref": "#/components/messages/ProviderException"
        },
        "TafData": {
          "$ref": "#/components/messages/TafData"
        }
      },
      "bindings": {}
    }
  },
```

this generates this java method:

```
@OnTextMessage
public void se.lfv.sysdev.swim.swim-opmet-consumer.SIGMET_send_topic(String message, WebSocketClientConnection connection) {
      Log.info("Received text message: " + message);
  }

```

The method name ´se.lfv.sysdev.swim.swim-opmet-consumer.SIGMET_send_topic´ is invalid and the code does not compile

### Expected behavior

the method name should be a valid java method name:

exemple: `seLfvSysdevSwimSwimOpmetConsumerSIGMETSendTopic` (ugly but valid)

### Screenshots

no screenshots

### How to Reproduce

1. have a valid asyncapi with a channel with a name that contains ´.´
[...]

PR #1745, opened 2025-10-19
fix: solve invalid java method name generation and handling multiple handlers

ssala034:
**Description**
This PR contains
a) The fix for #1741 
Current template was automatically using operation names for java method names. 
Fixed by parsing operation names and converting to camelCase so its now a valid Java method name.
 
Screenshot 2025-10-19 104356

b) Multiple send operations
I noticed that the template could not  support schemas with multiple sendOperations. Since I never tested with a template that had one never came to my mind.
Fixed by making a singular onTextMessage handler and other helper function for each operation

c) Formatting
There was some formatting issues with the `onClose` and `onOpen` modules in components for java. I just added some small changes. 

d) Dependency
I forgot to add the k8s quarkus dependency in the websocket template for java/quarkus. I added it and update the readme, so it tells users how use k8s + quarkus

**Related issue(s)**
#1741

## Summary by CodeRabbit

* **New Features**
  * Generated WebSocket clients now use a single dispatch for incoming text messages, with per-operation handlers (camel-cased) and a default no-op handler; connectors and timed connections receive operation info for operation-aware behavior.

* **Refactor**
  * Consolidated per-operation handling into one dispatch path; query-params rendering skips empty parameter lists.

* **Documentation**
  * Added "Deploying with Kubernetes" guidance.

* **Chores**
  * Updated platform version and added Kubernetes/container image build support.

* **Style**
  * Adjusted formatting and spacing for connection lifecycle output.

### DIFF
--- a/.changeset/lucky-zebras-wonder.md
+++ b/.changeset/lucky-zebras-wonder.md
@@ -1,0 +1,10 @@
+---
+"@asyncapi/generator-components": patch
+"@asyncapi/generator": patch
+---
+
+Fix invalid Java method name generation and properly handle multiple WebSocket handlers
+
+- Generate valid Java method names for WebSocket operation handlers with strange identifiers.
+- Produce distinct @OnTextMessage handler methods for each send operation and add safe default routing for unrecognized messages.
+- Update onClose/onOpen formatting.


Adi-204 (approved):
lgtm!

ssala034:
> lgtm but you need to add changeset as otherwise the fix will not be released
> 
> https://github.com/asyncapi/generator/blob/master/Development.md#release-process

Should this be done in this PR or a separate one.

derberg:
this one, no need for separate

ssala034:
Okay 👍 

And the changset is to update the release components library. Correct?

Issue #1768, opened 2025-11-24
Generator do not handle template params that default to `false`


derberg:
> checked with latest 3.3.1
> 
> Problem might be in https://github.com/asyncapi/generator/blob/master/apps/generator/lib/templates/config/loader.js#L47. Tests should be checked for `loadDefaultValues`.
> 
> in `html-template` config https://github.com/asyncapi/html-template/blob/master/package.json#L90-L93 `singleFile` default is `false`. So if you do not pass this parameter, `false` value should be injected with context into generation process.
> 
> This `false` default value is critical for another setting (conditional generation): https://github.com/asyncapi/html-template/blob/master/package.json#L63-L75 which means that required `css` and `js` folders will be rendered in output only fi `singleFile` is set to `false`. 
> 
> `asyncapi generate fromTemplate test/spec/asyncapi_v3.yml ./ -o dupa --use-new-generator --force-write -p singleFile=false` works as expected, `false` goes through, so directories are there. Of course flah `singleFile=true` works as well.
> 
> ---
> Problem is that during generation in case like `asyncapi generate fromTemplate test/spec/asyncapi_v3.yml ./ -o dupa --use-new-generator --force-write` the template gets `singleFile` as undefined (doesn't get it at all). I can only see the following defaults
> 
> ```
> {
>   "outFilename": "index.html",
>   "pdfTimeout": 30000
> }
> ```
> 
> So there is for sure a bug in generator, it is not handling default values set to `false` 

 _Originally posted by @derberg in [#744](https://github.com/asyncapi/html-template/issues/744#issuecomment-3571549669)_

lightning-sagar:
Working on it!!🫡

VIVEKDHEDHI021:
Thanks for the detailed explanation. I can confirm this issue as well.
The generator currently fails to load template parameters whose default value is false, which causes the HTML template to behave incorrectly when singleFile isn’t explicitly passed.

During generation, the template context only contains:

`
{
  "outFilename": "index.html",
  "pdfTimeout": 30000
}
`
The expected singleFile: false value is missing entirely.
This leads to incorrect conditional rendering, since the template logic relies on:
 `{{#unless singleFile}}
  // render css/js folders
{{/unless}}
`
Because singleFile becomes undefined instead of false, required folders are not created.

[...]

PR #1769, opened 2025-11-25
fix: correctly detect default values including false in template parameters

lightning-sagar:
**Description**

This PR fixes an issue where template parameters with a default value of `false` (e.g., `singleFile`) were not being applied, causing them to appear as `undefined` inside templates.

The problem was caused by this check:

```js
 const defaultValues = Object.keys(parameters || {}).filter(key => parameters[key].default);
```

Since `false` is falsy, it was incorrectly ignored.
The fix updates the condition to properly detect all default values:

```js
const defaultValues = Object.keys(parameters || {}).filter(key => parameters[key].default !== undefined);
```

### How I verified the fix

* Patched the generator locally and added simple debug logs in the HTML template:
  [https://github.com/lightning-sagar/html-template/blob/test/template-params/components/index.js#L48-L49](https://github.com/lightning-sagar/html-template/blob/test/template-params/components/index.js#L48-L49)

* Ran the generator using:

  ```
  asyncapi generate fromTemplate test/spec/asyncapi_v3.yml ./ -o dupa --use-new-generator --force-write
  ```

* Output confirmed the fix:
  [https://github.com/lightning-sagar/html-template/blob/test/template-params/dupa/index.html#L15-L16](https://github.com/lightning-sagar/html-template/blob/test/template-params/dupa/index.html#L15-L16)
  (`param: false | computed: false`)

---

**Related issue(s)**

Fixes [https://github.com/asyncapi/generator/issues/1768](https://github.com/asyncapi/generator/issues/1768)

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected default-value detection so falsy defaults (0, "", false) are recognized and applied, fixing conditional generation and injection of template defaults.

* **Documentation**
  * Clarified guidance to align parameter default types with expected validation types to avoid mismatches and generation issues.

* **Chores**
  * Added a release note entry describing the fix for template parameter default handling.

### DIFF
--- a/Development.md
+++ b/Development.md
@@ -31,7 +31,7 @@ git clone https://github.com/{your_username}/generator.git
 cd generator
 ```
 
-After cloning the repository, you should setup the fork properly and configure the `remote` repository as described [here](https://github.com/asyncapi/community/blob/master/git-workflow.md)
+After cloning the repository, you should setup the fork properly and configure the `remote` repository as described in the [AsyncAPI git workflow guidelines](https://github.com/asyncapi/community/blob/master/docs/010-contribution-guidelines/git-workflow.md)
 
 2. Install dependencies:
 
--- a/apps/generator/docs/template.md
+++ b/apps/generator/docs/template.md
@@ -29,7 +29,7 @@ Baked-in templates benefit from:
 
 In contrast, **standalone templates** (described below) are maintained as independent Node.js packages, may live in separate repositories, and can be managed or installed separately from the core generator.
 
-Learn more from document [Baked-in Templates](#baked-in-templates).
+Learn more from document [Baked-in Templates](baked-in-templates).
 
 ### Standalone templates
 
--- a/apps/generator/test/test-templates/react-template/package-lock.json
+++ b/apps/generator/test/test-templates/react-template/package-lock.json
@@ -16,6 +16,7 @@
       "version": "2.3.0",
       "resolved": "https://registry.npmjs.org/@ampproject/remapping/-/remapping-2.3.0.tgz",
       "integrity": "sha512-30iZtAPgz+LTIYoeivqYo853f02jBYSd5uGnGpkFV0M3xOt9aN73erkgYAmZU43x4VfqcnLxW9Kpg3R5LC4YYw==",
+      "peer": true,
       "dependencies": {
         "@jridgewell/gen-mapping": "^0.3.5",
         "@jridgewell/trace-mapping": "^0.3.24"
@@ -2334,7 +2335,6 @@
       "version": "24.1.0",
       "resolved": "https://registry.npmjs.org/@types/node/-/node-24.1.0.tgz",
       "integrity": "sha512-ut5FthK5moxFKH2T1CUOC6ctR67rQRvvHdFLCD2Ql6KXmMuCrjsSsRI9UsLCm9M18BMwClv4pn327UvB7eeO1w==",
-      "peer": true,
       "dependencies": {
         "undici-types": "~7.8.0"
       }
@@ -2389,7 +2389,6 @@
       "version": "8.17.1",
       "resolved": "https://registry.npmjs.org/ajv/-/ajv-8.17.1.tgz",
       "integrity": "sha512-B/gBuNg5SiMTrPkC+A2+cW0RszwxYmn6VYxB/inlBStS5nx6xHIt/ehKRhIMhqusl7a8LjQoZnjCs5vhwxOQ1g==",
-      "peer": true,
       "dependencies": {
         "fast-deep-equal": "^3.1.3",
         "fast-uri": "^3.0.1",
@@ -2633,7 +2632,6 @@
           "url": "https://github.com/sponsors/ai"
         }
       ],
-      "peer": true,
       "dependencies": {
         "caniuse-lite": "^1.0.30001726",
         "electron-to-chromium": "^1.5.173",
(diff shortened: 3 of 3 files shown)


lightning-sagar:
Note: After fixing this and **changing** the value from the string "false" to the boolean false, the CSS/JS will generate correctly🫡

By using `asyncapi generate fromTemplate test/spec/asyncapi_v3.yml ./ -o dupa --use-new-generator --force-write -p singleFile="false"` this command **not by**  `-p singleFile=false`

apps/generator/lib/templates/config/loader.js
```diff
@@ -46,7 +46,7 @@ async function loadTemplateConfig(templateDir, templateParams) {
  */
 function loadDefaultValues(templateConfig, templateParams) {
     const parameters = templateConfig.parameters;
-    const defaultValues = Object.keys(parameters || {}).filter(key => parameters[key].default);
+    const defaultValues = Object.keys(parameters || {}).filter(key => parameters[key].default !== undefined);
```
derberg: why not below
```suggestion
    const defaultValues = Object.keys(parameters || {}).filter(key => hasOwnProperty('default'));
```

what happens in your case when someone sets a property with default set to `undefined`

lightning-sagar: Using `hasOwnProperty("default")` works too, but I used `default !== undefined` because it **excludes** only the parameters that don’t **define a default in the template**.

In my case, when a parameter’s default is undefined (meaning no default was defined), it simply won’t appear in the defaultValues list. The list include real values (like 0, false, and "")

if you prefer, I can switch this to `parameters[key].hasOwnProperty("default")`

Jatin24062005:
@lightning-sagar  I reviewed the changes you made. Using: `filter(key => parameters[key].default !== undefined);` which i say is not bad does fix the false issue, but it may introduce new problems. For example, templates can intentionally set: `"default": null
`  or even: `"default": undefined
`  In those cases, !== undefined would incorrectly treat these values as valid defaults. to Avoid that problem @derberg Solution to use `   const defaultValues = Object.keys(parameters || {}).filter(key => hasOwnProperty('default'));`  but it also have a  major issue This checks whether the template author actually defined the default field, regardless of its value. It avoids accidentally ignoring defaults like false, 0, or "".

We avoid using obj.hasOwnProperty() directly because it can throw errors if the object does not inherit from Object.prototype. So  to Avoid crashing and error we can use explicit version of js something like `const defaultValues = Object.keys(parameters || {}).filter(key =>
  Object.prototype.hasOwnProperty.call(parameters[key], "default")
);`
 this guarantees no crashes and works with all template config objects.
>> https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/hasOwnProperty
image

lightning-sagar:
yes, that makes sense. your solution is definitely more robust.
using `Object.prototype.hasOwnProperty.call(...)` handles both JSON and JS-generated templates safely, and avoids any issues with falsy defaults or prototype-less objects. thanks for the help!

update: I switched to` Object.hasOwn(...)` initially since it reads cleaner.

lightning-sagar:
@derberg I noticed coderabbit flagged a bunch of issues in files I didnt touch (Keeper, WebSocket client templates, Constructor.js, etc) looks like these came in from the recent merge into master, not from the changes in this pr

before I do anything, could you let me know what you'd prefer?

- should I rebase/merge my branch with the latest master and fix those warnings here or

- should this pr stay focused only on the default-parameter fix, and the other warnings be handled separately?

just want to keep the scope clean and avoid mixing unrelated fixes

lightning-sagar:
@derberg please review whenever you have time

derberg (commented):
@lightning-sagar change is good. Ignore rabbit.

most important is how did you test that it works? solves the html-template error?

did you do some manual testing?

after v5 release we have dedicated `node ./test/cli` option to run generator test cli locally from local sources against any template, you can also point to template in your local drive

lightning-sagar:
Hi @derberg, sorry for the late response...

I was testing this manually in two ways:
* using `npm link` between the generator and `html-template`
* and by patching the generator code directly inside `node_modules`

this confirmed the generator fix works (the default `singleFile: false` is now working)
but, it also exposed an issue on the html-template side: the template currently expects `singleFile` as a **string** (`"false"`), while the generator now correctly provides it as a boolean (`false`). Because of this type mismatch, CSS/JS conditional generation still doesnt trigger as expected

so, there are two possible solution:
1. update `html-template` to treat `singleFile` as a boolean  or
2. change the default in `html-template/package.json` to `"false"` (string) and handle it consistently as a string...

let me know which approach should i move forward with?

derberg:
did you check it with `node ./test/cli` ? 
when you run a command with local cli, and you pass `false`, are you sure CLI is not affecting the type of passed parameter?

you'd have to verify with docs, and afaik we explain that whatever you pass through CLI, no matter what CLI - it is always `string` and then you need to properly handle it further. Generator should not guess if `false` is intentional string or should cast to boolean. If we do casting in Generator, that might be a breaking change

lightning-sagar:
@derberg, yes you are right and I did test this using `node ./test/cli.js` already

when `singleFile` is passed via CLI (e.g. `-p singleFile=false`), the CLI correctly treats it as a `string`, and that path works as expected...

the issue only appears when `singleFile` is not passed at all... in that case, [loadDefaultValues](https://github.com/asyncapi/generator/blob/master/apps/generator/lib/templates/config/loader.js#L16) injects the default from `html-template/package.json`, where `singleFile` is defined as `false` (boolean). That boolean value is then passed directly into the template context... not by cli

so the flow is:

* CLI param passed -> value is a string (`"false"`) -> works
* CLI param not passed → default comes from template config → value is boolean (`false`)
* therefore template logic expects a string → css/js conditional generation doesnt trigger

I agree the generator should not cast values coming from CLI, as that would be a breaking change.... `the generator fix is correct`....  the remaining issue is the type mismatch in `html-template` when handling default (boolean) vs CLI (string) values.

that why we need to update the `html-template/package.json`, by converting `singleFile` default value to [`"false"` (string)](https://github.com/lightning-sagar/html-template/blob/test/template-params/package.json#L93) or [update validation](https://github.com/lightning-sagar/html-template/blob/test/template-params/package.json#L63C4-L76C7)

derberg:
ok, then, imho we need to change the default to string: https://github.com/lightning-sagar/html-template/blob/test/template-params/package.json#L93C23-L93C25

please in this PR also explain that if someone uses `conditionalGeneration` they should make sure that types defined in it, must match the type of default value from parameter config: https://github.com/asyncapi/generator/blob/master/apps/generator/docs/configuration-file.md

derberg:
we have https://github.com/asyncapi/html-template/pull/757 that we can already use

lightning-sagar:
> we have [asyncapi/html-template#757](https://github.com/asyncapi/html-template/pull/757) that we can already use

yes, that was exactly my main idea as well..
also @derberg, please let me know if the updated docs wording looks ok from your side or if you want any change??

Issue #1771, opened 2025-11-28
[BUG] deoptimisation warning


lightning-sagar:
### Describe the bug

**Generator:** latest
**Template:** latest

```
[BABEL] Note: The code generator has deoptimised the styling of /Users/roman/Code/GitHub/**********/node_modules/@asyncapi/html-template/template/js/asyncapi-ui.min.js as it exceeds the max of 500KB
```

### Expected behavior

The warning should not appear in the console, or there should be a way to disable it.

### Screenshots

<img width="1297" height="45" alt="Image" src="https://github.com/user-attachments/assets/324398aa-97a6-4268-bf7a-7883c8de0291" />

### How to Reproduce

1. Install the latest generator and `html-template`.
2. Run the `asyncapi generator generate` command using `html-template`.
3. Observe the output in the console.

### 🖥️ Device Information (optional)

* **Operating System (OS):**
* **Browser:**
* **Browser Version:**

### 👀 Have you checked for similar open issues?

* [x] I checked and didn't find a similar issue.

### 🏢 Have you read the Contributing Guidelines?

* [x] I have read the [Contributing Guidelines](https://github.com/asyncapi/.github/blob/master/CONTRIBUTING.md).

### Are you willing to work on this issue?

Yes, I am willing to submit a PR!

---

This issue was noticed in [#717](https://github.com/asyncapi/html-template/issues/717).

---

SHUBHANSHU602:
Hey , can I take this?

lightning-sagar:
sorry @SHUBHANSHU602 , I am working on this issue. please look for the stale issue or some other issue in the mean time.

Jatin24062005:
@lightning-sagar @derberg  can i look into this issue and try to solve?

lightning-sagar:
Sorry this issue is already fixed in latest version of html-template and generator.. So I am closing the issue

PR #717, opened 2021-11-30
chore: add CODEOWNERS file

jonaslagoni:
**Description**
This PR adds the maintainers as listed on https://www.asyncapi.com/community/tsc

PR #1772, opened 2025-11-28
fix: exclude asyncapi-ui.min.js from transpilation to prevent Babel deoptimisation warning

lightning-sagar:
**Description**

This PR fixes the long-standing Babel warning:

```
[BABEL] Note: The code generator has deoptimised the styling of asyncapi-ui.min.js as it exceeds the max of 500KB
```

The issue occurs because the React transpiler processes `asyncapi-ui.min.js`, which is a large minified dependency that should **not be transpiled**.

### What was changed

* Updated the React SDK transpiler to **ignore** `asyncapi-ui.min.js` before passing files to Rollup/Babel.
* Implemented a simple filter:

  ```js
  files = files.filter(f => !f.endsWith("asyncapi-ui.min.js"));
  ```

  (and equivalent logic where needed)
* After applying the fix, the generator runs without any Babel deoptimisation warnings.

### 📌 Additional note

The old repo (`asyncapi-archived-repos/generator-react-sdk`) is archived, so this change was made in the active path:

```
apps/react-sdk/src/transpiler/transpiler.ts
```

**Related issue(s)**

Fixes: *<link to issue you opened>*
#1771 
[#717](https://github.com/asyncapi/html-template/issues/717)

## Summary by CodeRabbit

* **Bug Fixes**
  * Reduced noisy build output by silencing known non-actionable tool messages during normal runs.
  * Ensured build warnings are surfaced in debug mode while keeping normal runs quieter.
  * Improved reliability of the transpile step by guaranteeing stderr behavior and build resources are restored/closed even on errors.

### DIFF
--- a/.coderabbit.yaml
+++ b/.coderabbit.yaml
@@ -1,4 +1,4 @@
-# CodeRabbit Configuration File
+# yaml-language-server: $schema=https://coderabbit.ai/integrations/schema.v2.json
 
 language: en-US
 reviews:
@@ -12,19 +12,45 @@ reviews:
   auto_review:
     enabled: true
     auto_incremental_review: true
+    ignore_usernames: [dependabot, asyncapi-bot]
   review_status: true
   commit_status: true
   changed_files_summary: true
   assess_linked_issues: true
   related_issues: true
   related_prs: true
+  pre_merge_checks:
+    title:
+      mode: error
+      requirements: "Use a concise title that follows our Conventional Commits guidelines and clearly summarizes the change using imperative mood (it means spoken or written as if giving a command or instruction, like 'add new helper for listing operations'). In Generator, prepend feat: or fix: in PR title only when PATCH/MINOR release must be triggered."
+  tools:
+    actionlint:
+      enabled: true
+    github-checks:
+      enabled: true
+      timeout_ms: 900000  # Wait 15 min for checks to complete
+
 chat:
   auto_reply: false
 
 knowledge_base:
+  # Don't opt out - this enables all learning features
+  opt_out: false
+  web_search:
+    enabled: true
+  # Code Guidelines - automatically detects team standards
+  code_guidelines:
+    enabled: true
+    filePatterns:
+      # Add custom patterns here if you have other guideline files:
+      - ".github/pr-review-checklist.md"
+      - "Development.md"
+      - "CONTRIBUTING.md"
   learnings:
     scope: auto
   issues:
     scope: auto
   pull_requests:
     scope: auto
+  mcp:
+    usage: auto
--- a/.github/workflows/scripts/mailchimp/package-lock.json
+++ b/.github/workflows/scripts/mailchimp/package-lock.json
@@ -44,13 +44,28 @@
       "resolved": "https://registry.npmjs.org/asynckit/-/asynckit-0.4.0.tgz",
       "integrity": "sha1-x57Zf380y48robyXkLzDZkdLS3k="
     },
-    "node_modules/call-bind": {
+    "node_modules/call-bind-apply-helpers": {
       "version": "1.0.2",
-      "resolved": "https://registry.npmjs.org/call-bind/-/call-bind-1.0.2.tgz",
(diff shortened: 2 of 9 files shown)


apps/react-sdk/src/transpiler/transpiler.ts
```diff
@@ -16,7 +16,8 @@ const ROOT_DIR = Path.resolve(__dirname, '../..');
  * @param options any extra options that should be passed.
  */
 export async function transpileFiles(directory: string, outputDir: string, options?: TranspileFilesOptions) {
-    const { files, dirs } = await getStatsInDir(directory);
+    let { files, dirs } = await getStatsInDir(directory);
+    files = files.filter(f => !f.endsWith("asyncapi-ui.min.js"));
```
derberg: `asyncapi-ui.min.js` is very html-template specific. Ideally we should not do changes like this here in sdk. 

best would be to suppress any warning by default and make them visible in the generator logs only if `debug` flag is passed

lightning-sagar:
@derberg ,
i noticed Babel prints warnings directly to stderr, so Rollup doesnot catch them, to follow your suggestion, i modified the transpiler so:
- babel warning are suppressed by default and
- they only appear when the generator is run with --debug flag

is this the right approach or would you prefer a different way of handling these warnings??

derberg:
@lightning-sagar I have to say I'm not a fan of `process.argv.includes('--debug');`. If we want to support debug flag, that would have to be done in the way that we pass this flag info in the right way inside the react sdk, without process reading. 

in this case no need to support debug flag - this warning just have to be made silent as it ain't helping in any way. But it must be solved without poluting `transpileFiles` with details, especially that in future we might have more warnings that should be removed - so think about introducing a separate function that does warning removal, and there is a clean way to pass "strings" that identify these errors, so it is also easily readable from code what we "silent"

lightning-sagar:
@derberg sorry for late resposnse again..
btw i moved the `stderr warning filtering` into a separate `createStderrFilter` file and removed the debug flag logic....
do let me know if this approach is ok?

apps/react-sdk/src/transpiler/transpiler.ts
```diff
@@ -17,48 +18,54 @@ const ROOT_DIR = Path.resolve(__dirname, '../..');
  */
 export async function transpileFiles(directory: string, outputDir: string, options?: TranspileFilesOptions) {
     const { files, dirs } = await getStatsInDir(directory);
-    if (files.length) {
+    const stderrFilter = createStderrFilter(getSilencedWarnings());
+    stderrFilter.apply();
+
+    try {
+        if (files.length) {
```
derberg, line 25: why this if must be inside try? why not have it as it was, before

could even be `if (!files?.length) return;`

derberg (commented):
can you explain why this is the only solution?

the problem we're talking about is the size of file, that is controlec by `compact` config in babel. Did you test how it would work if in babel plugin config you would add `compact: false`

https://babeljs.io/docs/options#compact

```
cwd: ROOT_DIR,
compact: false
babelHelpers: "bundled",
```

apps/react-sdk/src/transpiler/transpiler.ts
lightning-sagar: yes, i think we can that do that too...

lightning-sagar:
hi @derberg this issue is already fixed in the latest version of the generator...
earlier, in html-template, we were using
["generator": ">=2.0.0 <3.0.0"](https://github.com/asyncapi/html-template/commit/98281ae90e764d8b28753f657dc2c86e76218fa9#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519L62), which is why the Babel warning was showing up, but with the recent change to `"generator": ">=2.0.0 <4.0.0"`, the warning is no longer appear....
so I dont think we need to go further with `Babel-specific` handling here

also, I did try the `compact: false` approach earlier, but it didnt resolve the issue at that time, which is why I switched to the current approach

what do you think, does it make sense to stop here since the root cause is resolved now??

derberg:
interesting, there were no changes to babel version that could potentially fix it

ok, but if already fixed, let's close it

Issue #1712, opened 2025-09-02
Proper Error handling mechanism for Baked-in Templates


Adi-204:
We have @asyncapi/generator-components which is used in template development but currently we don't support all languages and protocol. In implementation their is no error handling and we need to solve this problem as provide accurate error messages to users is crucial.

batchu5:
I would to work on this issue, can you guide me through it

batchu5:
@Adi-204

Adi-204:
@varshitha-besthu this issue requires thorough discussion, research, and a deep understanding of template development. I’d recommend picking another issue for now. However, if you still wish to work on this one, please first share your proposed approach for feedback, and proceed with creating a PR only after the maintainers agree that the approach is suitable.

VIVEKDHEDHI021:
I’m really interested in this issue and would love to contribute to improving the error-handling experience. I believe solving this    will make the generator more reliable and much easier for developers to work with 


Pull Request: Improve Error Handling for Unsupported Languages & Protocols

 Summary
This PR improves the error handling inside @asyncapi/generator-components.
Currently, when a template uses an unsupported language or protocol, the generator either fails silently or throws unhelpful errors. This PR adds clear, user-friendly error messages to make the development experience smoother.

 Problem
- Unsupported languages and protocols do not show clear errors
- The generator sometimes fails silently or with confusing stack traces
- Developers struggle to understand what went wrong
- Missing default config values are not validated properly

This leads to poor debugging, inconsistent behavior, and frustration for template developers.

 Solution
This PR introduces:
- Validation checks for supported languages and protocols
- Clear, human-friendly error messages instead of silent failures
- Proper try/catch blocks to avoid swallowed errors
- Config loader improvements so missing defaults show explicit warnings
- More reliable feedback during template development

These updates make it easier to identify what went wrong and guide the user toward a fix.

How to Test
1. Run generator with an unsupported language (e.g., rust if not supported).
2. You should now see a clear error like:
   “Unsupported language: rust. Please choose a supported language.”
3. Use a template with a missing default config value.
4. Confirm that a human-friendly error explains exactly which field is missing.
5. Run a valid setup to ensure normal behavior is unchanged.

 Related Issue
Fixes: “Lack of error handling in @asyncapi/generator-components when unsupported languages/protocols are used.”

 Notes
If needed, more improvements can be added to extend protocol/language support.

Jatin24062005:
[...]

PR #1790, opened 2025-12-13
refactor(readme): remove language-specific hardcoding from README generation

### LINKED ISSUE #1787: Refactor components to not be hardcoded to one language
This is a follow-up issue for #1725. The `AvailableOperations` and `Installation` don't need to be hardcoded to just one language. As it is currently only applicable to JavaScript

lightning-sagar:
**Description**

- Refactored `Installation` to be language-aware instead of language-locked
- Render `AvailableOperations` based on the presence of AsyncAPI operations,
  not on a specific language

**Related issue(s)**

- Resolves #1787 

## Summary by CodeRabbit

* **New Features**
  * Installation section always shown and accepts a language selection, displaying the appropriate install command.
  * Code examples now include per-language variants (e.g., JavaScript and Python) with language-specific payload formatting.

* **Refactor**
  * Available operations list is rendered only when one or more operations exist.


packages/components/src/components/readme/Installation.js
```diff
@@ -1,14 +1,20 @@
 import { Text } from '@asyncapi/generator-react-sdk';
 
-export function Installation() {
+const installCommands = {
+  python: 'pip install -r requirements.txt',
+  javascript: 'npm install',
+};
+
+export function Installation({ language }) {
+  const command = installCommands[language] || installCommands['python']; // default to python command
```
lightning-sagar: @derberg @thulieblack 
is it okay to default to the python installation command when language is undefined or unsupported??
or it would be safer to return null instead, to avoid showing potentially incorrect installation instructions for other languages??

packages/components/src/components/readme/AvailableOperations.js
```diff
@@ -10,7 +10,7 @@ export function AvailableOperations({ operations }) {
     <>
       <Text newLines={2}>### Available Operations</Text>
       {operations.map((operation) => (
-        <Text newLines={2}>
+        <Text key={operation.id()} newLines={2}>
```
Adi-204: `<Text/>` is coming form `@asyncapi/generator-react-sdk` and we don't need key property it is just a warning.

packages/components/src/components/readme/Installation.js
Adi-204: @lightning-sagar for now we don't need to add a default we have a issue #1712 to introduce proper handling, so you can remove it.

packages/components/src/components/readme/AvailableOperations.js
```diff
@@ -10,7 +10,7 @@ export function AvailableOperations({ operations }) {
...
           <OperationHeader operation={operation} />
           <MessageExamples operation={operation} />
         </Text>
```
Adi-204, line 16: `<AvailableOperations />` is using two other components OperationHeader and MessageExamples. The  OperationHeader seems to be lang independent so NO changes in that component. But in MessageExamples I can see it renders example code snippets in Readme so currently it only has js add python also.

packages/components/src/components/readme/MessageExamples.js
```diff
@@ -1,12 +1,28 @@
...
+  },
+  python: {
+    label: 'Python',
+    codeBlock: 'python',
+    render: (operationId, payload) => `
+**Example (Python):**
+\`\`\`python
+client.${operationId}(${JSON.stringify(payload, null, 2)})
+\`\`\`
+`
```
Adi-204: For python you can look at generated clients and the methods name is in snakecase but in your implementation it is camelCase which is used in js. So you can use from `@asyncapi/generator-helpers` https://github.com/asyncapi/generator/blob/master/packages/helpers/src/utils.js#L70 to convert operationId.

packages/components/src/components/readme/MessageExamples.js
```diff
@@ -17,7 +33,9 @@ export default function MessageExamples({ operation }) {
     const examples = getMessageExamples(message) || [];
     examples.forEach((example) => {
       const payload = example.payload();
-      messageExamples.push(exampleTemplate(operationId, payload));
+      Object.values(languageConfig).forEach(({ render }) => {
+        messageExamples.push(render(operationId, payload));
```
Adi-204: you are only using render function from the lang config than why you label and codeBlock properties for every lang in config?

packages/components/src/components/readme/MessageExamples.js
lightning-sagar: ohk, thanks for pointing this out, and sorry for the late response, I was at a hackathon for 2 days and then traveling for 3 days more....

i got the point... i will update the python example to convert the operationId to snake_case using toSnakeCase. so it matches the generated python client methods....

packages/components/src/components/readme/MessageExamples.js
lightning-sagar: sorry my bad.. i will update this to use `label and codeBlock` as well. if you would prefer having only `render` in the config do let me know and i will update that accordingly....

packages/components/src/components/readme/MessageExamples.js
```diff
@@ -1,12 +1,30 @@
-import { getMessageExamples, getOperationMessages } from '@asyncapi/generator-helpers';
+import { getMessageExamples, getOperationMessages,toSnakeCase } from '@asyncapi/generator-helpers';
```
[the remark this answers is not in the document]
Adi-204: @lightning-sagar this is valid.

packages/components/src/components/readme/MessageExamples.js
```diff
@@ -1,12 +1,30 @@
...
+      ? toSnakeCase(operationId)
+      : operationId;
+
+  return `
+**Example (${label}):**
+\`\`\`${codeBlock}
+client.${opId}(${JSON.stringify(payload, null, 2)})
 \`\`\`
 `;
+}
```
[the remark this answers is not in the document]
Adi-204: @lightning-sagar this is also valid.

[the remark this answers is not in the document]
lightning-sagar: hi @Adi-204 , sorry but just to confirm do you mean the logic is correct and the current implementation is fine... so I can mark the coderabbit conversation as resolved??

Adi-204:
@lightning-sagar Did you go through both CodeRabbit comments carefully?
The first one is a straightforward formatting improvement. The second one is not related to logic at all—it only suggests adding JSDoc comments and even provides guidance on what can be used.
Read [this ](https://github.com/asyncapi/generator/pull/1790#issuecomment-3649639430) points carefully especially point 10 so that PR review is much easier.

lightning-sagar:
> @lightning-sagar Did you go through both CodeRabbit comments carefully? The first one is a straightforward formatting improvement. The second one is not related to logic at all—it only suggests adding JSDoc comments and even provides guidance on what can be used. Read [this ](https://github.com/asyncapi/generator/pull/1790#issuecomment-3649639430) points carefully especially point 10 so that PR review is much easier.

sorry about the confusion, I do read both coderabbit comments...,. but I confused, what you meant by “this is valid” that is why I asked for confirmation.... sorry about that

Adi-204:
@lightning-sagar oh ok I just mean that comment from coderabbit is valid and please do the changes suggested by coderabbit

PR #1814, opened 2026-01-01
feat: automatic routing of message for slack example using discriminators

### LINKED ISSUE #1548: explore how to take advantage of the fact that AsyncAPI document contains different handlers definitions
if we have AsyncAPI document, like Slack one, that has 3 operations, why not generate code that understands it? Meaning enable client user to write custom logic only for `hello` message, so generated code understands incomming message, understand it is hello message, so the dedicated handler is triggered only if hello message is sent. So the client user do not have to learn docs, identify messages and learn how to validate and react to them

Adi-204:
**Description**
This is for now a draft and more cleanup need to be done so that other examples are not affected but it is working for slack perfectly. 

Screenshot 2026-01-01 171738

**Related issue(s)**
Fixes #1548 

## Summary by CodeRabbit

* **New Features**
  * Automatic operation-based message routing for Python WebSocket clients; messages dispatch to operation-specific handlers via discriminators.
  * New registration methods to attach handlers per message type, with optional manual discriminator configuration.

* **Documentation**
  * Added guide for using auto-routing in WebSocket clients (includes usage and discriminator guidance).

* **Examples**
  * New Slack WebSocket client example demonstrating auto-routing.

* **Tests**
  * Added tests validating discriminator extraction and routing behavior.

### DIFF
--- a/.changeset/slack-auto-routing.md
+++ b/.changeset/slack-auto-routing.md
@@ -4,27 +4,47 @@
 "@asyncapi/generator-helpers": minor
 ---
 
-Enhanced Python webSocket client generation with **automatic operation-based message routing**:
+- **Updated Component**: `OnMessage` (Python) - Added discriminator-based routing logic that automatically dispatches messages to operation-specific handlers before falling back to generic handlers
+- **New Helpers**: 
+  - `getMessageDiscriminatorData` - Extracts discriminator key and value from individual messages
+  - `getMessageDiscriminatorsFromOperations` - Collects all discriminator metadata from receive operations
+- Enhanced Python webSocket client generation with **automatic operation-based message routing**:
 
-**How It Works:**
+## How python routing works
+
 - Generated WebSocket clients now automatically route incoming messages to operation-specific handlers based on message discriminators. Users can register handlers for specific message types without manually parsing or filtering messages.
 - When a message arrives, the client checks it against registered discriminators (e.g., `type: "hello"`, `type: "events_api"`)
 - If a match is found, the message is routed to the specific operation handler (e.g., `onHelloMessage`, `onEvent`)
 - If no match is found, the message falls back to generic message handlers
 - This enables clean separation of message handling logic based on message types
 
-**Design Rationale:**
+> `discriminator` is a `string` field that you can add to any AsyncAPI Schema. This also means that it is limited to AsyncAPI Schema only, and it won't work with other schema formats, like for example, Avro. 
+
 The implementation automatically derives discriminator information from your AsyncAPI document:
-- **Discriminator key** is extracted from the `discriminator` field in your AsyncAPI spec
-- **Discriminator value** is extracted from the `const` property defined in message schemas
+- Discriminator `key` is extracted from the `discriminator` field in your AsyncAPI spec
+- Discriminator `value` is extracted from the `const` property defined in message schemas
 
-**Important Limitations:**
-- When defaults aren't available in the AsyncAPI document, users must provide **both** `discriminator_key` and `discriminator_value` when registering handlers
-- Providing only one parameter is not supported - you must provide either both or neither
-- **Why this limitation exists**: When a receive operation has multiple messages sharing the same discriminator key (e.g., all use `"type"` field), we need the specific value (e.g., `"hello"`, `"disconnect"`) to distinguish between them. Without both pieces of information, the routing becomes ambiguous
-- AsyncAPI 3.0 currently supports discriminator values as strings only.
+Example AsyncAPI Schema with `discriminator` and `const`:
+```yaml
+schemas:
+    hello:
+      type: object
+      discriminator: type # you specify name of property
+      properties:
+        type:
+          type: string
+          const: hello # you specify the value of the discriminator property that is used for routing
+          description: A hello string confirming WebSocket connection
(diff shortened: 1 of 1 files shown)


aeworxet:
@asyncapi/bounty_team

Adi-204:
@derberg I’ve implemented the changes we discussed during the meeting. While working on them, I realized that the initial solution assumed only a single message per receive operation. Since a receive operation can contain multiple messages, I had to refactor the approach more extensively.

The updated design now uses `self.receive_operation_discriminators` in the constructor as an array of objects, where each object follows this structure:

```js
self.receive_operation_discriminators = [
  { "key": "type", "value": "hello", "operation_id": "onHelloMessage" },
  { "key": "type", "value": "events_api", "operation_id": "onEvent" },
  { "key": "type", "value": "disconnect", "operation_id": "onDisconnectMessage" }
]
```

With this implementation, the routing logic correctly supports multiple messages within the same receive operation, rather than assuming a one-to-one relationship.

### Limitation of the current approach

As discussed, by default we derive:

* `discriminator_key` from the discriminator field, and
* `discriminator_value` from the corresponding `const` defined in the `asyncapi-slack` document.

For non-default cases, the user must explicitly provide **both** `discriminator_key` and `discriminator_value`. Partial inputs are not supported—only `(0, 0)` or `(1, 1)` combinations are allowed. This is why I removed the `(0, 1)` and `(1, 0)` cases.

### Why we can’t support only one of `discriminator_key` or `discriminator_value`

Consider the `onHelloMessage` operation having three messages: A, B, and C.
In this scenario:

* All messages share the same `discriminator_key`
* The discriminator values are not available via `const` (i.e., no default)

If we allowed only `discriminator_key` to be provided, updating `self.receive_operation_discriminators` would result in multiple identical entries with the same key and `operation_id`. At that point, there would be no reliable way to determine which discriminator value corresponds to which message provided by the user.

Requiring both `discriminator_key` and `discriminator_value` avoids this ambiguity and keeps message routing deterministic..

packages/templates/clients/websocket/python/example-slack-with-routing.py
```diff
@@ -0,0 +1,41 @@
+import sys
```
derberg, line 1: looks great and clean, just please document it in https://github.com/asyncapi/generator/blob/master/packages/templates/clients/websocket/python/README.md like all other examples

packages/templates/clients/websocket/python/components/ReceiveOperationsDiscriminators.js
```diff
@@ -0,0 +1,78 @@
+import { Text } from '@asyncapi/generator-react-sdk';
+
+/**
+* Extracts discriminator metadata from a message for operation routing.
+* 
+* @param {object} message - The AsyncAPI message object containing payload and discriminator info.
+* @param {string} operationId - The operation ID associated with this message.
+* @returns {object|null} An object with key, value, and operation_id if discriminator is valid; otherwise null.
+*/
+const getDiscriminatorData = (message, operationId) => {
```
derberg: this is generic helper right? can be actually added to https://github.com/asyncapi/generator/tree/master/packages/helpers, tested and also released?

packages/templates/clients/websocket/python/components/AutoRouteMessages.js
```diff
@@ -0,0 +1,43 @@
+import { Text } from '@asyncapi/generator-react-sdk';
```
derberg: can you elaborate why better to have the auto route message component like this instead of just extending `OnMessage` component?

in the end it is default behavior that we are doing here imho, just extending from dummy generic implementation to descriminator support, wdyt?

packages/templates/clients/websocket/python/components/ReceiveOperationsDiscriminators.js
```diff
@@ -0,0 +1,78 @@
...
+};
+
+/**
+* Generates Python initialization code for receive operation discriminators.
+* 
+* @param {object} props - Component props.
+* @param {Array} props.receiveOperations - Array of receive operations from AsyncAPI document.
+* @returns {React.Element|null} A Text component rendering Python code, or null if no operations.
+*/
+export function ReceiveOperationsDiscriminators({ receiveOperations }) {
```
derberg, line 11: ideally all custom logic to extract info from asyncapi should not be here but in generic helpers as in the end any template will need this if they want discriminator support

Adi-204:
@derberg I did some refactoring as per your suggestions the only thing I didn't really get is are we also releasing minor version of generator? And in changeset I need to explain the approach with limitations?

derberg:
@Adi-204 exactly, 
- limitations need to be visible in release notes - release notes source is text from changeset
- generator needs minor release every time we change some login in baked-in templates, otherwise they will not be available to users

Adi-204:
@derberg updated changeset and also run other example 👍 should I change `OperationsDiscriminators.js` back and remove from components if you feel it's not worth adding it?

derberg:
yes please, move `OperationsDiscriminators.js` to python directly, related helpers stay in helpers lib

Adi-204:
@derberg revert changes! please read changeset, I tried to explain in simple words and summarize way if any changes need.

PR #1821, opened 2026-01-07
feat: proper error handling for @async-api/generator-components

### LINKED ISSUE #1712: Proper Error handling mechanism for Baked-in Templates

Jatin24062005:
## **Description**

This PR introduces **centralized and structured error handling** for language and framework resolution in `@asyncapi/generator-components`.

### What was the problem?

* Language/framework resolution logic was duplicated across multiple components.
* Unsupported languages or missing/unsupported frameworks resulted in **silent failures** or unclear runtime errors (e.g. `undefined is not a function`).
* Error messages were inconsistent and difficult for users to understand.

### What this PR does

* Introduces a **single centralized resolver** (`resolveLanguageConfig`) to handle:

  * Unsupported languages
  * Missing frameworks (only when required)
  * Unsupported frameworks
  * Invalid configuration shapes
* Adds a **shared `GeneratorError`** class to provide:

  * Clear, user-facing error messages
  * Consistent error codes and metadata
* Refactors `MethodGenerator` and related components to rely on the centralized resolver instead of local logic.
* Adds **unit tests for the resolver** covering all supported configuration patterns:

  * Function-based configs
  * Direct language configs
  * Framework-based configs (e.g. Java/Quarkus)

### Why this approach

* Keeps individual components **declarative and DRY**
* Ensures **consistent behavior and error messages** across all templates
* Makes it easier to extend support for new languages/frameworks in the future
* Improves developer and user experience without changing template APIs

---

## **Related issue(s)**

* Fixes #1712 

## Summary by CodeRabbit

* **Refactor**
  * Consolidated language and framework configuration resolution into a single resolver to make code generation behavior more consistent and maintainable.

* **Tests**
  * Added comprehensive tests for configuration resolution, covering valid shapes and error cases for unsupported or missing language/framework combinations.

### DIFF
--- a/packages/components/test/components/resolveLanguageConfig.test.js
+++ b/packages/components/test/components/resolveLanguageConfig.test.js
@@ -16,7 +16,7 @@ describe('resolveLanguageConfig', () => {
           language: 'python',
           context
         })
-      ).toThrow(GeneratorError);
+      ).toThrow(ComponentConfigError);
 
       expect(() =>
         resolveLanguageConfig({
@@ -91,7 +91,7 @@ describe('resolveLanguageConfig', () => {
           language: 'java',
           context
         })
-      ).toThrow(GeneratorError);
+      ).toThrow(ComponentConfigError);
 
       expect(() =>
         resolveLanguageConfig({
@@ -110,7 +110,7 @@ describe('resolveLanguageConfig', () => {
           framework: 'spring',
           context
         })
-      ).toThrow(GeneratorError);
+      ).toThrow(ComponentConfigError);
 
       expect(() =>
         resolveLanguageConfig({
@@ -135,7 +135,7 @@ describe('resolveLanguageConfig', () => {
           language: 'javascript',
           context
         })
-      ).toThrow(GeneratorError);
+      ).toThrow(ComponentConfigError);
 
       expect(() =>
         resolveLanguageConfig({


Jatin24062005:
@Adi-204 No, for now i don't think so it will grow I introduced the errors folder mainly to make error definitions an explicit concern, instead of mixing them into resolution or generator logic. I can remove it and move `GeneratorError` defination inside the  resolveLanguageConfig.js
 
https://github.com/asyncapi/generator/blob/master/apps/generator/lib/logMessages.js This approaches won't align properly because The resolver is a control-flow boundary, not a logging concern.
Unlike logMessages.js, these failures must stop generation and be distinguishable programmatically (unsupported language vs missing framework vs invalid config).

This logic is dynamic and conditional — the failure depends on runtime shape checks and user input, not static situations — which makes a string-based helper insufficient.

A string-based approach would still require throwing generic Errors, force tests to rely on message matching, and make it difficult to attach or evolve structured context (e.g. supported languages/frameworks).
GeneratorError is intentionally minimal and exists to represent typed, fatal generator failures with stable semantics, not just human-readable messages.

Apologies for making PR first rather than Explaining my approach. To avoid extending the native Error class let me remove this folder !.

Adi-204:
> A string-based approach would still require throwing generic Errors, force tests to rely on message matching, and make it difficult to attach or evolve structured context (e.g. supported languages/frameworks). GeneratorError is intentionally minimal and exists to represent typed, fatal generator failures with stable semantics, not just human-readable messages.

@Jatin24062005  Well I don't think you looked at the file I shared carefully. In that file, there are some hard-coded logging messages, but below them there are dynamic functions that generate log messages based on input.
The reason I shared that example is because our current implementation works in a very similar way: calling `throw new GeneratorError(...)` with parameters triggers logic inside the constructor. Conceptually, the same behavior could be achieved using a function.
This is the core of my question: if we introduce a class-based error and only rely on its constructor by repeatedly calling `throw new GeneratorError(...)`, I don’t see a clear advantage of using a class here. The same outcome could be achieved with simple functions returning errors.
It would be helpful if you could explain, how this class-based approach provides additional value or flexibility compared to a function-based implementation.

Jatin24062005:
@Adi-204 A class gives us a stable error identity (instanceof GeneratorError) that cleanly separates intentional generator failures from generic runtime errors we can use them on other generic function to verify are expected vs unexpected error.
This is difficult to guarantee with a function without re-implementing type signaling manually.
While today we only rely on the constructor, the class provides a semantic boundary and allows future behavior (formatting, exit codes, CLI handling) to evolve without changing call sites.
That said, the abstraction is intentionally minimal, and I’m open to a function-based approach if the project prefers simplicity over explicit typing.

Adi-204:
@Jatin24062005 I can see the value of using a class-based approach as the library grows and more components are added, so let’s go with it.
I would recommend starting with a single file that contains the error class. If, in the future, we genuinely feel the need for multiple error classes, we can move them into separate files within a dedicated folder.
Also, I’d suggest not changing the code too rapidly in the PR. The questions I raised were intended to understand your reasoning behind the chosen structure—it may very well be the right approach.
Lastly, let’s avoid over-engineering. Please try to keep the implementation simple, readable, and consistent with the existing codebase.
I haven’t done a detailed review yet, but one immediate concern is the `GeneratorError `class name. This package is related to components, not the generator itself, so the name feels misleading. Please use a more meaningful and accurate name, and ensure that error messages are descriptive and easy to understand.

Jatin24062005:
@Adi-204 Thanks, Adi — that makes sense, and I appreciate the clarification.
Good point on the naming as well — I’ll rename GeneratorError to something more accurate for this package (e.g. ComponentConfigError / ComponentResolutionError) and the error messages  already remain clear and descriptive.

I’ll make these adjustments before the detailed review.

Adi-204:
@Jatin24062005, we discussed this PR in the last generator meeting. You can refer to the meeting recording in the #generator Slack channel.
I agree that a class-based error handling approach can make sense. However, in the current PR there is too much abstraction, and the design seems to be driven by only two components. While we do follow DRY principles, that doesn’t mean we should introduce complexity that makes the code harder to understand and maintain—especially in an open-source project. At the moment, it’s not clear how many error classes we will eventually need, what the overall structure should look like, or how maintainers are expected to scale and maintain this setup.

My recommendations:
1. You can experiment with these changes locally, but please avoid creating PRs directly. As mentioned earlier, this is not a suitable “first issue” (unlike documentation updates, bug fixes, or tests). Please be patient and take time to properly research and understand the problem before proposing a solution.
2. First, write a detailed proposal in an issue explaining the complete approach you plan to implement (design, structure, and reasoning). If the approach is approved by one or more maintainers, then you can open a draft PR, and we can start the review process from there.

I hope this clarifies the next steps. I’m closing this PR for now.

Jatin24062005:
@Adi-204 Thanks for the clarification and for discussing this in the generator meeting.
I understand the concern around introducing architectural abstractions without prior design alignment, and I appreciate the guidance on process.
I’ll take some time to properly research and write a detailed proposal as an issue, covering the design, structure, and trade-offs, before proposing any further changes.
Thanks for the feedback and direction.

PR #1866, opened 2026-01-27
test: add test coverage for OnTextMessageHandler

### LINKED ISSUE #1862: [TEST] Add component tests for OnTextMessageHandler.js (Java WebSocket Template)
## Description

The Java WebSocket template's `OnTextMessageHandler.js` component contains complex conditional logic but has no unit tests, violating the project's testing principle (point 8 of [assumptions](https://github.com/asyncapi/generator/tree/master/packages#assumptions-and-principles)):

> "Only template components with conditions require separate testing in their respective directories."

## Component Location
`packages/templates/clients/websocket/java/quarkus/components/OnTextMessageHandler.js`

Harsh16gupta:
**Description**
Adds unit tests for the OnTextMessageHandler component in the Java Quarkus WebSocket client template. The component generates routing logic for incoming text messages and previously had no test coverage.

**Changes**

- Added test suite and snapshots for OnTextMessageHandler
- Updated package.json to run Jest with coverage ("test": "jest --coverage")

image

**Related issue(s)**
Fixes #1862

## Summary by CodeRabbit

* **Tests**
  * Added integration tests for the WebSocket message handler covering multiple send-operation scenarios
  * Added an AsyncAPI-based test fixture for WebSocket component validation

* **Chores**
  * Adjusted test script behavior so test runs no longer pass automatically when no tests are present

### DIFF
--- a/apps/generator/CHANGELOG.md
+++ b/apps/generator/CHANGELOG.md
@@ -1,5 +1,11 @@
 # @asyncapi/generator
 
+## 3.1.2
+
+### Patch Changes
+
+- 4a09f57: Bump @asyncapi/parser to 3.6.0 to support AsyncAPI 3.1.0
+
 ## 3.1.1
 
 ### Patch Changes
--- a/apps/generator/package.json
+++ b/apps/generator/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@asyncapi/generator",
-  "version": "3.1.1",
+  "version": "3.1.2",
   "description": "The AsyncAPI generator. It can generate documentation, code, anything!",
   "main": "./lib/generator.js",
   "engines": {
@@ -53,7 +53,7 @@
     "@asyncapi/generator-hooks": "*",
     "@asyncapi/generator-react-sdk": "*",
     "@asyncapi/multi-parser": "^2.2.0",
-    "@asyncapi/parser": "^3.4.0",
+    "@asyncapi/parser": "^3.6.0",
     "@npmcli/arborist": "5.6.3",
     "@npmcli/config": "^8.0.2",
     "ajv": "^8.12.0",
--- a/package-lock.json
+++ b/package-lock.json
@@ -21,7 +21,7 @@
     },
     "apps/generator": {
       "name": "@asyncapi/generator",
-      "version": "3.0.0",
+      "version": "3.1.1",
       "license": "Apache-2.0",
       "dependencies": {
         "@asyncapi/generator-components": "*",
@@ -29,7 +29,7 @@
         "@asyncapi/generator-hooks": "*",
         "@asyncapi/generator-react-sdk": "*",
         "@asyncapi/multi-parser": "^2.2.0",
-        "@asyncapi/parser": "^3.4.0",
+        "@asyncapi/parser": "^3.6.0",
         "@npmcli/arborist": "5.6.3",
         "@npmcli/config": "^8.0.2",
         "ajv": "^8.12.0",
@@ -2039,10 +2039,12 @@
       }
     },
     "node_modules/@asyncapi/parser": {
-      "version": "3.4.0",
+      "version": "3.6.0",
+      "resolved": "https://registry.npmjs.org/@asyncapi/parser/-/parser-3.6.0.tgz",
+      "integrity": "sha512-6S0Yr8vI418a1IrpGsOYbfWVo9+aHvSqN2oSkiY0YJltS/C7oDOt9e0mo6hSld8bg+EeKrtgkVmpW4obh1JFvA==",
       "license": "Apache-2.0",
       "dependencies": {
-        "@asyncapi/specs": "^6.8.0",
+        "@asyncapi/specs": "^6.11.1",
         "@openapi-contrib/openapi-schema-to-json-schema": "~3.2.0",
         "@stoplight/json": "3.21.0",
         "@stoplight/json-ref-readers": "^1.2.2",
@@ -2058,8 +2060,8 @@
         "ajv-errors": "^3.0.0",
         "ajv-formats": "^2.1.1",
         "avsc": "^5.7.5",
-        "js-yaml": "^4.1.0",
-        "jsonpath-plus": "^10.0.0",
+        "js-yaml": "^4.1.1",
+        "jsonpath-plus": "^10.0.7",
         "node-fetch": "2.6.7"
       }
     },
@@ -2120,7 +2122,9 @@
       }
     },
     "node_modules/@asyncapi/specs": {
-      "version": "6.8.1",
+      "version": "6.11.1",
+      "resolved": "https://registry.npmjs.org/@asyncapi/specs/-/specs-6.11.1.tgz",
+      "integrity": "sha512-A3WBLqAKGoJ2+6FWFtpjBlCQ1oFCcs4GxF7zsIGvNqp/klGUHjlA3aAcZ9XMMpLGE8zPeYDz2x9FmO6DSuKraQ==",
       "license": "Apache-2.0",
       "dependencies": {
         "@types/json-schema": "^7.0.11"
@@ -17774,7 +17778,7 @@
     },
     "packages/components": {
       "name": "@asyncapi/generator-components",
-      "version": "0.4.0",
+      "version": "0.5.0",
       "license": "Apache-2.0",
       "dependencies": {
         "@asyncapi/generator-helpers": "*",
@@ -17792,7 +17796,7 @@
     },
     "packages/helpers": {
       "name": "@asyncapi/generator-helpers",
-      "version": "1.0.0",
+      "version": "1.1.0",
       "license": "Apache-2.0",
       "devDependencies": {
         "jest": "^27.3.1"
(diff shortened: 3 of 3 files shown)


Harsh16gupta:
Maintainers: I followed the component testing approach used in the Python template.

Harsh16gupta:
Nothing breaks:
image
image
It is showing a lot of details as it has only one file currently.

Harsh16gupta:
When I run npm test, I see the following React warning. I initially ignored it since it is already under discussion:
image

However, while working on another PR, I encountered the same warning again:
image

So I checked how other templates handle this.
**What I found**

- The Python template does not show this warning (still figuring why )
- Python’s RegisterReceiveOperations.js includes keys, although that component is currently not covered by tests
- The Kafka Java template (EndpointFields.js) also includes keys
- The Java WebSocket (Quarkus) components do not include keys, which is why the warning appears during tests

**Question**
Should I add key props to the Java components as well? (I know it is under discussion)

- Tests pass with 100% coverage either way
- This would only remove a warning and would not affect functionality

Please guide me on the preferred approach.
Thank you 🙏

Adi-204:
@Harsh16gupta don't worry about the key we have separate issue, so no need to add key.

packages/templates/clients/websocket/java/quarkus/test/components/OnTextMessageHandler.test.js
```diff
@@ -0,0 +1,47 @@
+import path from 'path';
+import { render } from '@asyncapi/generator-react-sdk';
+import { Parser, fromFile } from '@asyncapi/parser';
+import OnTextMessageHandler from '../../components/OnTextMessageHandler.js';
+
+const parser = new Parser();
+const asyncapiFilePath = path.resolve(__dirname, '../../../../../../../helpers/test/__fixtures__/asyncapi-websocket-query.yml');
```
Adi-204: @derberg I noticed this now in all other components also we are using fixture from `generator-helpers` and I don't really like it. As when someone is testing new helpers and makes changes in the fixture it might affect other components snapshots. wdyt 🤔 ? 

@Harsh16gupta what do you think? you did it correct just like other components but I'm thinking we might need to change fixture used in other components also. feel free to drop out opinion.

Harsh16gupta: @Adi-204 Yeah that makes sense, we can switch to a local fixture with just the test cases we need for these components. That way we have full control and won't be affected by helper fixture changes.
I am ready to update to a local fixture in this PR (after your confirmation).

Adi-204: @Harsh16gupta but are you thinking to introduce new fixtures? where exactly?

Harsh16gupta: @Adi-204 Yes, planning to create a local fixture at:
`packages/templates/clients/websocket/java/quarkus/test/__fixtures__/asyncapi-java-quarkus-tests.yml`

This would contain minimal test data needed specifically for Java Quarkus component tests just operations with different naming conventions and query parameters.The tests would then use this local fixture instead of the shared one from `generator-helpers`.

I even tried it locally it works fine and also we can do the same things for all the backed in template.
what do you think?

Harsh16gupta: @Adi-204 
I’ve added a local fixture at `packages/templates/clients/websocket/java/quarkus/test/__fixtures__/asyncapi-java-quarkus-tests.yml` and updated the tests to use it, so they’re isolated from generator-helpers fixture changes.
I have pushed it so that you can have a look.
Happy to adjust if you’d prefer a different structure.

Adi-204: @Harsh16gupta remember it is better to discuss before implementation. We had this initially that's why we created https://github.com/asyncapi/generator/tree/master/packages/templates/clients/websocket/test/__fixtures__ so that all websocket template fixtures at one place and redundant. wdyt?
My suggestion move the fixture name it something better and additionally try to keep fixture minimal as when more components come in it will grow.

Harsh16gupta: @Adi-204 I was quite impatient, will keep this in mind in future.

Harsh16gupta:
@Adi-204 I followed your suggestion, have a look at the name `asyncapi-java-quarkus-operations.yml` , I’m happy to rename it to something more generic if you prefer.

Harsh16gupta:
@Adi-204 also can I work on #1867 , #1885 and #1886
I have already covered all the cases for `URIParams.js`(#1867) locally.
image

Harsh16gupta:
@Adi-204 also can I start working on #1867 , #1887 and #1888
I have already covered all the cases for `URIParams.js`(#1867) locally.
image

Adi-204:
> @Adi-204 also can I start working on #1867 , #1887 and #1888 I have already covered all the cases for `URIParams.js`(#1867) locally. image

@Harsh16gupta I checked all the issues they are valid, but as told earlier this is very imp PR as we are making changes in component testing approach. So focus on this one and once it is merge rest would be much faster and easier to merge.

Harsh16gupta:
@Adi-204 
If we use `asyncapi-websocket-components.yml` as a shared fixture across all WebSocket templates (Java, Python, Dart, JavaScript), won't future changes to this fixture require snapshot updates in all those templates?
I initially thought we'd create separate fixtures per template to avoid this coupling. Could you clarify the long-term maintenance strategy?

Adi-204:
> @Adi-204 If we use `asyncapi-websocket-components.yml` as a shared fixture across all WebSocket templates (Java, Python, Dart, JavaScript), won't future changes to this fixture require snapshot updates in all those templates? I initially thought we'd create separate fixtures per template to avoid this coupling. Could you clarify the long-term maintenance strategy?

@Harsh16gupta nice question! So because we are experimenting things in `packages` folder and we are also not having too many components per websocket templates client my guess is it might worth adding only 1 asyncapi document to test all the components for now and at point we would have that document extensive like slack-example and no need for anyone to add something new for testing new added component. But as I mention we can just try it out and if in future we might feel that this template components are completely different we can add separate asyncapi document for that template. wdyt 🤔 ?

Adi-204:
so basically originally we were using the asyncapi doc from helpers but I think that is too much coupling as helpers are completely different and doesn't make sense so I think let's keep the assumption of having one file for component testing valid, but let's have it separate and not use the helpers one. I hope it answers your question feel free to drop your opinion.

Harsh16gupta:
@Adi-204 Thanks for the detailed explanation , this approach makes sense. Starting with one shared fixture and growing it incrementally feels right. We don’t have enough components yet to justify separation, and this also avoids coupling with helper fixtures.
I’ll rename it to `asyncapi-websocket-components.yml` and extend it as more component tests are added.

Harsh16gupta:
@Adi-204 please have a look. Happy to tweak the description further if you’d prefer something more detailed or different in tone.

Adi-204:
@Harsh16gupta just a question in asyncapi document you add why all the send operations have different casing? what is the purpose?

Harsh16gupta:
@Adi-204 The different casings are there to test the `toCamelCase` conversion. I wanted to make sure the component handles all the common naming styles developers might use:

- `sendMessage` - already camelCase
- `SendNotification` - PascalCase  
- `send_status_update` - snake_case

All three should convert properly to valid Java method names.

I noticed the old gemini fixture had the same approach (`PascalCaseOperation`, `operation_with_snake_case`, etc.), so I kept that pattern when creating the fixture. 

**old gemini fixture:**
```yaml
PascalCaseOperation:
  action: receive
  channel:
    $ref: '#/channels/marketDataV1'
  messages:
    - $ref: '#/channels/marketDataV1/messages/noExamples'
  
operation_with_snake_case:
  action: send
  channel:
    $ref: '#/channels/marketDataV1'
  messages:
    - $ref: '#/channels/marketDataV1/messages/noExamples'
```

Adi-204:
@Harsh16gupta yeah man but it was for testing helpers `toSnakeCase` and  `toCamelCase` which are tested there so I don't get point of again testing in components. If you notice template development style we are using helpers in different components. And if we are already testing those helpers separately, there is no need to try to test them again in component. I would recommend please change all to camel case.

Harsh16gupta:
@Adi-204 I did not analysed helpers package correctly, now got what you are implying to tell .
Will go through it thoroughly and will also review my PR, after that I will ping you for a final review.

Harsh16gupta:
@Adi-204 I removed one operation (only need 2 for testing OnTextMessageHandler.js) and changed all names to camelCase. Got what you were trying to convey.

The helpers package already tests toCamelCase with different casings:
```js
describe('toCamelCase integration test with AsyncAPI', () => {
  it('should convert snake_case operation names to camelCase format', () => {
    const operation = ops.get('operation_with_snake_case');
    const actualOperationId = toCamelCase(operation.id());
    expect(actualOperationId).toBe('operationWithSnakeCase');
  });
  it('should convert PascalCase operation names to camelCase format', () => {
    const operation = ops.get('PascalCaseOperation');
    const actualOperationId = toCamelCase(operation.id());
    expect(actualOperationId).toBe('pascalCaseOperation');
  });
});
```
Will be more careful in the future.

Adi-204 (approved):
lgtm!

PR #1890, opened 2026-01-30
docs: update the correct path for JS client in WebSocket client README

lightning-sagar:
**Description**

- Updated the correct path reference: ./test/temp/snapshotTestResult/client_postman/client.js
- Discussed in slack: https://asyncapi.slack.com/archives/C072JMTJ85Q/p1769694527217159

**Related issue(s)**

## Summary by CodeRabbit

* **Documentation**
  * Simplified and corrected example code paths in the WebSocket client template documentation for improved accuracy and usability.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

PR #1918, opened 2026-02-02
chore: remove async/await from websocket-client send operations and convert the test to use synchronous library

### LINKED ISSUE #1857: [BUG] Python Template - Async/Sync Mismatch Causes Runtime Errors
### Describe the bug.

**The template imports the synchronous websocket-client library:**

`import websocket`

However, the generated code declares async methods and uses await on websocket-client methods.

Example:

```
async def send_echo_message(self, message):
    await self._send(message, self.ws_app)

async def _send(message, socket):
    await socket.send(message)
```

websocket-client methods like socket.send() are synchronous and return None.

### Expected behavior

The generated Python WebSocket client should work without runtime errors.

Either the template should generate fully synchronous code when using websocket-client, or it should use an async-compatible WebSocket library if async/await is intended.

### Screenshots

I tried testing with 
```
import asyncio
from client import HoppscotchEchoWebSocketClient

async def test():
    client = HoppscotchEchoWebSocketClient()
    client.connect()
    await asyncio.sleep(1)
    await client.send_echo_message({"text": "hello"})

asyncio.run(test())
```
and got this error

Image

Note: Hoppscotch was down so I used Postman

### How to Reproduce

1. Look at the client.py file at `/home/harsh/projects/asyncapi/generator/packages/templates/clients/websocket/python/test/temp/snapshotTestResult/client_hoppscotch/client.py`
after running `npm test` in the integration test

2. The generated code contains async/await usage with a synchronous WebSocket client:
```
async def send_echo_message(self, message):
    """
    Send a send_echo_message message using the WebSocket connection attached to this instance.

    Args:
        message (dict or str): The message to send. Will be serialized to JSON if it's a dictionary.

    Raises:
        Exception: If sending fails or the socket is not connected.
    """
    await self._send(message, self.ws_app)
```

- Operating System (OS):Ubuntu 22.04
- Browser:
- Browser Version:

Harsh16gupta:
**Description**

Fixes async/sync mismatch in the Python WebSocket client template. The template used `async/await` with the synchronous `websocket-client` library, causing runtime errors.

This PR:
- Removes `async/await` from send operations
- Convert test to use synchronous `websockets.sync` library

Fixes #1857

**Screenshots**
Templates
Python WebSocket Client
image
Client for Slack
image
Client for Slack with Auto-Routing
image

Acceptance test
image
image

## Summary by CodeRabbit

* **Refactor**
  * Python WebSocket send helpers and client send methods converted from asynchronous to synchronous execution while preserving behavior and error handling.
  * Message-discriminator logic simplified to a single conditional expression, removing an early-return guard.

* **Tests**
  * Acceptance tests converted from async/await to a synchronous, thread-driven server lifecycle; startup/shutdown and timing now use thread coordination and blocking waits, and expected test message flow updated.

### DIFF
--- a/.github/workflows/scripts/mailchimp/package-lock.json
+++ b/.github/workflows/scripts/mailchimp/package-lock.json
@@ -334,9 +334,9 @@
       "integrity": "sha512-3ouUOpQhtgrbOa17J7+uxOTpITYWaGP7/AhoR3+A+/1e9skrzelGi/dXzEYyvbxubEF6Wn2ypscTKiKJFFn1ag=="
     },
     "node_modules/qs": {
-      "version": "6.14.1",
-      "resolved": "https://registry.npmjs.org/qs/-/qs-6.14.1.tgz",
-      "integrity": "sha512-4EK3+xJl8Ts67nLYNwqw/dsFVnCf+qR7RgXSK9jEEm9unao3njwMDdmsdvoKBKHzxd7tCYz5e5M+SnMjdtXGQQ==",
+      "version": "6.14.2",
+      "resolved": "https://registry.npmjs.org/qs/-/qs-6.14.2.tgz",
+      "integrity": "sha512-V/yCWTTF7VJ9hIh18Ugr2zhJMP01MY7c5kh4J870L7imm6/DIzBsNLTXzMwUA3yZ5b/KBqLx8Kp3uRvd7xSe3Q==",
       "dependencies": {
         "side-channel": "^1.1.0"
       },
@@ -723,9 +723,9 @@
       "integrity": "sha512-3ouUOpQhtgrbOa17J7+uxOTpITYWaGP7/AhoR3+A+/1e9skrzelGi/dXzEYyvbxubEF6Wn2ypscTKiKJFFn1ag=="
     },
     "qs": {
-      "version": "6.14.1",
-      "resolved": "https://registry.npmjs.org/qs/-/qs-6.14.1.tgz",
-      "integrity": "sha512-4EK3+xJl8Ts67nLYNwqw/dsFVnCf+qR7RgXSK9jEEm9unao3njwMDdmsdvoKBKHzxd7tCYz5e5M+SnMjdtXGQQ==",
+      "version": "6.14.2",
+      "resolved": "https://registry.npmjs.org/qs/-/qs-6.14.2.tgz",
+      "integrity": "sha512-V/yCWTTF7VJ9hIh18Ugr2zhJMP01MY7c5kh4J870L7imm6/DIzBsNLTXzMwUA3yZ5b/KBqLx8Kp3uRvd7xSe3Q==",
       "requires": {
         "side-channel": "^1.1.0"
       }
--- a/apps/generator/docs/usage.md
+++ b/apps/generator/docs/usage.md
@@ -40,29 +40,29 @@ asyncapi generate fromTemplate asyncapi.yaml @asyncapi/html-template
 
 **The shortest possible syntax:**
 ```bash
-asyncapi generate fromTemplate asyncapi.yaml @asyncapi/html-template@3.0.0 --use-new-generator
+asyncapi generate fromTemplate asyncapi.yaml @asyncapi/html-template@3.5.4
 ```
 
 **Generating from a URL:**
 ```bash
-asyncapi generate fromTemplate https://bit.ly/asyncapi @asyncapi/html-template@3.0.0 --use-new-generator
+asyncapi generate fromTemplate https://bit.ly/asyncapi @asyncapi/html-template@3.5.4 
 ```
 
 **Specify where to put the result:**
 ```bash
-asyncapi generate fromTemplate asyncapi.yaml @asyncapi/html-template@3.0.0 --use-new-generator -o ./docs
+asyncapi generate fromTemplate asyncapi.yaml @asyncapi/html-template@3.5.4 -o ./docs
 ```
 
 **Passing parameters to templates:**
 ```bash
-asyncapi generate fromTemplate asyncapi.yaml @asyncapi/html-template@3.0.0 --use-new-generator -o ./docs -p title='Hello from param'
+asyncapi generate fromTemplate asyncapi.yaml @asyncapi/html-template@3.5.4  -o ./docs -p title='Hello from param'
 ```
 
-In the template you can use it like this: ` {{ params.title }}`
+In the template you can use it like this: `{{ params.title }}`
 
 **Disabling the hooks:**
 ```bash
-asyncapi generate fromTemplate asyncapi.yaml @asyncapi/html-template@3.0.0 --use-new-generator -o ./docs -d generate:before generate:after=foo,bar
+asyncapi generate fromTemplate asyncapi.yaml @asyncapi/html-template@3.5.4 -o ./docs -d generate:before generate:after=foo,bar
 ```
 
 The generator skips all hooks of the `generate:before` type and `foo`, `bar` hooks of the `generate:after` type.
@@ -81,7 +81,7 @@ asyncapi generate fromTemplate asyncapi.yaml https://github.com/asyncapi/html-te
 
 **Map schema references from baseUrl to local folder:**
 ```bash
-asyncapi generate fromTemplate test/docs/apiwithref.json @asyncapi/html-template@3.0.0 --use-new-generator -o ./build/ --force-write --map-base-url https://schema.example.com/crm/:./test/docs/
+asyncapi generate fromTemplate test/docs/apiwithref.json @asyncapi/html-template@3.5.4 -o ./build/ --force-write --map-base-url https://schema.example.com/crm/:./test/docs/
 ```
 
 The parameter `--map-base-url` maps external schema references to local folders.
@@ -104,7 +104,7 @@ docker run --rm -it \
    --user=root \
    -v ${PWD}/test/fixtures/asyncapi_v1.yml:/app/asyncapi.yml \
    -v ${PWD}/output:/app/output \
-   asyncapi/cli generate fromTemplate -o /app/output /app/asyncapi.yml @asyncapi/html-template@3.0.0 --use-new-generator --force-write
+   asyncapi/cli generate fromTemplate -o /app/output /app/asyncapi.yml @asyncapi/html-template@3.5.4 --force-write
 ```
 Note: Use ``` ` ``` instead of `\` for Windows.
 
@@ -115,7 +115,7 @@ Note: Use ``` ` ``` instead of `\` for Windows.
 Use the following npx command on your terminal:
 
 ```bash
-npx -p @asyncapi/cli asyncapi generate fromTemplate ./asyncapi.yaml @asyncapi/html-template@3.0.0 --use-new-generator
+npx -p @asyncapi/cli asyncapi generate fromTemplate ./asyncapi.yaml @asyncapi/html-template@3.5.4 
 ```
 
 ## Using as a module/package
--- a/apps/generator/docs/versioning.md
+++ b/apps/generator/docs/versioning.md
@@ -18,10 +18,9 @@ It is better to lock a specific version of the template and the generator if you
 Generate HTML with the latest AsyncAPI CLI using the html-template.
 ```
 npm install -g @asyncapi/cli
-asyncapi generate fromTemplate asyncapi.yaml @asyncapi/html-template@3.0.0 --use-new-generator
+asyncapi generate fromTemplate asyncapi.yaml @asyncapi/html-template@3.5.4
 ```
 
-> AsyncAPI CLI has multiple versions of the generator, and to use the latest version, you may need to pass the `--use-new-generator` flag. For more details you can also check [asyncapi generate fromTemplate ASYNCAPI TEMPLATE](https://www.asyncapi.com/docs/tools/cli/usage#asyncapi-generate-fromtemplate-asyncapi-template)
 
 Generate HTML using a particular version of the AsyncAPI CLI using the html-template.
 
--- a/apps/generator/lib/conditionalGeneration.js
+++ b/apps/generator/lib/conditionalGeneration.js
@@ -42,7 +42,8 @@ async function isGenerationConditionMet (
       return conditionalSubjectGeneration(
         asyncapiDocument,
         templateConfig,
-        matchedConditionPath
+        matchedConditionPath,
+        templateParams
       );
     }
     return conditionalParameterGeneration(templateConfig,matchedConditionPath,templateParams);
@@ -111,9 +112,7 @@ async function conditionalSubjectGeneration (
   const server = templateParams.server && asyncapiDocument.servers().get(templateParams.server);
   const source = jmespath.search({
     ...asyncapiDocument.json(),
-    ...{
-      server: server ? server.json() : undefined,
-    },
+    server: server ? server.json() : undefined,
   }, subject);
 
   if (!source) {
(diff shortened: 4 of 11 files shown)


Harsh16gupta:
@Adi-204 I tried to look at the failed test but can't figure out why it is failing, please guide. All executed checks (SonarCloud, CodeRabbit) are passing. Please let me know if you’d like me to rerun anything after approval.

Harsh16gupta:
> @Harsh16gupta acceptance test for python are failing please look into it and in the PR desc I was expecting screenshots of running clients with `python example.py` command and not the test results.

@Adi-204 thanks for the review, and sorry for the misunderstanding.
Initially the tests were failing, so I was updating the snapshots and added those screenshots to the description. I understand now that screenshots of the client running via python example.py are expected. I’ll update the PR accordingly and fix the failing Python acceptance tests.

Harsh16gupta:
@Adi-204 I've investigated the failing Python acceptance test and found.

### The Issue

The acceptance test file (`packages/templates/clients/websocket/test/python/test_acceptance.py`) has a mismatch between WebSocket libraries:

1. **The generated client** (after my PR fix): Uses the synchronous `websocket-client` library
2. **The test's WebSocket server**: Uses the asynchronous `websockets` library

In `test_hoppscotch_client_sends_message()`, the test tries to call:
```python
await HoppscotchEchoWebSocketClient.send_echo_message_static(expected_outgoing_message, websocket)
```

But:
- My PR correctly made `send_echo_message_static()` synchronous (removing `async`/`await`)
- The `websocket` parameter comes from the `websockets` library (async), not `websocket-client`
- These are two completely different libraries with incompatible APIs

### Proposed Solution

I see two approaches:

**Option 1: Send directly in the test (simpler)**
```python
async def handler(websocket):
    message = expected_outgoing_message
    if isinstance(message, dict):
        message = json.dumps(message)
    await websocket.send(message)  # Use websockets library directly
```
This replicates the client's message formatting logic without mixing libraries.

**Option 2: Refactor test to use websocket-client**
Make the test server also use `websocket-client`, but this requires more significant test infrastructure changes.

### Question

Which approach would you prefer?
Please let me know your preference, and I'll implement it right away!

Harsh16gupta:
@Adi-204 Since the last test-related PR has been merged, I wanted to check if I can go ahead and start working on #1887.

Adi-204:
hey @Harsh16gupta first let's wait for https://asyncapi.slack.com/archives/C072JMTJ85Q/p1770225233478619 
to be 100% sure and can you please explain Option 1 in more detail?

Harsh16gupta:
@Adi-204 Sorry for not updating the screenshots until now. I was using WSL, and the command
`pip install -r test/temp/snapshotTestResult/custom_client_hoppscotch/requirements.txt`
was not working there, so I was unable to run the snapshot tests at that time.

During the current debugging (after my PR fix),
image

I found that the Python generated client.py (for postman) does not include:

```
self.receive_operation_handlers = {}
self.receive_operation_discriminators = []
```
Because of this, the runtime crashes when on_message tries to access these attributes. To verify this, I manually added these lines to the generated client.py:

image

After adding them, the runtime error was resolved. This confirmed that the issue was caused by missing initialization in the generated client.

The main question then was why this is generated for Hoppscotch but not for Postman. While tracing the generator logic, in the file
`packages\templates\clients\websocket\python\components\ReceiveOperationsDiscriminators.js`
I found that this was due to conditional handling of receiveOperation so I changed it to:
before
```
const hasOperations = Array.isArray(receiveOperations) && receiveOperations.length > 0;
  if (!hasOperations) {
    return null;
  }

  const operationDiscriminators = getMessageDiscriminatorsFromOperations(receiveOperations);
```
after
```
const operationDiscriminators =
  Array.isArray(receiveOperations) && receiveOperations.length > 0
    ? getMessageDiscriminatorsFromOperations(receiveOperations)
    : [];
```
With this change, the constructor always initializes the required fields and uses an empty list when there are no receive operations.
I then ran everything again, and the correct client.py was generated.
Final output:

image

Waiting for your approval before committing all these changes.
Thank you.

Harsh16gupta:
> can you please explain Option 1 in more detail?

So the main problem for the failing of the test was that we are using `websockets` (Asynchronous) in the test server 
```
async def handler(websocket):
    await HoppscotchEchoWebSocketClient.send_echo_message_static(
        expected_outgoing_message, 
        websocket  # wrong library
    )
```
But send_echo_message_static() expects a socket from websocket-client library (sync)
```

@staticmethod
  def _send(message, socket):
      """
      Internal helper to handle the actual sending logic.

      Args:
          message (dict or str): The message to send.
          socket (websockets.WebSocketCommonProtocol): The WebSocket to send through.

      Notes:
          If message is a dictionary, it will be automatically converted to JSON.
      """
      try:
          if isinstance(message, dict):
              message = json.dumps(message)
          socket.send(message)
      except Exception as e:
          print("Error sending:", e)
```
So instead of calling the client’s send method from the async test server, the test sends the message directly using the websockets library, while replicating the same message-formatting logic used by the client:

```
// IN the client.py we use this 
async def handler(websocket):
    message = expected_outgoing_message
    if isinstance(message, dict):
        message = json.dumps(message)
    await websocket.send(message)
```

so I changed the test server with the option 1 (to use the same logic )
```
async def handler(websocket):
        ###############
        #
        # Most important part of test where we test clients send message
        #
        ###############
        message = expected_outgoing_message
    
        if isinstance(message, dict):
            message = json.dumps(message)
        
        await websocket.send(message)
```
and then ran the test locally 
image
image
and then the test passed

What I concluded from this:

- We should not call the client’s static send method from an async test server
- Instead, the test should replicate the client’s message-formatting logic
- The message should then be sent directly using the websockets library
- Microcks validates the message payload against the AsyncAPI spec
- If Microcks accepts it, that confirms the client would send a correctly formatted message

so what do you think, is the approach good?

Harsh16gupta:
### **Dart template**

image

### **Js template**
image
While testing the JavaScript template, I found a small issue in the README.

```
By default this is testing against Hoppscotch echo service. You can modify `packages/templates/clients/websocket/javascript/example.js` and change first line to `const WSClient = require('./tests/temp/snapshotTestResult/client_postman/client.js');` and run `node example.js` again. You will see example still works but now it is using a different API. This is possible since both AsyncAPI documents define the same name of operation for sending messages: `sendEchoMessage` so each client generated has the same API.
```

The README currently mentions the following path: `./tests/temp/snapshotTestResult/client_postman/client.js` is wrong 
However, the correct path is:
`./test/temp/snapshotTestResult/client_postman/client.js`

### **Java template**
image

### **Python template**
image

All templates (Dart, JavaScript, Java, and Python) were verified by running them locally after applying the discussed fixes, and all are working correctly.

Please let me know if you’re okay with me committing all the remaining changes based on this.
Thank you for your time and review

Harsh16gupta:
@Adi-204 Thank you for the detailed feedback. I understand your concern now.

Option 1 doesn't actually test the client's send method. If someone modifies the client's `_send()` logic in the future, the test would still pass because it's using hardcoded logic instead of the actual client code. This defeats the purpose of the acceptance test. But we can't directly pass an async `websockets` socket to the client's synchronous `send_echo_message_static()` method.
This is why I went with Option 1, to work around the library incompatibility (I thought the best way to do this is by doing it hardcoded but didn't think about the future changes). But I see now that this approach sacrifices the test's actual purpose.

What's the recommended approach to solve this? Should I:

1. Change the test infrastructure to use a synchronous WebSocket server library?
2. Keep the test as it is and make a switch to Async WebSocket Library? 
3. A different approach you'd suggest?

Please let me know your preference and I'll update the PR accordingly.
Thank you for your patience!

Adi-204:
@Harsh16gupta I tried to debug locally and find a solution such that we don't need to do a lot of code changes, but that doesn't seem to be easy task. 

Two options 
1. Sync operations use websocket-client (Make changes in Acceptance test infra)
2. Async operations use websockets (Make changes in Generated client)

I checked with option 2 we will have to make a lot of changes and I'm not 100% sure about how much changes needs to done in option 1 as acceptance test was setup by @derberg so he is the best person to know about that. Maybe you can try analyze acceptance test infra, @derberg wdyt 🤔 ? approach should we go for?

Harsh16gupta:
@Adi-204 That makes sense, thanks for digging into it.
I’ll try both approaches locally (adjusting the acceptance test infra vs. switching the generated client) and share my findings.
Since this issue is critical, would it be okay if I open separate PRs for the two additional issues I discovered during debugging, so we can keep concerns isolated and easier to review?

1st one
> I found that the Python generated client.py (for postman) does not include:
> 
> ```
> self.receive_operation_handlers = {}
> self.receive_operation_discriminators = []
> ```
> 
> Because of this, the runtime crashes when on_message tries to access these attributes. To verify this, I manually added these lines to the generated client.py:

2nd one
 While testing the JavaScript template, I found a small issue in the README.
> ```
> By default this is testing against Hoppscotch echo service. You can modify `packages/templates/clients/websocket/javascript/example.js` and change first line to `const WSClient = require('./tests/temp/snapshotTestResult/client_postman/client.js');` and run `node example.js` again. You will see example still works but now it is using a different API. This is possible since both AsyncAPI documents define the same name of operation for sending messages: `sendEchoMessage` so each client generated has the same API.
> ```
> 
> The README currently mentions the following path: `./tests/temp/snapshotTestResult/client_postman/client.js` is wrong 
However, the correct path is: `./test/temp/snapshotTestResult/client_postman/client.js`
> 
Just a suggestion, happy to keep it in this PR

Adi-204:
@Harsh16gupta yeah the 1st one you mention is missing but just adding that won't enable clients to work and scope of issue is to solve not working clients.
The wrong path you are talking about was fixed with PR https://github.com/asyncapi/generator/pull/1890 update your branch.

Harsh16gupta:
Hey maintainers 

I tried Option 1 (it worked and was much simpler than Option 2), as discussed.
I’ve documented the full reasoning, the changes made, and the test output here:
https://docs.google.com/document/d/1DFdwB9S40MkFTIq5ekm8qJ7OXPHelsuc_zHmKwUZnAk/edit?usp=sharing

Please let me know if this direction makes sense or if you’d prefer a different approach.
Thanks!

Adi-204:
@Harsh16gupta no need to prepare docs and stuff, always just directly drop your research in github comments. Because here we don't have much options I would recommend you can proceed with your proposed solution. But remember if you are doing changes in acceptance test we need review/approval from @derberg as he introduce this testing and have much more idea/vision than me. He won't review docs, so just push your changes in this PR!

Harsh16gupta:
Thanks for the guidance, Adi!
I’ll proceed with the proposed solution directly in this PR.

Harsh16gupta:
Addressing CodeRabbit feedback.

Harsh16gupta:
Hi maintainers, all requested changes are done. Templates are verified (screenshots attached in the PR description), and acceptance tests are also passing locally.
Happy to make any changes if needed. Thanks!

packages/templates/clients/websocket/python/test/components/__snapshots__/Requires.test.js.snap
```diff
@@ -1,4 +1,4 @@
-// Jest Snapshot v1, https://jestjs.io/docs/snapshot-testing
+// Jest Snapshot v1, https://goo.gl/fbAQLP
```
Adi-204, line 1: why you made this changes?

packages/templates/clients/websocket/test/python/test_acceptance.py
```diff
@@ -54,37 +53,56 @@ async def test_hoppscotch_client_sends_message():
         "testEndpoint": "ws://websocket-acceptance-tester-py:8083/ws",
         "runnerType": "ASYNC_API_SCHEMA",
         "timeout": 30000,
-        "filteredOperations": ["RECEIVE handleEchoMessage"]
+        "filteredOperations": ["SEND handleEchoMessage"]
```
Adi-204: I don't think this is correct change https://github.com/asyncapi/generator/blob/3c5a48e0aacb3234e75d2f92e6b0dd47226e96ec/packages/templates/clients/websocket/test/__fixtures__/asyncapi-hoppscotch-server.yml#L27 
as it is RECEIVE operation according to AsyncAPI file.

packages/templates/clients/websocket/python/test/components/__snapshots__/Requires.test.js.snap
Harsh16gupta: When I updated the snapshots it got changed, I checked with other snapshots and they had the same link so I kept the change.

packages/templates/clients/websocket/test/python/test_acceptance.py
Harsh16gupta: I changed it to SEND because the test server sends a message, but I now understand that `filteredOperations` must match the operation definition in the AsyncAPI document, Will update it to RECEIVE.

Harsh16gupta:
## How the Synchronous Test Approach Works the Same as Before

### Same Validation Flow:
1. Start WebSocket server
2. Trigger Microcks test
3. Microcks connects and validates message
4. Poll for results

### What Changed:

| Aspect | Old (Async) | New (Sync) | Why |
|--------|------------|-----------|-----|
| **Concurrency** | `async/await` (asyncio) | `threading` | Sync client needs sync server |
| **Library** | `websockets` (async) | `websockets.sync.server` | Provides sync sockets |
| **Socket Type** | `WebSocketServerProtocol` (async) | `ServerConnection` (sync) | Compatible with websocket-client |
| **Sleep** | `await asyncio.sleep()` | `time.sleep()` | Sync equivalent |
| **Server Lifecycle** | `await websockets.serve()` | `with serve()` + threading | Sync pattern with background execution |

### The Critical Improvement:

**Old Version:**
```python
async def handler(websocket):
    await HoppscotchEchoWebSocketClient.send_echo_message_static(
        expected_outgoing_message, 
        websocket
    )
```
Incompatible with sync client

**New Version:**
```python
def handler(websocket):
    HoppscotchEchoWebSocketClient.send_echo_message_static(
        expected_outgoing_message, 
        websocket
    )
```

Now working correctly:
- The sync `websocket` (ServerConnection) is compatible with client's sync method
- Client uses `websocket-client` library with sync socket interface
- No async/sync mismatch

### Threading

**Old approach:** Used async/await for concurrency
```python
server = await websockets.serve(handler, port=port)
await asyncio.sleep(1)
# ... poll Microcks ...
await asyncio.sleep(2)
```

**New approach:** Uses threading for concurrency
```python
def run_server():
    with serve(handler, "0.0.0.0", port) as server:
        server_ready.set()
        server.serve_forever()  # Blocks this thread

server_thread = threading.Thread(target=run_server, daemon=True)
server_thread.start()
# Main thread continues to poll Microcks
```

**Both provide concurrency, just different mechanisms:**
- Async/await: Event loop manages coroutines in single thread
- Threading: OS manages multiple threads in parallel

**Result:** Same validation logic, correct sync/sync compatibility. Both test client functionality, message formatting, and AsyncAPI compliance identically.

packages/templates/clients/websocket/python/test/components/__snapshots__/Requires.test.js.snap
Harsh16gupta: After checking the [Jest upgrade notes](https://github.com/jestjs/jest/blob/fe7f28c9d1941f5c2726831cd9d9e479b401610e/website/versioned_docs/version-30.0/UpgradingToJest30.md?plain=1#L135C12-L135C18), I see that the full documentation link is used starting from Jest 30 and both links resolve to the same page.
Since the repository is currently using an older Jest version and other snapshots still use the shortened link, I think it makes sense to keep the existing format for consistency, what do you think?

PR #2003, opened 2026-02-15
feat: add error handling to generator components

### LINKED ISSUE #1712: Proper Error handling mechanism for Baked-in Templates

batchu5:
Add Error handling in generator components by handling edge cases and Invalid params.

Fixes #1712

## Summary by CodeRabbit

* **New Features**
  * Standardized, descriptive error responses for invalid or unsupported inputs.

* **Bug Fixes**
  * Early runtime validation for language, framework, role, and other inputs to prevent silent failures.
  * Conditional rendering adjusted to avoid showing incorrect output for unsupported languages.
  * Improved error messages now list supported options for clearer guidance.

* **Tests**
  * Updated tests to assert the new error-throwing behavior for invalid inputs.


packages/components/src/components/MethodGenerator.js
```diff
@@ -133,6 +135,28 @@ export function MethodGenerator({
...
 }) {
+  const supportedLanguages = Object.keys(defaultMethodConfig);
+  
+  if (!supportedLanguages.includes(language)) {
+    throw unsupportedLanguage(language, supportedLanguages);
+  }
+
+  if (typeof methodName !== 'string') {
+    throw invalidMethodName();
+  }
```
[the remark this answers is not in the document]
batchu5: @Adi-204 I'm a little bit confused here, because in https://github.com/asyncapi/generator/blob/2267f02be57c3ac8b21021fbe53d0423fd19d582/packages/templates/clients/websocket/java/quarkus/components/InitConnector.js#L21 
We are calling CloseConnection, which in turn calls the `methodGenerator` with an empty string as the methodName.

What do you suggest?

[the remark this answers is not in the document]
Adi-204: @batchu5 honestly I don't have much idea as the template was develop during last year GSoC project by other guy. It also depends what params for a component you are keeping as optional/required and logic inside the component. 
A small advice from myside would be after refactoring you do, please make sure integration test snapshots are not changed too much as it means template is generated as older version.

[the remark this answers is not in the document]
Adi-204: @batchu5 one imp thing remember you will have to look at how components are used in different templates and decide which params of component are required or optional. As you can't make every param as required becoz than it would fail for some templates.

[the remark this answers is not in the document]
batchu5: okay @Adi-204

packages/components/src/utils/ErrorHandling.js
```diff
@@ -0,0 +1,71 @@
...
+  throw new Error('Invalid AsyncAPI operation. Expected a valid operation with id().');
+}
+
+export function invalidRole(role, supported = []) {
+  throw new Error(`Unsupported role "${role}". Supported roles: ${supported.join(', ')}`);
+}
+
+export function negativeIndent(indent) {
+  throw new Error(`Indent size must be >= 0. Received: ${indent}`);
+}
```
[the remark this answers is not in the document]
batchu5: @Adi-204, should we add the jsDoC comments here? I feel like they are just throwing the string, so adding jsdocs wouldn't make much sense 

and also we are not adding here - 
https://github.com/asyncapi/generator/blob/master/apps/generator/lib/logMessages.js

Adi-204 (commented):
@batchu5 please drop a comment in PR, when it is ready to review. I can see many coderabbit comments and take your time!

packages/components/src/utils/ErrorHandling.js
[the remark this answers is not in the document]
Adi-204: @batchu5 I agree, it doesn't make a lot of sense to add JSDocs here.

packages/components/src/components/MethodGenerator.js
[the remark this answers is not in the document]
batchu5: At the moment, I'll keep it as it is!

batchu5:
@Adi-204, I have added the error handling that I have written in the proposal with tests and JSdocs. Please review it. 
This is a code coverage report for the tests.

image

In` Dependency Provider`. The branch coverage is not 100% because of the `resolveFrameworkDependencies` function.

The last condition is to return `frameworkConfig.dependencies || []` but in dependenciesConfig, we don't have dependencies in Java -> Quarkus, rather Java -> Quarkus -> role.

We can remove it for now; however, I'm not sure about this.

packages/components/src/utils/ErrorHandling.js
```diff
@@ -0,0 +1,71 @@
...
+  return new Error(`Unsupported framework "${framework}" for language "${language}". Supported frameworks: ${supportedList}`);
+}
+
+export function invalidMethodName(methodName) {
+  return new Error(`Invalid method name. Expected a non-empty string. Received: ${stringify(methodName)}`);
+}
+
+export function invalidNewLines(newLines) {
+  return new Error(`Invalid newLines value. Must be >= 0. Received: ${stringify(newLines)}`);
+}
```
Adi-204: . negativeIndent and invalidNewLines are the same concept
Both validate "must be >= 0" numeric values. Unify them:

```js
export function invalidNonNegativeInteger(field, value) {
  return new Error(`"${field}" must be >= 0. Received: ${stringify(value)}`);
}
```

packages/components/src/utils/ErrorHandling.js
```diff
@@ -0,0 +1,71 @@
...
+  return new Error(`Unsupported language "${language}". Supported languages: ${supportedList}`);
+}
+
+export function unsupportedFramework(language, framework, supported = []) {
+  const supportedList = supported.join(', ');
+  if (framework === undefined || framework === null || framework === '') {
+    return new Error(`Framework is required for language "${language}". Supported frameworks: ${supportedList}`);
+  }
+  return new Error(`Unsupported framework "${framework}" for language "${language}". Supported frameworks: ${supportedList}`);
+}
```
Adi-204, line 55: The unsupportedLanguage and unsupportedFramework functions both check for empty/null values inline. You could extract that:

```js
const isEmpty = (v) => v === undefined || v === null || v === '';
const list = (arr) => arr.join(', ');
```

packages/components/src/utils/ErrorHandling.js
```diff
@@ -0,0 +1,71 @@
...
+  return new Error('Invalid AsyncAPI operation. Expected a valid operation with id().');
+}
+
+export function invalidRole(role, supported = []) {
+  return new Error(`Unsupported role "${role}". Supported roles: ${supported.join(', ')}`);
+}
+
+export function negativeIndent(indent) {
+  return new Error(`Indent size must be >= 0. Received: ${stringify(indent)}`);
+}
```
Adi-204, line 111: @batchu5 the initial version is nice but I think we can do much better. 

1. I don't think we need to return "new Error()" from every function rather create a function createError() which returns error object.
2. I like the concept of having "Error codes" as it makes programmatic handling much better. we can have a something like below 
```js
const ERROR_CODES = {
  // Input validation
  UNSUPPORTED_LANGUAGE:      'ERR_UNSUPPORTED_LANGUAGE',
  UNSUPPORTED_FRAMEWORK:     'ERR_UNSUPPORTED_FRAMEWORK',
  // AsyncAPI document
  MISSING_ASYNC_API:         'ERR_MISSING_ASYNC_API',
  MISSING_INFO:              'ERR_MISSING_INFO',
};
```
The above is just sample/example.

I hope it make sense! Feel free to ask questions.

Adi-204:
@batchu5 I would recommend to read https://github.com/asyncapi/generator/pull/2003#discussion_r2835918101 comment first as it is for complete ErrorHandling.js file

batchu5:
@Adi-204, I have added `ERROR_CODES`, please look into it.

packages/components/src/utils/ErrorHandling.js
```diff
@@ -0,0 +1,107 @@
...
+  return createError(ERROR_CODES.INVALID_CLIENT_FILE_NAME, `Invalid client file name. Expected a non-empty string. Received: ${stringify(clientFileName)}`);
+}
+
+export function invalidParams(params) {
+  return createError(ERROR_CODES.INVALID_PARAMS, `Invalid params object. Received: ${stringify(params)}`);
+}
+
+export function invalidAsyncAPI() {
+  return createError(ERROR_CODES.MISSING_ASYNC_API, 'AsyncAPI document is missing.');
+}
```
Adi-204: I think function naming can be better because `invalidAsyncAPI` means actually document is present but it is not a valid asyncapi doc something is wrong, but when I look at the error it is returning says "doc is missing" which is not aligned. I think `missingAsyncAPIDocument` could be much better. wdyt?

packages/components/src/utils/ErrorHandling.js
```diff
@@ -0,0 +1,107 @@
...
+
+export function invalidInfo() {
+  return createError(ERROR_CODES.MISSING_INFO, 'AsyncAPI "info" object is missing.');
+}
```
Adi-204: same feedback as https://github.com/asyncapi/generator/pull/2003/changes#r2854538876

packages/components/src/utils/ErrorHandling.js
```diff
@@ -0,0 +1,107 @@
...
+
+export function invalidServer() {
+  return createError(ERROR_CODES.MISSING_SERVER, 'AsyncAPI "server" object is missing.');
+}
```
Adi-204, line 87: same feedback as https://github.com/asyncapi/generator/pull/2003/changes#r2854538876

packages/components/src/utils/ErrorHandling.js
```diff
@@ -0,0 +1,107 @@
...
+
+export function invalidOperation() {
+  return createError(ERROR_CODES.INVALID_OPERATION, 'Invalid AsyncAPI operation. Expected a valid operation with id().');
+}
+
+export function invalidRole(role, supported = []) {
+  return createError(ERROR_CODES.INVALID_ROLE, `Unsupported role "${role}". Supported roles: ${supported.join(', ')}`);
+}
```
Adi-204: I think `unsupportedRole` would make more sense and consistency with lang/framework. Also change error code. We can also use list() instead of supported.join(', ').

packages/components/src/utils/ErrorHandling.js
```diff
@@ -0,0 +1,107 @@
+const stringify = (v) => {
+  try {
+    return JSON.stringify(v);
+  } catch {
+    return String(v);
+  }
+};
+
+const isEmpty = (v) => v === undefined || v === null || v === '';
+const list = (arr) => arr.join(', ');
```
Adi-204:
```suggestion
const list = (arr) => arr.length ? arr.join(', ') : '(none)';
```

usually supported should not be empty but if in case the list will be an empty string and error message will be not helpful.

packages/components/src/utils/ErrorHandling.js
```diff
@@ -0,0 +1,107 @@
...
+  INVALID_CLIENT_NAME: 'ERR_INVALID_CLIENT_NAME',
+  INVALID_CLIENT_FILE_NAME: 'ERR_INVALID_CLIENT_FILE_NAME',
+  INVALID_PARAMS: 'ERR_INVALID_PARAMS',
+  MISSING_ASYNC_API: 'ERR_MISSING_ASYNC_API',
+  MISSING_INFO: 'ERR_MISSING_INFO',
+  MISSING_SERVER: 'ERR_MISSING_SERVER',
+  INVALID_OPERATION: 'ERR_INVALID_OPERATION',
+  INVALID_ROLE: 'ERR_INVALID_ROLE',
+  NEGATIVE_INDENT: 'ERR_NEGATIVE_INDENT',
+};
```
Adi-204, line 27: why you have exported it?

packages/components/src/utils/ErrorHandling.js
```diff
@@ -0,0 +1,107 @@
...
+  return createError(ERROR_CODES.INVALID_OPERATION, 'Invalid AsyncAPI operation. Expected a valid operation with id().');
+}
+
+export function invalidRole(role, supported = []) {
+  return createError(ERROR_CODES.INVALID_ROLE, `Unsupported role "${role}". Supported roles: ${supported.join(', ')}`);
+}
+
+export function invalidNonNegativeInteger(code, field, value) {
+  return createError(code, `"${field}" must be >= 0. Received: ${stringify(value)}`);
+}
```
Adi-204: why you have exported it?

Adi-204 (commented):
@batchu5 left few comments! Please also move all the functions which are not exported at top of file, so easier to track.

packages/components/src/utils/ErrorHandling.js
batchu5: I thought of using error codes in the test files, if you approve it

packages/components/src/utils/ErrorHandling.js
batchu5: yup. will change it!

packages/components/src/utils/ErrorHandling.js
Adi-204: @batchu5 you missed this comment

packages/components/test/components/Connect.test.js
```diff
@@ -13,4 +13,21 @@ describe('Testing of Connect function', () => {
...
   });
+
+  test('throws an error when unsupported language is provided', () => {
+    expect(() => render(
+      <Connect 
+        language='java'
+        title='HoppscotchEchoWebSocketClient'
+      />
+    )).toThrow(/Unsupported language "java". Supported languages:/);
+  });
```
Adi-204, line 24: why the list of supported language is not logged? `Supported languages:`
I see in some test it is logged but in others it is not?

batchu5: If a new language is added, we need to change the tests also, so isn't it better not to log the supported languages? wdyt??

Adi-204: @batchu5 got it make sense I missed something and it looks good. But I would like consistency so please from the test in which you have added supported lang list as someone might get confuse.

Adi-204 (commented):
@batchu5 nice job 👏 just few changes in test as I wrote in last comment. Also please open follow-up PR to update JSdocs such that we can mention in docs about required params for a component and also info about error handling, but it will be follow-up.

packages/components/src/utils/ErrorHandling.js
batchu5: I have done it by mistake. I think it is not needed as we are not importing it into any other files.

Adi-204 (dismissed):
Nice job @batchu5 👏

Adi-204 (dismissed):
lgtm @derberg can you have a final look and merge it!

Adi-204:
Screenshot 2026-03-22 at 00 00 22

@derberg also I'm not sure why I can't merge PR but most probably because of changeset.

derberg (approved):

PR #2023, opened 2026-03-05
test: fix shared mock state and improve test isolation

### LINKED ISSUE #1946: [BUG] :-Tests in `generator.test.js` rely on shared mutable mocks and non-awaited async assertions, breaking isolation
### Describe the bug.

While reviewing `generator.test.js`, I noticed two related testing issues that break test isolation and can lead to flaky or misleading results.
Problem 1: Shared mutable mock state across tests
 
The file globally mocks several modules:

Image

Later, individual tests mutate internal mock state, for example:

Image
However ,

This state is not reset between tests
Mocks are loaded once and shared across all test cases
Test behavior becomes order-dependent
Parallel Jest execution is unsafe
This violates test isolation and can cause hidden coupling between tests.

Problem 2: Async assertions using `setTimeout` are not awaited

Image

This is problematic because:

Jest does not wait for this callback
Assertions may never execute
Tests can pass even if the expectation fails
Results are non-deterministic

### Expected behavior

Each test should run in full isolation, with no shared mutable mock state leaking between tests.
Mocked modules should be reset or reloaded per test.
All async assertions should be properly awaited, so that failures are correctly detected by Jest.
Test results should be deterministic, order-independent, and reliable in CI

### Screenshots

Image

Image

### How to Reproduce

Clone the repo
visit:-generator/apps/generator/test/generator.test.js
Analyse the file

- Operating System (OS):Window 11
- Browser: Brave
- Browser Version:

SHUBHANSHU602:
## Description

This PR fixes test isolation issues in `apps/generator/test/generator.test.js`.

Several tests mutate shared state in the mocked `utils` module (such as `__files`, `__contentOfFetchedFile`, `__isFileSystemPathValue`, and `__getTemplateDetails`). Since the mock is loaded once at the module level, these mutations persisted across tests and could lead to unintended side effects between test cases.

## Changes

* Added a `beforeEach` block to reset mutable state in the `utils` mock before every test.
* Cleared Jest mock call history using `jest.clearAllMocks()` to ensure each test starts with a clean state.
* Replaced `setTimeout`-based assertions with deterministic expectations to avoid async assertions that Jest does not wait for.
* Fixed minor lint issues (such as removing a useless path segment and adding missing trailing commas).

##Result

* Prevents shared mock state from leaking between tests.
* Improves reliability and determinism of the test suite.
* Ensures tests remain independent and CI-safe.

Screenshot 2026-03-05 161145

**Resolves #1946

## Summary by CodeRabbit

* **Tests**
  * Improved test reliability by resetting mocks and mutable test state before each test.
  * Replaced timing-dependent assertions with immediate synchronous checks for more consistent, deterministic test outcomes.


apps/generator/test/generator.test.js
```diff
@@ -1,4 +1,7 @@
 /* eslint-disable sonarjs/no-duplicate-string */
+jest.mock('@asyncapi/parser', () => ({
+  convertToOldAPI: jest.fn((doc) => doc),
+}));
```
Adi-204: why we need this change?

SHUBHANSHU602: @Adi-204  I know it seems like voilating  the scope, BUT 

While fixing the isolation issue, the tests started reaching the actual parser conversion logic inside `lib/parser.js`. Previously, because of shared mutable state in the mocks and the way tests were structured, this code path was not always executed consistently.

When the tests began running in a properly isolated state, they attempted to call `convertToOldAPI` from `@asyncapi/parser`. However, newer versions of `@asyncapi/parser` no longer expose this function, which results in the runtime error:

```
TypeError: convertToOldAPI is not a function
```

Since the goal of this test suite is to verify the behavior of the `Generator` class rather than the internals of the parser library, I mocked `convertToOldAPI` to return the document unchanged. This keeps the tests focused on generator behavior while avoiding failures caused by parser implementation details.

If you prefer, I can also adjust the tests to avoid mocking the parser and instead rely on the actual parser behavior.

Adi-204: Hey @SHUBHANSHU602, I tried removing the mock locally and ran the tests, and they all passed without any TypeError. So I’m not able to reproduce the issue on my end. Because of that, I’m not sure how the tests are reaching the parser conversion logic in lib/parser.js. Could you share a bit more detail on how you observed this behavior? That would help me better understand the issue and verify it properly.

SHUBHANSHU602: @Adi-204 I think I have made the things quite complicated , I will rewrite all the things , to keep the things simple and to avoid scope voilation and confusion.

Adi-204: @SHUBHANSHU602 any updates on this one?

SHUBHANSHU602: @Adi-204 I will fix it soon.

SHUBHANSHU602:
@coderabbitai can you please review my pr and tell me if my solution satisfy the need of issue?

apps/generator/test/generator.test.js
SHUBHANSHU602: @Adi-204 have a look and let me know if you want me to make any further changes.

Adi-204:
@derberg can you please have final look and merge it!

derberg (approved):

SHUBHANSHU602:
Thanks @derberg

PR #2049, opened 2026-04-10
ci: change filters of dorny/paths-filter@v4

Adi-204:
**Description**
I tested changes in my fork and it should work now correctly.

**Related issue(s)**
Fixes #2043 

## Summary by CodeRabbit

No applicable changes for end-users. This release contains internal adjustments to continuous integration workflow path-filtering logic that do not impact user-facing features or functionality.

### DIFF
--- a/.github/workflows/pr-testing-with-test-project.yml
+++ b/.github/workflows/pr-testing-with-test-project.yml
@@ -19,14 +19,9 @@ jobs:
           predicate-quantifier: 'some'
           filters: |
             modified_files:
-              - 'package.json'
-              - 'package-lock.json'
-              - 'apps/**'
-              - 'packages/**'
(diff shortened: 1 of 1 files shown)


derberg:
why did you have to remove so many patterns

Adi-204:
@derberg the initial paths where not working and this should work fine why - "According to my understanding if a file is true for a single modified_files then it does't check all the checks". 

For example if a contributor only changed - `apps/generator/example.md`

In the old setup we had this file will match with `apps/**` and return true which leads to running test which we don't want and doesn't check for `'!**/*.md'` which we have later in the check. 

So in my solution I removed all other conditions and only kept "!" with all the docs files. 

I hope it answers your question but I can explain in more detail if you want.

derberg (approved):

PR #2051, opened 2026-04-13
fix: remove async from Promise executor in hooksRegistry (SonarQube S4146)

armorbreak001:
## Description

Fixes **SonarQube Bug S4146**: "Promise executor functions should not be async."

### Problem

`registerLocalHooks()` in `hooksRegistry.js` used `new Promise(async (resolve, reject) => {...})`. This is an anti-pattern because:

1. If the async executor throws synchronously before any `await`, the error becomes a rejected promise instead of being thrown
2. Errors in async code within the executor can be silently swallowed
3. It adds unnecessary complexity

### Fix

Removed `async` keyword from:
- The Promise executor function itself
- The walker `"file"` event handler (no await needed)
- The walker `"end"` event handler (no await needed)

### Behavior

**Identical** before and after. No functional change.

### SonarQube Impact

Resolves: https://sonarcloud.io/project/issues?id=asyncapi_generator&resolved=false&types=BUG&rules=javascript%3AS4146

## Release Notes

* **Refactor**
  * Internal optimization of hook registration logic with improved code efficiency. No user-facing changes.

### DIFF
--- a/apps/generator/lib/hooksRegistry.js
+++ b/apps/generator/lib/hooksRegistry.js
@@ -25,19 +25,19 @@ async function registerHooks (hooks, templateConfig, templateDir, hooksDir) {
  * @param {String} hooksDir Directory where local hooks are located.
  */
 async function registerLocalHooks(hooks, templateDir, hooksDir) {
-  return new Promise(async (resolve, reject) => {
-    const localHooks = path.resolve(templateDir, hooksDir);
+  const localHooks = path.resolve(templateDir, hooksDir);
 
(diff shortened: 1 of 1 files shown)


Adi-204:
Closing this PR as the activity looks like bot/spam. If this is a genuine contribution, please first show interest by opening or commenting on an issue. Once it’s assigned to you, you can raise a PR following our contribution guidelines:
https://github.com/asyncapi/generator/blob/master/CONTRIBUTING.md

PR #2061, opened 2026-04-13
feat: add generateOnly option to filter which template files are generated

### LINKED ISSUE #581: Make it possible to generate only given template files
#### Reason/Context
I'd like to use generator not only to scaffold my project when I start working on my application. I also want to use it continuously, every time I change something in AsyncAPI document.

The problem is that regeneration means basically another generation of the entire template. This leads to a process of regenerate -> carefully go through each change introduced after regenerating and ignore changes that will break your code

More context [here](https://deploy-preview-237--asyncapi-website.netlify.app/blog/websocket-part3#keeping-asyncapi-in-sync-with-code).

#### Description

A partial solution for the problem, but also not very complicated would be a configuration I could pass to the library/cli where I could specify what files to include/exclude in a generation.

Let's say I use some Node.js template. I generated my app and later added some more dependencies for my business login. Later I want to regenerate, but by telling the generator to ignore **package.json** and **package-lock.json** files.

What do you think? is the use case clear? you think it makes sense?

armorbreak001:
Fixes #581

Adds a new `generateOnly` option (analogous to the existing `noOverwriteGlobs`) that allows users to specify which files should be generated during template rendering.

## What changed

- Added `generateOnly` to the Generator constructor options
- When set (e.g., `generateOnly: ['**/*.md']`), only files matching the glob patterns are generated
- All other files are silently skipped during directory structure generation
- The filter applies to both regular file generation and separated file generation (e.g., per-channel/message files)
- Empty/unset `generateOnly` means generate everything (backward compatible, no behavior change)

## How it works

The new `shouldGenerateFile(filePath)` method uses `minimatch` (same library as `noOverwriteGlobs`) to check if a file path matches any of the provided glob patterns. The check is performed early in `generateFile()` and `generateSeparateFile()`, before any rendering or I/O occurs.

## How to verify

**Unit tests:**
```bash
cd apps/generator && npm test -- --testPathPattern=generator.test.js
```
6 new test cases cover: unset/empty generateOnly, matching/non-matching patterns, multi-pattern support.

**Integration test:**
```bash
cd apps/generator && npm test -- --testPathPattern=integration.test.js
```
Verifies that with `generateOnly: ['**/*.md']`, only .md files are generated while .html and other files are skipped.

**Manual usage:**
```js
const generator = new Generator('@asyncapi/html-template', './output', {
  generateOnly: ['**/*.md', '**/*.json'],
  forceWrite: true,
});
await generator.generateFromFile('./asyncapi.yaml');
```

Closes #581

## Summary by CodeRabbit

* **New Features**
  * Added `generateOnly` configuration option to control which template files are generated. Specify one or more glob patterns to selectively generate only matching files. Non-matching files are skipped. When unset or empty, all files are generated, maintaining full backward compatibility with existing workflows.

### DIFF
--- a/apps/generator/lib/generator.js
+++ b/apps/generator/lib/generator.js
@@ -42,7 +42,7 @@ const DEFAULT_TEMPLATES_DIR = path.resolve(ROOT_DIR, 'node_modules');
 
 const TRANSPILED_TEMPLATE_LOCATION = '__transpiled';
 const TEMPLATE_CONTENT_DIRNAME = 'template';
-const GENERATOR_OPTIONS = ['debug', 'disabledHooks', 'entrypoint', 'forceWrite', 'install', 'noOverwriteGlobs', 'output', 'templateParams', 'mapBaseUrlToFolder', 'url', 'auth', 'token', 'registry', 'compile'];
+const GENERATOR_OPTIONS = ['debug', 'disabledHooks', 'entrypoint', 'forceWrite', 'generateOnly', 'install', 'noOverwriteGlobs', 'output', 'templateParams', 'mapBaseUrlToFolder', 'url', 'auth', 'token', 'registry', 'compile'];
 const logMessage = require('./logMessages');
 
 const shouldIgnoreFile = filePath =>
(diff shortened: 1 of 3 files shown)


Adi-204:
Closing this PR as the activity looks like bot/spam. If this is a genuine contribution, please first show interest by opening or commenting on an issue. Once it’s assigned to you, you can raise a PR following our contribution guidelines:
https://github.com/asyncapi/generator/blob/master/CONTRIBUTING.md

PR #2091, opened 2026-05-17
chore: update build config

Florence-Njeri:
## Build Configuration Update

Updated build tooling configuration.

_Generated by SmokedMeat_

PR #2092, opened 2026-05-17
chore: hardening the GA workflows

Florence-Njeri:
**Description**

- Pinned all unpinned actions to a commit SHA. The goal is to move from tag pinned to SHA pinned to avoid tag move compromised from affecting our org
- The `manual_netlify_priview.yml` workflow satisfied most of the requirements of a` pwn_request` so I split it into two which is Github best practices. https://github.com/asyncapi/generator/pull/2091#issuecomment-4469637619
- Add a [zizmor](https://github.com/zizmorcore/zizmor) workflow which scans for common pipeline security issues such as:
    - **Hardcoded Secrets in Workflows** - API keys, passwords, or tokens hardcoded in YAML files
    - **Excessive Permissions** - Workflows with unnecessary write permissions
    - **Untrusted Actions** - Using third-party actions without proper verification
    - **Insecure Triggers** - Workflows triggered by external events without validation
    - **Missing Security Controls** - No approval processes for sensitive operations
- harden `update-docs-in-website.yml` workflow

- Add `permissions: {}` at workflow level with minimal job-level scopes 
  (contents: write, pull-requests: write) to fix excessive-permissions (without this block, the workflow inherits the write-all permissions of the parent)
- Set persist-credentials: false on the generator checkout since it is
  read-only; keep it true on the website checkout for git push
- Replace embedded-token push URL with git push origin HEAD to prevent
  token exposure in logs and process listings (template-injection)
- Move github.sha expansions into env vars to keep shell steps
  injection-safe
- Scope git config to the repository instead of --global
- 

Generated-by: Claude Code Opus 4.7 and Sonnet 5

**Related issue(s)**
TBC

## Summary by CodeRabbit

* **Chores**
  * Pinned third-party GitHub Actions to fixed revisions across multiple workflows for improved supply-chain stability.
  * Tightened workflow/job token permissions and improved CI credential handling with more least-privilege, read-only checkouts.
  * Updated PR review and maintenance automation (including workflow triggers, permissions, and Node.js upgrade where applicable).
* **New Features**
  * Added per-PR Netlify docs preview workflows (build + draft deploy) that post preview links back to pull requests.


derberg (commented):
- [x]  please address comments from bots
- [x] `uses: asyncapi/.github/.github/actions/get-node-version-from-package-lock@master` must stay, it is safe as it comes from our own repo. Changing to SHA will not work, as when we make a change to these actions and push them, we do not know SHA, so we would not be able to set it in workflows at the same time
- [ ] in few workflows where you switched to SHA, you also upgraded versions used, did you double check if there are no breaking changes? also for changes where you switch from one action to another?
- [x] I do not understand changes to netflify files, for one you created new file instead of editing old, so now hard to track what was changed - also why we need separate build workflow

Florence-Njeri:
> * in few workflows where you switched to SHA, you also upgraded versions used, did you double check if there are no breaking changes? also for changes where you switch from one action to another?

Hey @derberg I will address the first two comments and about testing the workfloas, I did not :( I am going to revert pinning to latest and only pin the actions to the sha of the version they are pinned to where possible

> * I do not understand changes to netflify files, for one you created new file instead of editing old, so now hard to track what was changed - also why we need separate build workflow

For the netlify file, the issue is that there is a pwn request (which allows for remote code execution of untrusted forks)This is cause because the workflow satisfies all the pwn request conditions:
1. It used `pull_request_target`, all prs including PRs created from forks with base repository permissions with the read write priviledged github token. 
2. `actions/checkout`-  it pulls attacker-controlled fork code into a runner that already holds the base repo's secrets.

The workflow does need the elevated priviledge to create the docs preview so the solutioin top this was to separate the code into two files. One file to pull the pull_request, checks out the PR, syncs docs, and builds the static site, then uploads the built output + PR number as an artifact.
The deploy jobs workflow triggers `workflow_run`  which runs in trusted context with secrets, downloads the prebuilt artifact, and runs the docs preview in a safe manner

Florence-Njeri:
@Shurtu-gal any ideas how to test the workflows still work after pînning actions SHA to the latest hashes?

Florence-Njeri:
> * [ ]   please address comments from bots
> * [x]  `uses: asyncapi/.github/.github/actions/get-node-version-from-package-lock@master` must stay, it is safe as it comes from our own repo. Changing to SHA will not work, as when we make a change to these actions and push them, we do not know SHA, so we would not be able to set it in workflows at the same time
> * [ ]  in few workflows where you switched to SHA, you also upgraded versions used, did you double check if there are no breaking changes? also for changes where you switch from one action to another?
> * [ ]  I do not understand changes to netflify files, for one you created new file instead of editing old, so now hard to track what was changed - also why we need separate build workflow

@derberg I changed the `uses: asyncapi/.github/.github/actions/get-node-version-from-package-lock` back to master but there is the risk of the master pointing to rogue commits like what happened this past week.

PR #2110, opened 2026-06-01
fix: security issue with `mapBaseUrlToFolder`

### LINKED ISSUE #1805: [BUG] Security - Path Traversal Vulnerability in URL-to-Folder Mapping
### Describe the bug.

A path traversal vulnerability exists in the `getMapBaseUrlToFolderResolvers` function in `apps/generator/lib/parser.js`. The function performs a simple string replacement without validating that the resulting path stays within the intended directory, which could allow an attacker to access files outside the specified base directory.

Location: `apps/generator/lib/parser.js:99-126`

Current Implementation:
function getMapBaseUrlToFolderResolvers({ url: baseUrl, folder: baseDir }) {
  const resolver = {
    order: 1,
    canRead: true,
    read(uri) {
      return new Promise(((resolve, reject) => {
        const path = uri.toString();
        const localpath = path.replace(baseUrl, baseDir);  // ⚠️ No validation
        try {
          fs.readFile(localpath, (err, data) => {
            if (err) {
              reject(`Error opening file "${localpath}"`);
            } else {
              resolve(data.toString());
            }
          });
        } catch (err) {
          reject(`Error opening file "${localpath}"`);
        }
      }));
    }
  };
  // ...
}

### Expected behavior

The generator should reject paths that attempt to escape the base directory Actual: The generator processes the path without validation, potentially accessing unauthorized files

### Screenshots

This is a code security vulnerability that doesn't require screenshots. 

### How to Reproduce

1.Set up a test environment with the mapBaseUrlToFolder option
2.Create a malicious AsyncAPI document with a schema reference like:
      components:
        schemas:
          User:
            $ref: 'https://schema.example.com/crm/../../../etc/passwd'
3.Configure the generator with:
        {
          mapBaseUrlToFolder: {
            url: 'https://schema.example.com/crm/',
            folder: './test/docs/'
          }
        }
4.Run the generator
5Observe that the path traversal sequence allows accessing files outside the base directory

Operating System (OS): N/A (code-level issue)
Browser: N/A
Browser Version: N/A

derberg:
fix https://github.com/asyncapi/generator/issues/1805

this security issue concerns only generator used as a service, but still it is there

added to changeset recommendation what people should have if this secuirty patch is a breaking change - although it should not

## Summary by CodeRabbit

* **Bug Fixes**
  * Hardened mapped `$ref` resolution to prevent path traversal outside the configured base folder, including cases that attempt to escape via symlinks.
  * Malicious references are now blocked with clearer error messaging.
* **Documentation**
  * Added a release note describing the security fix and guidance for remapping base URLs when cross-folder references are required.
* **Tests**
  * Expanded resolver tests to cover malicious escape attempts, successful in-folder resolutions, and symlink-based escape scenarios.


apps/generator/lib/parser.js
```diff
@@ -100,22 +102,38 @@ function getMapBaseUrlToFolderResolvers({ url: baseUrl, folder: baseDir }) {
...
+        // pointing elsewhere would pass the check above yet read a file outside baseDir.
+        // Canonicalize with realpath to close that gap. realpath needs the path to exist; a
+        // missing target is already proven contained lexically, so ENOENT is safe to ignore.
+        const realBase = await utils.realpath(resolvedBase);
+        const realTarget = await utils.realpath(resolvedTarget).catch((err) => {
+          if (err.code === 'ENOENT') return null;
+          throw err;
+        });
+        if (realTarget !== null && realTarget !== realBase && !realTarget.startsWith(realBase + path.sep)) {
+          throw new Error(logMessages.mappedRefOutsideBaseFolder(uriStr, baseUrl));
```
Adi-204: if we already have complete separate logic for symlink case I think it will be better to throw more helpful error. wdyt 🤔 maybe something like - `Reference "${ref}" resolves a symbolic link that points outside the mapped base "${mappedBaseUrl}". Blocked to prevent path traversal.`

derberg: done

magicmatatjahu (approved):

derberg:
`Missing AI-assistance disclosure` yes but that is for new PRs

/rtm

PR #2139, opened 2026-06-29
feat: generate spec-aware example files for Dart, JavaScript, and Python WebSocket clients

### LINKED ISSUE #1627: Start generating examples
Examples like:

- https://github.com/asyncapi/generator/blob/master/packages/templates/clients/websocket/javascript/example.js
- https://github.com/asyncapi/generator/blob/master/packages/templates/clients/websocket/python/example.py
- https://github.com/asyncapi/generator/blob/master/packages/templates/clients/websocket/dart/example.dart

These are hardcoded, but it causes issue:
- these examples are only available in the repository, so if someone generates something with the template, they do not have the example available first-hand
- these are easy to break, whenever we change something in templates, we forget to update examples

Of course it is not enouth to move example into template, and make sure it is dynamicly generated. We need to figure out how we can make sure these examples work. Maybe we should somehow use them in the acceptance tests?

Adi-204:
**Description**

- **Dart**: Replaces the hardcoded `example.dart` with a generated, spec-aware file emitted by `template/example.dart.js`. The example uses the real exported client class, registers per-op receive handlers, runs a bounded send loop with payloads from `message.examples()`, and wraps everything in proper error handling.
- **JavaScript**: Same pattern for JS — replaces the hardcoded example with `template/example.js.js`, using async/await, try/catch, a bounded `for` loop, and `main()` entry point.
- **Python**: Same pattern for Python — `template/example.py.js` emits a spec-aware `example.py` with PEP 8 indentation, per-op `register_on_<op>_handler` registration, `time.sleep(30)` keep-alive for receive-bearing specs (Python's `connect()` is non-blocking), and `if __name__ == '__main__': main()` entry point.
- All three clients share the same architecture: 9 local JSX components per client (composer + 8 sub-components) under `<client>/components/`, spec-awareness driven by two inline booleans (`hasSend`, `hasReceive`), four distinct body shapes (send+receive, send-only, receive-only, none), snapshot tests for components with conditional logic, integration snapshots regenerated to capture the new `example.*` output per fixture, legacy example files left in place (cleanup is follow-up work), and one `@asyncapi/generator` minor changeset.

**Related issue(s)**

Fixes #1627

**AI assistance**

- [x] This PR was created with AI assistance — `Generated-by:` Cursor Agent (Claude Opus 4.8)
- [ ] No AI assistance was used

## Summary by CodeRabbit

* **New Features**
  * Added new reusable WebSocket example components (Imports, OpenConnection/Close, Handlers, OutgoingProcessor, SendInvocations) and exposed them as public exports.
  * Updated runnable WebSocket client example generation for JavaScript, Python, and Dart to compose these shared components and conditionally include send/receive logic.
  * Added `exampleFileName` option for JS, Python, and Dart.
* **Documentation**
  * Expanded the “Reusable Components API” docs with full entries and examples for the new WebSocket components.
* **Tests**
  * Added snapshot and unsupported-language/error-handling coverage for the new components across supported languages.


derberg (dismissed):
looking great!

are you able to go back in cursor to the session in which you worked on this PR? would be great to ask agent to generate a first version of template authoring skill from it as in this PR you had to go through multiple different parts of template. Wdyt?

aeworxet:
@asyncapi/microgrant_team

Adi-204:
> looking great!
> 
> are you able to go back in cursor to the session in which you worked on this PR? would be great to ask agent to generate a first version of template authoring skill from it as in this PR you had to go through multiple different parts of template. Wdyt?

I tried during development but I was not able to build something strong which is more deterministic and works better in general not particular to examples.

derberg (approved):

PR #1509, opened 2025-04-18
chore: add Slack complex asyncapi fixture document with models and query params support

derberg:
Resolves #1418 

Almost done with first real-life Slack Websocket API example. Nice preview under: https://studio.asyncapi.com/?share=09273ef4-0431-4764-bbc8-8f318cca5b17

It will give us new challenges in templates development:
- handling models generation
- query params!
- dedicated handlers generation 
- handling reply

generated project for slack with models included:
```
└── 📁client_slack
    └── Acknowledge.py
    └── Authorization.py
    └── ChannelJoin.py
    └── client.py
    └── ConnectionInfo.py
    └── DebugInfo.py
    └── Disconnect.py
    └── DisconnectReasonEnum.py
    └── Event.py
    └── EventPayload.py
    └── Hello.py
    └── MessageDeleted.py
    └── ReactionAdded.py
    └── requirements.txt
```

some followup issues that I create after we merge this PR
- at the moment it is complex to put generated models in custom subfolder, like for example `models`. Some solve it with hooks: https://github.com/asyncapi/java-template/blob/master/hooks/createFolderStructure.js but I'd love us to explore how we can have this functionality available by default as part of https://github.com/asyncapi/generator/blob/master/packages/components/src/components/models.js or maybe even `File` component.
- add unit tests for `bindings.js#getQueryParams`
- add unit test for `utils.js#listFiles` to assure only filenames are returned without paths
- unify `npm run test:update` across the whole project. Basically make sure that whenever we have `jest --updateSnapshot` in package.json, the script is called `test:update` and instead of calling jest directly, it calls regular test script like so: `npm run test -- -u`. We also need `test:update` in main package.json
- client for slack: discuss how we can share credentials to not leak anything sensitive to the public, but still any maintainer should have easy way of plaing with example
- snapshot test for python Constructor.js
- snapshot test for python InitSignature.js
- snapshot test for python QueryParamsArgumentsDocs.js
- snapshot test for python QueryParamsVariables.js
- snapshot test for python Requires.js
- if you look into `python/test/integration.test.js` you see that once I introduced `listFiles` now basically all tests are the same, and only input arguments are different. I believe we can create a helper for generating snapshots that basically accepts differemt templateParams and fixtures and testsResults. There is too much boilerplate code. Also investigate, maybe we should have integration tests only in `websocket/test` and not in separate language-based folders
- explore how to handle `reply` operation in code generation
[...]

PR #2142, opened 2026-07-01
feat: add query Parameters support to javascript websocket client

### LINKED ISSUE #1957: JavaScript WebSocket client template doesn't support query parameters from channel bindings
## Problem
The JavaScript WebSocket client template doesn't extract or use query parameters defined in channel bindings, unlike the Python template. This prevents authentication tokens and other important parameters from being passed to WebSocket connections.

## Expected Behavior
When a channel defines query parameters in its bindings (e.g., `auth_token` for authentication), the generated JavaScript client should:
1. Accept these parameters in the constructor
2. Automatically append them to the WebSocket URL as query strings

Example:
```javascript
// Expected generated code
constructor(url, auth_token) {
  const params = {};
  if (auth_token) params['auth_token'] = auth_token;
  const qs = new URLSearchParams(params).toString();
  this.url = qs ? `${url}?${qs}` : url;
}

### PR DESCRIPTION
**Pull request by batchu5**
## Description

This PR adds support for query parameters to the generated JavaScript WebSocket client, bringing it closer to feature parity with the Python and Java clients. 

### Changes Made
- **JavaScript WebSocket Template** (`packages/templates/clients/websocket/javascript`): 
  - Updated `ClientClass.js` and `Constructor.js` to process and append query parameters to the WebSocket URL using the native Node.js `querystring` module.
  - Implemented `QueryParamsArgumentsDocs.js` to dynamically generate JSDoc `@param` descriptions for each query parameter.
  - Updated `InitSignature.js` to inject query parameters into the constructor signature, defaulting to the values specified in the AsyncAPI document.
  - Conditionally injected the `querystring` module dependency only when `queryParams` are present in the AsyncAPI document.

Generated-by: Claude Opus 4.6

Fixes #1957

## Summary by CodeRabbit

* **New Features**
  * WebSocket JavaScript clients now support optional URL query parameters, including constructor defaults and environment-variable values.
  * Generated clients provide documented, safely normalized query-parameter names.
  * Added a Slack WebSocket client example with connection and message handling.

* **Bug Fixes**
  * Query strings are appended only when parameters are provided, with improved URL assembly.

* **Tests**
  * Added coverage for query-parameter generation, documentation, name normalization, and Slack integration.

### CODE REVIEW

**Comment by batchu5** [comment]
> : _📐 Maintainability & Code Quality_ | _🔵 Trivial_ | _💤 Low value_
> 
> **Duplicated parser/fixture setup across test files.**
> 
> The parser instantiation, fixture path, and `beforeAll` block (lines 1-16) are identical to `QueryParamsArgumentsDocs.test.js`. Consider extracting a shared test helper to load the parsed document once for both test files.

@Adi-204 I wonder if we can actually put the Common JS file for parsing the AsyncAPI document, because in all other tests we don't follow that way 🤔

File: packages/components/src/components/QueryParamsVariables.js [hunk 1]
```diff
@@ -68,20 +68,20 @@ const queryParamLogicConfig = {
...
-        text: `params["${paramName}"] = ${paramName};`,
-        indent: 10,
+        text: `params["${paramName}"] = _${paramName};`,
+        indent: 2,
       },
       closing: {
         text: '}',
-        indent: 8,
+        indent: 0,
         newLines: 1,
```
Adi-204, line 86: The indent will not be 0 as it will render inside class so it depends upon indentation of surrounding code. Also any reason why you added "_" before variables? as in JS I don't think it makes any difference.

packages/templates/clients/websocket/javascript/components/getSafeJsName.js
```diff
@@ -0,0 +1,29 @@
...
+  
+  let candidate = safe;
+  let suffix = 1;
+  while (usedNames.has(candidate)) {
+    candidate = `${safe}_${suffix++}`;
+  }
+  usedNames.add(candidate);
+  
+  return candidate;
+}
```
Adi-204, line 89: Firstly I don't get point of this getSafeJsName becoz we can't just change param name given in the input asyncapi file. Secondly it is not a component it should be in file https://github.com/asyncapi/generator/blob/master/packages/helpers/src/utils.js

Adi-204 (commented):
@batchu5 the main problem over here is in JS we don't have slack support for now and currently slack example is only having query params. So it is difficult to review also how did you test changes locally?

packages/templates/clients/websocket/javascript/components/getSafeJsName.js
batchu5: The main use of getSafeJsName is, if the paramName that the user has sent is, for example, `user-id`, as it is not valid js variable name it would give a syntax error so to cut that off we are basically using this function

Sure thing, I should have put this in the utils.js. 
Also, since we can't change the paramName in the AsyncAPI doc, what if paramName is invalid? What can we actually do here??

packages/components/src/components/QueryParamsVariables.js
batchu5: image

 This is the generated client. For example, if I don't put `_` in front of the heartbeat, then I'll get an error like heartbeat is already initialized.  To avoid that error, I have put an underscore.
 
 Indentation, I have tried to put the indentation to 8, but if I do it like that, it is taking too much space at the front
 like this
image

batchu5:
I have tested the code using this [file](https://github.com/asyncapi/generator/blob/97161a533dfb497eb420bdada5249e587907e7c7/packages/templates/clients/websocket/test/__fixtures__/asyncapi-websocket-components.yml).

Adi-204:
> I have tested the code using this [file](https://github.com/asyncapi/generator/blob/97161a533dfb497eb420bdada5249e587907e7c7/packages/templates/clients/websocket/test/__fixtures__/asyncapi-websocket-components.yml).

@batchu5  yeah I'm not sure we should develop in this way as difficult to test going forward. Let's keep this on hold we can discuss in next meeting!

Adi-204:
hey @batchu5 we discuss in the generator meeting https://fathom.video/share/P7tie39Wzz6zzHK3VuY-kkBh8nbR_jBX you can have a look at it basically right now it is a bit difficult to isolate this adding a new component to a template as we are not able to test it properly I think better would be changing the scope of issue and you can continue in this PR adding support for slack client in JS. NO need to complicate it too much for Phase 1 we don't want everything that python is doing right now rather main objective is to have slack support in JS client. Feel free to ask question if any.

Adi-204:
@batchu5 new scope of the issue - "basic version of working slack example"

If you notice in python template https://github.com/asyncapi/generator/tree/master/packages/templates/clients/websocket/python we have 3 examples 

[example.py](https://github.com/asyncapi/generator/blob/master/packages/templates/clients/websocket/python/example.py), [example-slack.py](https://github.com/asyncapi/generator/blob/master/packages/templates/clients/websocket/python/example-slack.py) and  [example-slack-with-routing.py](https://github.com/asyncapi/generator/blob/master/packages/templates/clients/websocket/python/example-slack-with-routing.py)

Now to keep scope limited you only need to make sure for JS template something like example-slack.py is working which was introduce in https://github.com/asyncapi/generator/pull/1509 

What you DON"T need to DO - "automatic routing of message for slack" which was introduce in https://github.com/asyncapi/generator/pull/1814 keep this OUT OF SCOPE. 

Feel free to ask question!

packages/templates/clients/websocket/javascript/components/getSafeJsName.js
Adi-204: @batchu5 as discuss earlier move it to helpers. Also the approach to figure out valid JS variable name is way manual and hardcode, I'm sure there must be other ways built-in or using external library, go for built-in approach if available.

Adi-204 (commented):
@batchu5 one major comment rest looks good!

Adi-204:
If we keep the current approach of generating one constructor variable per query parameter, I think the helper should be moved to `packages/helpers/src/utils.js` because the same mapping is needed by `InitSignature`, `QueryParamsArgumentsDocs`, and `QueryParamsVariables`.

The current approach has two separate problems:

1. Some AsyncAPI names are not valid JavaScript variables:

```text
client-id -> clientId
class     -> _class
1st       -> _1st
```

Instead of maintaining the JavaScript reserved-word list manually, we can use [`@babel/helper-validator-identifier`](https://babeljs.io/docs/babel-helper-validator-identifier) in the shared helper. It already knows JavaScript identifier, keyword, and strict-mode binding rules.

2. Different names can become the same variable:

```text
client-id -> clientId
client_id -> clientId
```

Babel only validates names, so we still need to handle collisions. We can either add a suffix such as `clientId_1`, or keep the behavior simpler by throwing a clear generation-time error from the shared helper:

```text
Cannot generate JavaScript client: query parameters client-id and client_id both produce clientId.
```

The helper can also throw an error that includes the original AsyncAPI name when it cannot produce a valid JavaScript identifier. This makes the problem clear to the user instead of generating broken JavaScript. The trade-off is that throwing rejects query names that are valid in a URL but cannot be represented safely by this positional-argument API.

This gives us one shared validation path, removes the manually maintained language rules, and ensures all three generated-code locations behave consistently.

PR #2143, opened 2026-07-02
fix: release and pin generator dependencies

### LINKED ISSUE #2128: Avoid resolving archived @asyncapi/generator-hooks from npm
## Description

`apps/generator/package.json` currently declares:

```json
"@asyncapi/generator-hooks": "*"
```

Because that dependency is published as a wildcard range, package managers resolve it from the public npm registry when `@asyncapi/generator` is installed by consumers. The public package that gets resolved is the old `@asyncapi/generator-hooks@0.1.0` package, which appears to be deprecated/archived and no longer published as part of the current generator setup.

This shows up for consumers as a deprecated transitive dependency, for example through `@asyncapi/cli -> @asyncapi/generator -> @asyncapi/generator-hooks`.

## Why this seems unintended

- The generator package still lists `@asyncapi/generator-hooks` with `"*"`: https://github.com/asyncapi/generator/blob/master/apps/generator/package.json
- The generator v2.5.0 release notes mention that generator hooks are part of the generator repository / integrated out of the box: https://github.com/asyncapi/generator/releases/tag/%40asyncapi%2Fgenerator%402.5.0
- The old hooks repository is archived: https://github.com/asyncapi-archived-repos/generator-hooks
- A previous dependency cleanup issue mentioned this package as deprecated, but this is more than dependency hygiene because the wildcard causes consumers to resolve an archived package from npm: https://github.com/asyncapi/generator/issues/2004

## Expected behavior

Installing `@asyncapi/generator` should not pull the archived public `@asyncapi/generator-hooks@0.1.0` package from npm.

Possible fixes could be:

- remove `@asyncapi/generator-hooks` from the published generator package if hooks are already built in, or
- use an internal/workspace-only setup that does not publish a wildcard dependency that consumers resolve from npm.

## Actual behavior

Consumers installing packages that depend on `@asyncapi/generator` can get the archived public package as a transitive dependency, e.g. `@asyncapi/generator-hooks@0.1.0`, because the published dependency range is `"*"`.

Stanzilla:
## Description

Fixes #2128.

Generated-by: Codex (GPT-5)

This PR replaces wildcard ranges for published AsyncAPI workspace dependencies with their current released versions. It keeps `@asyncapi/generator-hooks` in the existing `apps/hooks` package and preserves the hook behavior and package boundary.

The changeset releases `@asyncapi/generator-hooks` as `0.1.1`. Changesets then updates the new `@asyncapi/generator` release to depend on `0.1.1`, so consumers no longer install the deprecated `0.1.0` package.

Affected packages:

- `@asyncapi/generator`
- `@asyncapi/generator-components`
- `@asyncapi/generator-hooks`

## Validation

- `npm exec --workspace=@asyncapi/generator-hooks -- npm test -- --runInBand`
- `npm exec --workspace=@asyncapi/generator-hooks -- npm run lint`
- `npm exec --workspace=@asyncapi/generator -- jest test/hooksRegistry.test.js --runInBand`
- `npm exec --workspace=@asyncapi/generator -- npm run lint`
- `npm install --package-lock-only --ignore-scripts --offline`
- `npx @changesets/cli status --output=/tmp/generator-changeset-status.json`
- temporary `changeset version` simulation confirmed hooks `0.1.1` and Generator dependency `0.1.1`
- `git diff --check`

## Summary by CodeRabbit

* **Improvements**
  * Standardized generator workspace package versions and pinned inter-package dependencies to fixed released versions for more consistent installation and reproducible behavior.

* **Chores**
  * Published patch release updates across the generator and related generator components packages (including hooks).

* **Tests**
  * Fixed the test suite ending structure for hooks-related tests to ensure the Jest run completes reliably.


derberg:
please explain changes to dependencies in package.json files, to explicit version instead of `*`

and also why such fix had to be applied for hooks

Stanzilla:
The explicit versions are needed because these package manifests are published to npm. `*` works inside the workspace during local development, but remains in the consumer-facing manifest and is resolved from the registry during installation. That makes installs nondeterministic and can select a package version different from the one tested with the Generator release.

After review, this PR keeps hooks in the existing separate `apps/hooks` package rather than duplicating its implementation inside Generator. The source dependencies are pinned to current workspace versions, and the changeset includes patch releases for Generator, components, and hooks.

`@asyncapi/generator-hooks@0.1.0` is currently deprecated, so pinning that version alone would not fix #2128. I verified the release flow with `changeset version`: it publishes hooks as `0.1.1` and rewrites the new Generator release to depend on `@asyncapi/generator-hooks@0.1.1`. Consumers therefore keep the existing hooks package boundary without installing the deprecated `0.1.0` release.

Stanzilla:
If you're curious, this is how it looks like when you install asyncapi right now:

Solo-2026-07-07-18 45 38

apps/generator/lib/officialHooks.js
```diff
@@ -0,0 +1,63 @@
...
+ *
+ * @param {Object} generator Generator instance.
+ * @param {String} generator.originalAsyncAPI Original AsyncAPI document content.
+ * @param {String} generator.targetDir Generation output directory.
+ * @param {Object} generator.templateParams Template parameters.
+ * @param {String} [generator.templateParams.asyncapiFileDir] Optional subdirectory within the generation output directory.
+ * @returns {Promise<void>} Resolves after the AsyncAPI document is written.
+ * @throws {Error} When asyncapiFileDir resolves outside targetDir, or when directory creation/file writing fails.
+ */
+async function createAsyncapiFile(generator) {
```
derberg: what about existing https://github.com/asyncapi/generator/blob/master/apps/hooks/src/index.js? it also has its own test

apps/generator/lib/officialHooks.js
```diff
@@ -0,0 +1,63 @@
...
+
+  if (!customDirInTarget) {
+    return resolvedTargetDir;
+  }
+
+  const outputDir = path.resolve(resolvedTargetDir, customDirInTarget);
+  const relativeOutputDir = path.relative(resolvedTargetDir, outputDir);
+
+  if (relativeOutputDir.startsWith('..') || path.isAbsolute(relativeOutputDir)) {
+    throw new Error(logMessage.asyncapiFileDirOutsideTarget(customDirInTarget));
```
derberg: this is basically a breaking change. Before there was no check like this, and now we will throw errors if someone pass `asyncapiFileDir` in a wrong way.

why was this introduces? is it AI overthinking or specific reason behind it?

apps/generator/lib/officialHooks.js
derberg: basically better if we first try to just set proper version on `@asyncapi/generator-hooks` in package.json and that should be enough - bundling hooks as part of generator is something we moved from on purose, to keep library separate

apps/generator/lib/officialHooks.js
Stanzilla: Agreed. The containment check came from automated review feedback and was not required for this issue. Removing the duplicate implementation also removes that check, so `asyncapiFileDir` keeps its existing behavior in `apps/hooks`.

apps/generator/lib/officialHooks.js
Stanzilla: Good point. I removed the duplicate `apps/generator/lib/officialHooks.js` implementation and its tests. Generator now resolves the existing `@asyncapi/generator-hooks` package from `apps/hooks`. The source dependency is pinned instead of using `*`, and the changeset includes a hooks patch release. A `changeset version` simulation confirms it publishes hooks as `0.1.1` and updates the new Generator release to depend on `0.1.1`, so consumers do not install the deprecated `0.1.0`. The existing hooks package tests remain the source of coverage.

apps/generator/package.json
```diff
@@ -48,10 +48,10 @@
   "license": "Apache-2.0",
   "homepage": "https://github.com/asyncapi/generator",
   "dependencies": {
-    "@asyncapi/generator-components": "*",
-    "@asyncapi/generator-helpers": "*",
-    "@asyncapi/generator-hooks": "*",
-    "@asyncapi/generator-react-sdk": "*",
+    "@asyncapi/generator-components": "0.7.0",
+    "@asyncapi/generator-helpers": "1.1.0",
+    "@asyncapi/generator-hooks": "0.1.0",
```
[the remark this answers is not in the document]
Stanzilla: The concern is valid for the intermediate `0.1.0` pin, but the latest commit adds `@asyncapi/generator-hooks` to the patch changeset. A `changeset version` simulation confirms hooks is published as `0.1.1` and the new Generator release is rewritten to depend on `0.1.1`, so consumers no longer install the deprecated `0.1.0`. The internal alias is intentionally not restored because the maintainer requested retaining the existing separate `apps/hooks` package boundary.

Stanzilla:
Updated this PR based on the review feedback.

What changed:

- removed the duplicate hooks implementation from `apps/generator`
- kept `createAsyncapiFile` in the existing `apps/hooks` package with its existing tests
- removed the new `asyncapiFileDir` containment check, preserving current behavior
- pinned the published AsyncAPI workspace dependencies instead of using `*`
- added `@asyncapi/generator-hooks` to the changeset so it is released as a new patch version

I also verified the release flow with `changeset version`. It produces `@asyncapi/generator-hooks@0.1.1` and updates the new `@asyncapi/generator` release to depend on `0.1.1`. This preserves the separate hooks package requested in review while ensuring consumers no longer install the deprecated `0.1.0` release.

@derberg is that what you had in mind?

derberg:
yeah, thanks, I will just undeprecate `@asyncapi/generator-hooks` and configure trusted publishing for it so it can also be separately published

derberg (approved):
🍏 

waiting for ci and will merge

trusted publishing in place, package deprecation removed

Stanzilla:
Thank you!

PR #2211, opened 2026-08-21
chore: define permissions in workflows  and pin actions to SHA

Florence-Njeri:
**Description**

- Pinned all unpinned actions to a commit SHA. The goal is to move from tag pinned to SHA pinned to avoid tag move compromised from affecting our org
- Add a [zizmor](https://github.com/zizmorcore/zizmor) workflow which scans for common pipeline security issues such as:
    - **Hardcoded Secrets in Workflows** - API keys, passwords, or tokens hardcoded in YAML files
    - **Excessive Permissions** - Workflows with unnecessary write permissions
    - **Untrusted Actions** - Using third-party actions without proper verification
    - **Insecure Triggers** - Workflows triggered by external events without validation
    - **Missing Security Controls** - No approval processes for sensitive operations
- harden `update-docs-in-website.yml` workflow

- Add `permissions: {}` at workflow level with minimal job-level scopes 
  (contents: write, pull-requests: write) to fix excessive-permissions (without this block, the workflow inherits the write-all permissions of the parent)
- Set persist-credentials: false on the generator checkout since it is
  read-only; keep it true on the website checkout for git push
- Replace embedded-token push URL with git push origin HEAD to prevent
  token exposure in logs and process listings (template-injection)
- Move github.sha expansions into env vars to keep shell steps
  injection-safe
- Scope git config to the repository instead of --global
- 

Generated-by: Claude Code Opus 4.7 and Sonnet 5

**Related issue(s)**
TBC

## Summary by CodeRabbit

- **Security**
  - Hardened automated workflows by restricting permissions and pinning external actions to specific revisions.
  - Improved credential handling to prevent unnecessary persistence and reduce exposure risks.

- **Workflow Updates**
  - Removed manual workflow dispatch options where no longer needed.
  - Updated automation triggers, release processing, documentation publishing, and maintainer workflows.
  - Upgraded workflow automation to Node.js 20 where applicable.


.github/workflows/local-generate-files.yml
```diff
@@ -44,7 +49,7 @@ jobs:
           branch: assets-update/${{github.event.release.tag_name}}
       - if: failure() # Only, on failure, send a message on the 94_bot-failing-ci slack channel
         name: Report workflow run status to Slack
-        uses: 8398a7/action-slack@v3
+        uses: rtCamp/action-slack-notify@e31e87e03dd19038e411e38ae27cbad084a90661 # v2.3.3
```
derberg, line 52: You changed the action that is used here, but the `with:` body did not change. The options left below do not match the new action: `rtCamp/action-slack-notify` declares no `inputs` at all, and it reads the webhook from `SLACK_WEBHOOK`, not `SLACK_WEBHOOK_URL`.

The question is: do we really need to switch the original action to a completely new one? What is the reason for it?

.github/workflows/release-with-changesets.yml
```diff
@@ -62,7 +69,7 @@ jobs:
         run: npm test --if-present
       - if: failure() # Only, on failure, send a message on the 94_bot-failing-ci slack channel
         name: Report workflow run status to Slack
-        uses: 8398a7/action-slack@v3
+        uses: rtCamp/action-slack-notify@e31e87e03dd19038e411e38ae27cbad084a90661 # v2.3.3
```
derberg, line 72: Same here. The action changed, the `with:` body did not, and the options left below do not match `rtCamp/action-slack-notify`.

.github/workflows/release-with-changesets.yml
```diff
@@ -133,7 +142,7 @@ jobs:
           GIT_COMMITTER_EMAIL: info@asyncapi.io
       - if: failure() # Only, on failure, send a message on the 94_bot-failing-ci Slack channel
         name: Report workflow run status to Slack
-        uses: 8398a7/action-slack@v3
+        uses: rtCamp/action-slack-notify@e31e87e03dd19038e411e38ae27cbad084a90661 # v2.3.3
```
derberg, line 145: Same here. The action changed, the `with:` body did not, and the options left below do not match `rtCamp/action-slack-notify`.

.github/workflows/update-docs-in-website.yml
```diff
@@ -8,32 +8,45 @@ on:
...
 
+permissions: {}
+
 jobs:
   Make-PR:
     name: Make PR on website repository with updated latest generator documentation
     runs-on: ubuntu-latest
+    permissions:
+      contents: write       # push to asyncapi/website via PAT
+      pull-requests: write  # gh pr create
```
derberg, line 19: These permissions apply to the `GITHUB_TOKEN` for this repo, not for `asyncapi/website`. We push and run `gh pr create` against the website repo, which is why the job uses `secrets.GH_TOKEN`: the built-in token cannot write to another repo. So nothing in this job uses the write access, and a PR about hardening ends up granting more than before.

PR #2215, opened 2026-08-28
test(generator): add unit tests for conditional generation and config…

devxdh:
Resolves [#1553 ](https://github.com/asyncapi/generator/issues/1553)
Related to [#1970 ](https://github.com/asyncapi/generator/issues/1970)

This PR adds comprehensive unit test coverage for conditional file and folder generation in @asyncapi/generator.
- Created a dedicated unit test suite in apps/generator/test/conditionalGeneration.test.js covering default generation, parameter-based conditions, subject-based JMESPath conditions (including server lookups), legacy conditionalFiles
backward compatibility, and missing validation handlers.
- Added test cases in apps/generator/test/templateConfigValidator.test.js to validate conditionalGeneration configuration entries (type checks for subject, parameter, mutual exclusivity, and Ajv schema compilation).
- Added a default return true fallback in isGenerationConditionMet for unconditioned paths.

**File Changes:**
```
M	apps/generator/lib/conditionalGeneration.js
D	apps/generator/test/conditionalGeneration.test.js
M	apps/generator/test/templateConfigValidator.test.js
```

**AI assistance**
- [X] This PR was created with AI assistance — Generated-by: Antigravity
- [ ] No AI assistance was used

## Summary by CodeRabbit

- **Bug Fixes**
  - Conditional generation now proceeds when no configured rule matches, preventing unexpected generation failures.

- **Tests**
  - Expanded coverage for configured and unmatched conditions, parameters, document fields, server protocols, legacy conditions, malformed queries, and missing validation.
  - Added validation coverage for invalid configurations and successful matching and non-matching values.

Issue #1553, opened 2025-05-09
deprecate conditionalFile configuration as we have introduced configurationGeneration config that will handle both file and folder generation.


AayushSaini101:
Reference: https://github.com/asyncapi/generator/pull/1515#pullrequestreview-2817388888

saurabhraghuvanshii:
Hi @AayushSaini101 can I give a try? and It will be helpful if you provide related docs!

derberg:
reopen, not open for work at the moment, but we need the issue for future

Satya900:
@derberg can i work on this issue

SHUBHANSHU602:
@derberg is this now open to work.

derberg:
it is open for contribution but not now, no rush to remove this functionality

devxdh:
Apologies. I messed up the 'References https://github.com/asyncapi/generator/issues/1970" which resulted in asyncapi-bot closing my Pull Requests.

PR #2225, opened 2026-09-04
fix: evaluate conditional generation for nested template paths on all platforms

### LINKED ISSUE #2224: Conditional generation is skipped for nested template paths on Windows
### Describe the bug

`conditionalGeneration` (and the deprecated `conditionalFiles`) keys are written with POSIX separators in a template's config, for example `conditionalFolder2/input.txt`. The lookups in `apps/generator/lib/generator.js` build their key with `path.relative`, which returns backslashes on Windows, so a nested key never matches and the condition is silently ignored.

Executed on `win32`:

```
path.relative gives : "conditionalFolder2\input.txt"
config key is       : "conditionalFolder2/input.txt"
lookup matches?     : false
```

Because the lookup misses, `shouldGenerate` stays `true` and the file is generated even when its condition fails. An end-to-end run against the repository's own test template, with `singleFile: 'true'` (which fails the `enum: ["false"]` validation), shows the top-level cases behaving correctly while the nested one does not:

```
conditionalFile.txt exists          (expected false): false
conditionalFolder exists            (expected false): false
conditionalFolder2/input.txt exists (expected false): true   <-- bug
```

The existing nested-path test only asserts the "should generate" direction, so it passes either way and the regression is invisible on the Windows CI leg.

There is a second, platform-independent part. When a `conditionalFiles` entry matches, the code sets `conditionalPath` to the top-level directory rather than the file key:

```js
conditionalPath = relativeSourceDirectory;
```

so `isGenerationConditionMet` finds no config for that path and returns `undefined`. On POSIX that means a file whose condition **is** met gets skipped. The documented example in `configuration-file.md` uses a nested key, so documented usage is affected.

### Expected behavior

A nested `conditionalGeneration` or `conditionalFiles` key is matched on every platform, and a matching `conditionalFiles` entry is evaluated against its own key.

### How to Reproduce

Run the generator on Windows with a template whose config has a nested `conditionalGeneration` key and a parameter that fails the condition; the file is generated anyway. Output above is from executed runs.

---

I have a fix with regression tests ready and would like to open a PR for this.

Generated-by: Claude Code 2.x

AI assisted in investigating this bug and drafting this issue. I have reviewed and verified the analysis and the reproduction output myself.

rajanpanth:
**Description**

- `conditionalGeneration` and `conditionalFiles` keys are written with POSIX separators in a template config (for example `conditionalFolder2/input.txt`), but the lookups build their key from `path.relative`, which returns backslashes on Windows. A nested key therefore never matched and the condition was silently ignored, so the file was generated even when its condition failed. Paths are now normalized to POSIX separators before the lookups.
- When a deprecated `conditionalFiles` entry matched, `conditionalPath` was set to the file's top-level directory instead of the file key, so `isGenerationConditionMet` found no config and returned `undefined`. On POSIX that silently skipped a file whose condition was met. The file key is now used.

Both parts are needed together: normalizing alone would flip Windows from "always generated" to "always skipped".

Executed on `win32`, showing the lookup miss:

```
path.relative gives : "conditionalFolder2\input.txt"
config key is       : "conditionalFolder2/input.txt"
lookup matches?     : false
```

End to end against the repository's own test template with `singleFile: 'true'` (which fails the `enum: ["false"]` validation):

```
conditionalFile.txt exists          (expected false): false
conditionalFolder exists            (expected false): false
conditionalFolder2/input.txt exists (expected false): true   <-- before this change
```

Three regression tests are added to `integration.test.js`. The existing nested-path test only asserted the "should generate" direction, which is why this passed on the Windows CI leg. With the source change reverted and the tests kept, two of them fail with `Expected: false, Received: true`.

Note on scope: `conditionalFiles` is deprecated and tracked for removal in #1553, so the durable value here is the `conditionalGeneration` fix; the `conditionalFiles` key correction is included because it is the same lookup bug. This does not overlap open PR #2215, which only adds a `return true` and tests in `conditionalGeneration.js`.

Verification: `apps/generator` unit suite passes 39/39, and `eslint --max-warnings 0` is clean on the changed test file. The `integration.test.js` suite does not run in my environment (all 12 cases, including untouched ones, fail with `TypeError: Cannot read properties of undefined (reading 'on')` on pristine `master` as well), so those specific cases are verified by CI rather than locally.

A changeset is included for `@asyncapi/generator` (patch).

**Related issue(s)**

Resolves #2224

**AI assistance**

- [x] This PR was created with AI assistance — `Generated-by: Claude Code 2.x`
- [ ] No AI assistance was used

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed conditional generation for files and directories nested within template paths.
  - Conditional rules now work consistently across operating systems by handling path separators correctly.
  - Deprecated `conditionalFiles` configurations now evaluate nested file-path conditions properly, generating files only when their conditions match.

- **Tests**
  - Added coverage for nested conditional files and directories, including matching and non-matching conditions.

### DIFF
--- a/.changeset/conditional-generation-nested-paths.md
+++ b/.changeset/conditional-generation-nested-paths.md
@@ -1,0 +1,5 @@
+---
+"@asyncapi/generator": patch
+---
+
+Fix conditional generation lookups for nested template paths. `conditionalGeneration` and `conditionalFiles` config keys are written with POSIX separators (e.g. `conditionalFolder2/input.txt`), but on Windows `path.relative` produces backslash-separated paths, so nested entries never matched and files were generated even when their condition failed. Additionally, when a deprecated `conditionalFiles` entry matched a nested file path, the condition was looked up under the file's top-level directory instead of the file key itself, so the condition was never evaluated and the file was silently skipped. Paths are now normalized to POSIX separators before the config lookups, and `conditionalFiles` conditions are resolved by their file key.
--- a/apps/generator/lib/generator.js
+++ b/apps/generator/lib/generator.js
@@ -903,7 +903,9 @@ class Generator {
     // Why: conditionalGeneration/conditionalFiles keys are written with POSIX separators in the
     // template config, while path.relative uses backslashes on Windows; normalize before lookups.
     const sourceFileConditionKey = relativeSourceFile.split(path.sep).join('/');
-    const relativeSourceDirectory = sourceFileConditionKey.split('/')[0] || '.';
+    // Why: a conditionalGeneration key can name any ancestor directory, e.g. 'parent/child',
+    // so resolve the full POSIX dirname rather than only the first segment.
+    const relativeSourceDirectory = sourceFileConditionKey.split('/').slice(0, -1).join('/') || '.';
   
     const targetFile = path.resolve(this.targetDir, this.maybeRenameSourceFile(relativeSourceFile));
     const relativeTargetFile = path.relative(this.targetDir, targetFile);
--- a/apps/generator/lib/generator.js
+++ b/apps/generator/lib/generator.js
@@ -734,12 +734,15 @@ class Generator {
   async ignoredDirHandler(root, stats, next) {
     const relativeDir = path.relative(this.templateContentDir, path.resolve(root, stats.name));
     const dirPath = path.resolve(this.targetDir, relativeDir);
-    const conditionalEntry = this.templateConfig?.conditionalGeneration?.[relativeDir];
+    // Why: conditionalGeneration keys are written with POSIX separators in the template config,
+    // while path.relative uses backslashes on Windows; normalize before the lookup.
+    const dirConditionKey = relativeDir.split(path.sep).join('/');
+    const conditionalEntry = this.templateConfig?.conditionalGeneration?.[dirConditionKey];
     let shouldGenerate =  true;
     if (conditionalEntry) {
       shouldGenerate = await isGenerationConditionMet(
         this.templateConfig,
-        relativeDir,
+        dirConditionKey,
         this.templateParams,
         this.asyncapiDocument
       );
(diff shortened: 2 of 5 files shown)


rajanpanth:
Good catch, fixed in f04fd2d.

You are right that `sourceFileConditionKey.split('/')[0]` only ever yields the first segment, so a `conditionalGeneration` key naming a deeper directory was matched by neither lookup:

```
parent/child/input.txt
  old directory key : parent                     -> no match for 'parent/child'
  file key          : parent/child/input.txt     -> no match for 'parent/child'
  posix dirname     : parent/child               -> matches
```

`relativeSourceDirectory` now resolves the full POSIX dirname. Checked the other cases before changing it, since this line is on the path for every generated file:

```
input.txt                    old: input.txt          new: .
conditionalFolder/input.txt  old: conditionalFolder  new: conditionalFolder
parent/child/input.txt       old: parent             new: parent/child
```

Single-level keys are unchanged, which is what the existing tests cover. A root-level file now resolves to `.` instead of its own filename; that is harmless here because the very next branch matches it on the file key, which is how `conditionalFile.txt` was already being resolved in practice.

For the coverage point, the react test template had no multi-level directory key at all, which is why this was invisible. Added a `nestedParent/nestedChild` fixture with a `conditionalGeneration` entry keyed on `singleFolder`, plus two tests: one asserting the file is generated when the condition is met, and one asserting it is not when the condition fails.

`eslint --max-warnings 0` is clean and the unit suite still passes 39/39. As noted in the description, `integration.test.js` does not run in my environment (all cases fail on pristine `master` too with `TypeError: Cannot read properties of undefined (reading 'on')`), so these new cases are verified by CI rather than locally.

derberg:
Please remember to always follow contributor guide. In case of this project PRs are not welcome until issue is noticed and approved for contribution

